PR #6, opened 2017-11-13
 feat: Add Prettier/ESLint/Hooks

lipis:


package.json
```diff
@@ -1,26 +1,37 @@
...
     "dist": "lerna run dist",
-    "lint": "yarn lint:script",
-    "lint:script": "prettier --ignore-path .gitignore --write **/*.{js,ts}",
+    "fix:config": "prettier --write '**/*.json'",
+    "fix:docs": "prettier --ignore-path .gitignore --write '**/*.md'",
+    "fix:jsScript": "npm run test:jsScript -- --fix",
+    "fix:script": "yarn fix:jsScript && yarn fix:tsScript",
+    "fix:style": "prettier --write '**/*.scss'",
+    "fix:tsScript": "prettier --write '**/*.ts'",
+    "fix": "yarn fix:docs && yarn fix:config && yarn fix:style && yarn fix:script",
```
lipis: another idea is to have one `fix:stuff: "prettier --ignore-path .gitignore --write '**/*.{md,json,scss,ts}'"` and `fix:eslint` for the `{js,jsx}` because of the ESLint.. to simplify it..

package.json
```diff
@@ -1,26 +1,37 @@
...
+    "lint-staged": "5.0.0",
+    "prettier": "1.8.2"
   },
   "lint-staged": {
-    "*.{js,ts}": [
-      "yarn lint:script",
-      "git add"
-    ]
+    "*.{js,jsx}": ["eslint --ignore-path .prettierignore --fix", "git add"],
+    "*.json": ["prettier --write", "git add"],
```
bennyn: Why not using ` "*.json": ["yarn fix:config", "git add"]` here?

.prettierrc.json
```diff
@@ -1,11 +0,0 @@
-{
```
bennyn: I prefer `JSON` over `YAML`. JSON is closer to JavaScript whereas YAML is closer to CoffeeScript and since we are on the JavaScript path, we should also be on the JSON path.

package.json
lipis: because this one will run/fix all the `*.json` files that are included in the project instead of just the staging ones..

bennyn: Ah, ok! Didn't knew that.

.prettierrc.json
lipis: YAML was introduced in 2001.. so has nothing to do with CoffeeScript (2009) and for configuration files it's much more sane format, because among other things it also supports comments and there are no useless symbols like: `{ } , ; " "`

Having said that though.. maybe we should vote

https://vote4code.com/5663284820705280/eslintyaml-vs-eslintjson

lipis:
Let's see if it will be 💚

.prettierignore
```diff
@@ -0,0 +1,6 @@
+node_modules/**
+packages/api-client/node_modules/**
```
bennyn: Do you have to exclude each node_modules directory individually when you are already exluding node_modules on the top level?

packages/queue-priority/src/test/node/spec/PriorityQueue.test.js
```diff
@@ -1,88 +1,89 @@
...
-
-      const queue = new PriorityQueue();
-      queue.add(() => notHappyFn()).catch(error => {
-        expect(error.message).toBe('not so happy');
-        done();
-      });
-    });
-  });
-});
+/* eslint-disable no-magic-numbers */
```
bennyn: Do we need "eslint-disable no-magic-numbers" for the whole file?

.prettierignore
lipis: Can't find the relevant issue now on ESLint but they have some issues with that.. and it looks like this trick works.. if you try to remove those lines `eslint ..` takes forever.. go figure :D

packages/queue-priority/src/test/node/spec/PriorityQueue.test.js
lipis: There are quite a few in this file and in general in tests there are quite a few cases that magic numbers could be avoided :D

PR #8, opened 2017-11-14
refactor(api-client): Use UI Kit in API-Client demo

Yserz:


packages/api-client/index.html
```diff
@@ -2,141 +2,16 @@
...
-        width: 100%;
-        border: 0;
-        margin: 0 0 15px;
-        padding: 15px;
-        box-sizing: border-box;
-        font-size: 14px;
-      }
+<body>
+  <div id="app"></div>
+</body>
```
lipis: Indent `<body>` and `<head>`?

AndyTheGiant: I don't think we need to, as you see them not indented in lots of places [citation needed]. The only parent element will always be `<html>` so there is no real gain in indenting them.

lipis: My OCD... but I can see that most of the formatters are doing that.. :) So let's keep it..

PR #9, opened 2017-11-14
feat(react-ui-kit): new loading spinner

AndyTheGiant:


packages/react-ui-kit/.eslintrc.yaml
```diff
@@ -19,7 +19,7 @@ rules:
   no-else-return: 2
   no-inner-declarations: 2
   no-lonely-if: 2
-  no-magic-numbers: [2, {ignore: [-1, 0, 1]}]
+  no-magic-numbers: [2, {ignore: [-1, 0, 1, 2]}]
```
lipis: nooooooooooooooooooooooooooooooo :)

lipis: This one will start affecting too many projects now.. and it's already moved away from here in the #6.. so either use a const whenever is needed or `eslint-disable...`

AndyTheGiant: aaargh ... no-magic-numbers has all my votes for the most annoying rule :)

PR #10, opened 2017-11-14
fix(api-client): styling for api client demo

AndyTheGiant:


packages/react-ui-kit/src/Base.js
```diff
@@ -29,4 +29,7 @@ export default css`
     -moz-osx-font-smoothing: grayscale;
     -webkit-font-smoothing: antialiased;
   }
+  * {
+    box-sizing: border-box;
+  }
```
lipis: let's put the `*` on top.. before the `body` rule as it's more generic

PR #11, opened 2017-11-14
Add webapp-module-logger

lipis:
https://github.com/wireapp/webapp-module-logger


.travis.yml
```diff
@@ -12,6 +12,8 @@ before_install:
   - export PATH="$HOME/.yarn/bin:$PATH"
 
 before_script:
+  - npm install -g grunt-cli
+  - npm install -g bower
```
lipis: Not sure if we have to have that here or not.. https://github.com/wireapp/webapp-module-logger/blob/master/.travis.yml#L11-L15

PR #14, opened 2017-11-15
refactor(react-ui-kit): Refactor file structure

AndyTheGiant:


packages/react-ui-kit/src/Form/Button.jsx
```diff
@@ -17,10 +17,10 @@
  *
  */
 
-import {COLOR} from './variables';
+import {COLOR} from '../variables';
```
bennyn: `import {COLOR} from '../variables';` -> `import {COLOR} from '../Theme';` or sth. similar

packages/react-ui-kit/src/Grid/Column.jsx
```diff
@@ -17,8 +17,8 @@
  *
  */
 
-import {SIZE} from './variables';
-import {media} from './mixins';
+import {SIZE} from '../variables';
```
bennyn: Sizes can be part of the "Grid"

packages/react-ui-kit/package.json
```diff
@@ -35,5 +23,24 @@
...
+    "clear": "rimraf dist",
+    "dist": "yarn clear && yarn babel src --out-dir dist --no-comments",
+    "fix:script": "yarn test --fix",
+    "fix": "yarn fix:script",
+    "precommit": "lint-staged",
+    "prepare": "yarn dist && node scripts/prepare.js",
+    "start": "yarn start:demo",
+    "start:static": "gulp",
+    "start:demo": "webpack-dev-server",
+    "test": "eslint -c .eslintrc.yaml --ignore-path .gitignore --ignore-path .prettierignore '**/*.{js,jsx}'",
```
lipis: remove the test from here and the fix related stuff

lipis: precommit etc.. the ones that are red in this diff :D

packages/react-ui-kit/src/Identity/colors.js
```diff
@@ -106,18 +106,3 @@ export const COLOR = {
   ...LIGHT_COLOR,
   ...OPAQUE_COLOR,
 };
-
-export const FONT = {
-  fontFamily: '-apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Helvetica Neue, Helvetica, Arial, sans-serif',
-};
```
lipis: I would argue that we could keep that even if it's used only in one place.. just to be able to find it most likely inside the variables/settings section..

AndyTheGiant: nope, we don't need it.

lipis: :(

PR #15, opened 2017-11-15
fix(core): Add parameter for client creation on login

Yserz:


packages/core/src/main/Account.ts
```diff
@@ -132,23 +132,29 @@ export default class Account extends EventEmitter {
...
 
-  private initClient(context: Context): Promise<RegisteredClient> {
+  private initClient(context: Context, shouldRegisterNewClient: boolean): Promise<RegisteredClient> {
     this.context = context;
     return this.loadExistingClient().catch(error => {
       if (error instanceof RecordNotFoundError) {
-        return this.registerNewClient();
+        if (shouldRegisterNewClient) {
+          return this.registerNewClient();
+        } else {
```
gregor: No else needed fater return

packages/core/src/main/Account.ts
```diff
@@ -132,23 +132,29 @@ export default class Account extends EventEmitter {
       .then(() => this.client);
   }
 
-  private initClient(context: Context): Promise<RegisteredClient> {
+  private initClient(context: Context, shouldRegisterNewClient: boolean): Promise<RegisteredClient> {
```
bennyn: We don't need a `shouldRegisterNewClient` parameter here.

PR #18, opened 2017-11-15
refactor: Move login parameters to login function

Yserz:


packages/core/src/main/Account.ts
```diff
@@ -80,23 +75,27 @@ export default class Account extends EventEmitter {
...
 
-    if (this.loginData.password) {
-      this.loginData.password = this.loginData.password.toString().replace(removeNonPrintableCharacters, '');
+    if (loginData.password) {
+      loginData.password = loginData.password.toString().replace(removeNonPrintableCharacters, '');
     }
 
-    if (this.loginData.handle) {
-      this.loginData.handle = this.loginData.handle.replace(removeNonPrintableCharacters, '');
+    if (loginData.handle) {
```
gregor: If you touch it, you could do the alphabet and also add phone number.

PR #19, opened 2017-11-16
refactor(react-ui-kit): update Layout structure

AndyTheGiant:


packages/react-ui-kit/src/Layout/index.js
```diff
@@ -4,3 +4,7 @@ setGlobalStyles();
 
 export * from './Column';
 export * from './Container';
+export * from './StyledApp';
+export * from './Content';
```
bennyn: alphabetical order

PR #25, opened 2017-11-17
feat(react-ui-kit): additional components

AndyLnd:


packages/react-ui-kit/src/Icon/Profile.jsx
```diff
@@ -0,0 +1,47 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const size = 31;
+
+function Profile({color, scale}) {
```
Yserz: Let's call it ProfileIcon for now

packages/react-ui-kit/src/Icon/Team.jsx
```diff
@@ -0,0 +1,49 @@
...
+ *
+ */
+
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const width = 28;
+const height = 31;
+
+function Team({color, scale}) {
```
Yserz: Let's call it TeamIcon for now

PR #31, opened 2017-11-20
 feat(react-ui-kit): new icons

AndyLnd:


packages/react-ui-kit/src/Icon/PlaneIcon.jsx
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import IconBase from './IconBase';
+import React from 'react';
+
+class PlaneIcon extends IconBase {
```
gregor: Let's stick to the names the design team uses: https://github.com/wireapp/wire-webapp/blob/dev/app/style/fonts/zeta-neue.css

PR #41, opened 2017-11-22
feat(react-ui-kit): codeinput component

AndyLnd:


packages/core/src/main/auth/LoginSanitizer.ts
```diff
@@ -1,8 +1,7 @@
 import {LoginData} from '@wireapp/api-client/dist/commonjs/auth/';
 
 export default class LoginSanitizer {
-  constructor() {
-  }
+  constructor() {}
```
lipis: do we actually need a constructor if it's empty?

PR #46, opened 2017-11-23
feat(react-ui-kit): added CheckboxLabel component, updated button style and font sizes

AndyLnd:


packages/react-ui-kit/src/Form/Checkbox.jsx
```diff
@@ -40,24 +43,42 @@ const StyledCheckbox = Input.withComponent('input').extend.attrs({
   margin-bottom: 0;
 `;
 
-const Checkbox = ({id, children, ...props}) => {
+const Checkbox = ({id, children, style, ...props}) => {
+  if (id === null) {
```
lipis: `if (!id) {` ...

packages/react-ui-kit/src/Form/Checkbox.jsx
```diff
@@ -17,10 +17,13 @@
  *
  */
 
+import {COLOR} from '../index';
 import {Input} from './Input';
 import PropTypes from 'prop-types';
 import React from 'react';
+import {Text} from '../Text';
 import styled from 'styled-components';
+import {v4 as uuid} from 'uuid';
```
bennyn: Do you really need a `uuid` in a styled component? To me UUID sounds like business logic. Nothing that you would want to important in a presentation layer.

packages/react-ui-kit/src/Form/CodeInput.jsx
```diff
@@ -126,6 +131,7 @@ class CodeInput extends React.PureComponent {
     for (let i = 0; i < this.props.digits; i++) {
       inputs.push(
         <DigitInput
+          autoFocus={i === 0 && this.props.autoFocus}
```
lipis: `{i && this.props.autoFocus}`

packages/react-ui-kit/src/Form/Checkbox.jsx
AndyLnd: actually i really want to check against null, as this is the default value i set.

packages/react-ui-kit/src/Form/CodeInput.jsx
lipis: `{!i && this.props.autoFocus}`

packages/react-ui-kit/src/Form/CodeInput.jsx
```diff
@@ -126,6 +131,7 @@ class CodeInput extends React.PureComponent {
     for (let i = 0; i < this.props.digits; i++) {
```
lipis: `index` instead of `i`

packages/react-ui-kit/src/Form/Checkbox.jsx
lipis: oki..

packages/react-ui-kit/src/Form/CodeInput.jsx
```diff
@@ -123,21 +128,22 @@ class CodeInput extends React.PureComponent {
...
-          onChange={e => this.setValue(i, e.target.value)}
-          onPaste={e => this.handlePaste(i, e.clipboardData.getData('Text'))}
+          autoFocus={index === 0 && this.props.autoFocus}
+          key={index}
+          onChange={e => this.setValue(index, e.target.value)}
+          onPaste={e => this.handlePaste(index, e.clipboardData.getData('Text'))}
           onFocus={e => e.target.select()}
-          onKeyDown={e => this.handleKeyDown(i, e)}
-          innerRef={node => (this.inputs[i] = node)}
+          onKeyDown={e => this.handleKeyDown(index, e)}
```
lipis: `e` -> `error`

lipis: https://github.com/wireapp/wire-web-packages/pull/47

packages/react-ui-kit/src/Form/Checkbox.jsx
AndyLnd: Maybe not. Actually I just need some random string. Something like Math.random().toString() would work as well.

packages/react-ui-kit/src/Form/Checkbox.jsx
AndyLnd: but yeah, as every falsy value would not be fit as a key, i am going to change that.

packages/react-ui-kit/src/Form/Checkbox.jsx
bennyn: 👍 , then we can also remove 'uuid' from the dependencies?

PR #68, opened 2017-11-28
 feat(react-ui-kit): input invalid marker, links in error messages, checkmark icon, sleaker text css

AndyLnd:


packages/react-ui-kit/src/Form/Input.jsx
```diff
@@ -27,6 +27,15 @@ const placeholderStyle = css`
   text-transform: ${props => props.placeholderTextTransform};
 `;
 
+const dotRadius = 4;
+const dotSize = dotRadius + dotRadius;
```
lipis: or this.. since you are avoiding magic numbers..

```js
const dotSize = dotRadius << 1;
```

AndyLnd: Maybe i will do that 😄 
But it wouldn't work with fractional numbers ...

lipis: Jokes aside I prefer `<< 1` over `+` here :)

lipis: also `dotDiameter` is more suitable in this case :)

packages/react-ui-kit/src/Icon/CheckIcon.jsx
```diff
@@ -0,0 +1,36 @@
...
+ *
+ */
+
+import IconBase from './IconBase';
+import React from 'react';
+
+class CheckIcon extends IconBase {
+  /* eslint-disable no-magic-numbers */
+  width = 16;
+  height = 12;
```
lipis: in theory we need to enable it as well afterwards..

packages/react-ui-kit/src/Form/Input.jsx
AndyLnd: i changed everything, since svg allows me to scale the image as a whole

PR #74, opened 2017-11-29
fix(react-ui-kit): plane icon centered, better code input behaviour, …

AndyLnd:
…error message link color

### DIFF
--- a/packages/react-ui-kit/demo/Demo.jsx
+++ b/packages/react-ui-kit/demo/Demo.jsx
@@ -215,10 +215,10 @@ export default function Demo() {
               <Input placeholder="Placeholder" type="email" required />
               <Button type="submit">Submit Button</Button>
               <InputBlock>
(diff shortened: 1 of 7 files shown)


packages/react-ui-kit/demo/Demo.jsx
```diff
@@ -215,10 +215,10 @@ export default function Demo() {
               <Input placeholder="Placeholder" type="email" required />
               <Button type="submit">Submit Button</Button>
               <InputBlock>
-                <Input placeholder="InputBlock" />
+                <Input placeholder="InputBlock" markInvalid />
```
lipis: what iz daz?

AndyLnd: a helper prop, in some cases we (and the design guys) want a little indicator if there is an error in this input field. the native :invalid selector unfortunately doesn't cut it.

PR #107, opened 2017-12-11
feat(react-ui-kit): Label component & webpack demo fixes

Yserz:


packages/react-ui-kit/src/Text/index.js
```diff
@@ -25,4 +25,5 @@ export * from './Text';
 export {Heading, H1, H2, H3, H4} from './Heading';
 export * from './Line';
 export * from './Link';
+export * from './Label';
```
gregor: ABCDEFGHI ;P

PR #108, opened 2017-12-11
feat: Component aware label

Yserz:


packages/react-ui-kit/src/Text/Label.jsx
```diff
@@ -17,23 +17,31 @@
...
-  /* appearance */
-  font-size: 12px;
-  font-weight: 600;
-  /* positioning */
-  width: 100%;
-  padding: 24px 0 8px;
-`;
+const Label = ({component, ...props}) => {
+  const StyledLabel = Text.withComponent(component).extend`
+    /* appearance */
```
bennyn: can be removed if there is nothing

PR #159, opened 2018-01-08
feat: Add wire-web-cbor

ffflorian:


packages/cbor/src/cbor/Types.js
```diff
@@ -0,0 +1,184 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+/* eslint no-magic-numbers: "off" */
```
ffflorian: I know, bad style, but maybe someone has a better idea instead of adding
```js
const two = 2;
return two;
```
for every function?

packages/cbor/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "webpack": "3.10.0"
+  },
+  "files": ["dist/"],
+  "license": "GPL-3.0",
+  "main": "src/cbor.js",
+  "name": "@wireapp/cbor",
+  "repository": "https://github.com/wireapp/wire-web-packages/tree/master/packages/cbor",
+  "scripts": {
+    "clear": "rimraf dist",
+    "coverage:html": "istanbul cover --report html ../../node_modules/jasmine/bin/jasmine.js",
```
Yserz: Can we improve on these paths?

packages/cbor/README.md
```diff
@@ -0,0 +1,89 @@
...
+## Usage
+
+### Browser
+
+* [index.html](./dist/index.html)
+
+### TypeScript
+
+```typescript
+import * as CBOR from 'wire-webapp-cbor';
```
Yserz: Is this import up-to-date?

PR #171, opened 2018-01-09
refactor: Enable all strict type checking options

bennyn:
This PR introduces two major improvements:
1. It enables all strict type checking options available in TypeScript ([--strict](https://www.typescriptlang.org/docs/handbook/compiler-options.html))
1. All our TypeScript modules now extend a common configuration


packages/cli-client/tsconfig.json
```diff
@@ -1,17 +1,8 @@
...
     "outDir": "dist/commonjs",
-    "removeComments": true,
-    "rootDir": "src/main",
-    "sourceMap": false,
-    "target": "ES5"
+    "rootDir": "src/main"
   },
-  "exclude": ["dist/commonjs", "node_modules"]
+  "exclude": ["dist/commonjs", "node_modules"],
+  "extends": "../../configs/tsconfig.json"
```
ffflorian: :+1:

packages/core/src/main/crypto/CryptographyService.ts
```diff
@@ -24,9 +24,9 @@ export default class CryptographyService {
         .map(preKey => {
           const preKeyJson: auth.PreKey = this.cryptobox.serialize_prekey(preKey);
           if (preKeyJson.id !== Proteus.keys.PreKey.MAX_PREKEY_ID) return preKeyJson;
```
ffflorian:
```js
if (preKeyJson.id !== Proteus.keys.PreKey.MAX_PREKEY_ID) {
  return preKeyJson;
}
```

PR #196, opened 2018-01-11
fix(react-ui-kit): Fix release packaging

Yserz:


packages/react-ui-kit/package.json
```diff
@@ -26,11 +26,24 @@
...
+    "Modal.js",
+    "Progress.js",
+    "Text.js"
+  ],
   "name": "@wireapp/react-ui-kit",
   "repository": "https://github.com/wireapp/wire-web-packages/tree/master/packages/react-ui-kit",
   "scripts": {
-    "build": "babel src --out-dir dist --no-comments && node scripts/prepare.js",
+    "build":
+      "babel alias --out-dir ./ --no-comments && babel src --out-dir dist --no-comments && node scripts/prepare.js",
```
AndyLnd: shouldn't the `&& node scripts/prepare.js` be removed as well?

bennyn:
Am I right that with this PR you would need to do the following when you create a new React UI Kit component?

1. code it
1. Exclude it from `.gitignore`
1. Create an `alias`
1. List it in `files` of `package.json`

Yserz:
It doesn't apply to every single component but every component group.

bennyn:
Instead of doing that can we maybe switch the working directory on `prepublish`?
 https://stackoverflow.com/questions/30286498/change-working-directory-for-npm-scripts

Yserz:
While this works when running directly with NPM this is a problem when executing publish with lerna. In addition to that we would have to overwrite the NPM publish command which doesn't result in the expected behaviour of actually overwriting it. Instead it gets executed after the execution of the internal NPM publish which in turn results in a double publish.

PR #197, opened 2018-01-11
refactor: Streamline NPM scripts

Yserz:


packages/react-ui-kit/package.json
```diff
@@ -42,17 +42,13 @@
...
     "clear": "rimraf dist",
     "dist": "yarn clear && yarn build",
-    "prepublish": "yarn test",
-    "publish": "cd ./dist && npm publish",
     "start:demo": "webpack-dev-server",
     "start:static": "gulp",
     "start": "yarn start:demo",
-    "test": "exit 0",
-    "prepare": "yarn dist"
+    "test": "yarn dist"
```
AndyLnd: is this correct on `test`?

bennyn: A `test` script should execute tests. A `dist` script should create distributables. Looks like a violation of this paradigm here.

Yserz: Yes, the idea is to have `dist` before testing anything. In this case there are no tests and it was `exit 0` before.

bennyn: Ah, I see. So it's a shrinked version of `"test": "yarn dist && yarn test:node && yarn test:browser"` because there are no tests atm. :)

Yserz: Exactly previously I had an explicit `exit 0` should I put it back in there?

bennyn: But then I would do it like we do in `cli-client/package.json`:

```json
"test": "yarn dist && yarn test:node && yarn test:browser",
"test:browser": "exit 0"
```

PR #202, opened 2018-01-15
fix: Release fixes

Yserz:


packages/react-ui-kit/.prettierignore
```diff
@@ -1,2 +1,11 @@
...
 dist
+
+index.js
+Form.js
+Icon.js
+Identity.js
+Layout.js
+Modal.js
+Progress.js
+Text.js
```
gregor: Can we generalize this?

.prettierignore
```diff
@@ -2,3 +2,4 @@
 **/dist/**
 **/node_modules/**
 **/bower_components/**
+packages/react-ui-kit/*.js
```
lipis: I guess there are reasons.. but why inviting those?

PR #238, opened 2018-01-30
refactor(proteus): Rewrite to TypeScript

ffflorian:


bennyn:
Maybe we should close it in favor of https://github.com/wireapp/wire-web-packages/tree/feat/proteus-async

PR #253, opened 2018-02-08
refactor(proteus): TypeScript rewrite with async libsodium

ffflorian:


package.json
```diff
@@ -11,9 +11,9 @@
     "prettier": "1.10.2"
   },
   "lint-staged": {
-    "*.{js,jsx}": ["eslint --ignore-path .prettierignore --fix", "git add"],
-    "*.json": ["prettier --write", "git add"],
-    "*.md": ["prettier --write", "git add"],
+    "*.{js,jsx}": ["eslint --ignore-path .prettierignore-jsScript --fix", "git add"],
+    "*.json": ["prettier --ignore-path .prettierignore-config --write", "git add"],
+    "*.md": ["prettier --ignore-path .prettierignore-docs --write", "git add"],
```
lipis: I don't get it.. why do we have three different ignore files?

bennyn: Because Prettier doesn't support having multiple ignore patterns.

lipis: w00t? but why not having one ignore file and put all the rules there..?

bennyn: Let's try

package.json
```diff
@@ -23,8 +23,8 @@
     "clear": "lerna run clear",
     "dist": "yarn clear && lerna run dist",
     "fix": "yarn fix:docs && yarn fix:config && yarn fix:style && yarn fix:script",
-    "fix:config": "prettier --write **/*.json",
-    "fix:docs": "prettier --ignore-path .gitignore --write **/*.md",
+    "fix:config": "prettier --ignore-path .prettierignore --write **/*.json",
+    "fix:docs": "prettier --ignore-path .prettierignore --write **/*.md",
```
lipis: also `.prettierignore` is ignored by default.. so you don't need to specifically say it

lipis: https://prettier.io/docs/en/ignore.html

lipis: also instead of `fix:docs, fix:config, fix:style` we can have one:

```
"fix:assets": "prettier --write **/*.{json,md,scss}",
```

package.json
```diff
@@ -12,8 +12,8 @@
   },
   "lint-staged": {
     "*.{js,jsx}": ["eslint --ignore-path .prettierignore --fix", "git add"],
-    "*.json": ["prettier --write", "git add"],
-    "*.md": ["prettier --write", "git add"],
+    "*.json": ["prettier --ignore-path .prettierignore --write", "git add"],
+    "*.md": ["prettier --ignore-path .prettierignore --write", "git add"],
     "*.scss": ["prettier --write", "git add"],
     "*.ts": ["prettier --write", "git add"]
```
lipis:
```
"*.{json,md,scss,ts}": ["prettier --write", "git add"]
```

package.json
```diff
@@ -11,9 +11,8 @@
     "prettier": "1.10.2"
   },
   "lint-staged": {
-    "*.{js,jsx}": ["eslint --ignore-path .prettierignore --fix", "git add"],
-    "*.json": ["prettier --write", "git add"],
-    "*.md": ["prettier --write", "git add"],
+    "*.{js,jsx}": ["eslint --fix", "git add"],
+    "*.{json,md,scss,ts}": ["prettier --write", "git add"],
     "*.scss": ["prettier --write", "git add"],
     "*.ts": ["prettier --write", "git add"]
```
lipis: delete the lines 16/17

bennyn:
@ffflorian: PR is green for the very first time! 🎉 

To proceed we need to merge this branch and make a **MAJOR** release of Proteus. 

After doing that we can fix Cryptobox for browsers (because that will be broken). Before merging this PR please talk with @Yserz. If possible we want to make a major Proteus release **WITHOUT** making a Cryptobox release. If it isn't possible to seperate these releases then Cryptobox should also get a **MAJOR** release.

bennyn:
Good news, [Cryptobox browser demo](https://github.com/wireapp/wire-web-packages/pull/253/commits/731fd210248869bac64cc4ba343216567903adc6) now works. So you can go ahead and release major versions of Proteus & Cryptobox (plus all their dependencies). 😍

PR #273, opened 2018-02-16
refactor(api-client,core): Drop internal default exports & default values for clients model and location

Yserz:


packages/core/src/main/Account.ts
```diff
@@ -399,7 +401,9 @@ export default class Account extends EventEmitter {
   public registerClient(
     loginData: LoginData,
     clientClassification: ClientClassification = ClientClassification.DESKTOP,
-    cookieLabel: string = 'default'
+    cookieLabel: string = 'default',
+    model: string = `${pkg.name} v${pkg.version}`,
```
bennyn: model, classification & cookie can be merged in one object

PR #280, opened 2018-02-19
fix(cryptobox): Queuing encryptions for multipe clients

bennyn:


packages/core/src/main/crypto/CryptographyService.ts
```diff
@@ -85,7 +85,10 @@ export default class CryptographyService {
     return this.cryptobox
       .encrypt(sessionId, plainText, decodedPreKeyBundle.buffer)
       .then((encryptedPayload: ArrayBuffer) => Encoder.toBase64(encryptedPayload).asString)
-      .catch((error: Error) => '💣')
+      .catch((error: Error) => {
+        console.log('EEEEEEEEER', error);
```
ffflorian: Let's remove this :wink:

PR #291, opened 2018-02-21
fix(core): Remove init call from login function

Yserz:


bennyn:
We can merge this but we also need to find a clean way of how to do a access token refresh on startup.

PR #295, opened 2018-02-22
fix(store-engine,api-client): Add isInitialized flag for stores

Yserz:


bennyn:
Will be fixed with https://github.com/wireapp/wire-web-packages/pull/291

PR #343, opened 2018-03-07
feat(priority-queue): Add convenience features (delete, deleteAll & label)

bennyn:


packages/priority-queue/src/main/Item.ts
```diff
@@ -15,14 +15,20 @@
...
+  label: string = '';
   priority: number = Priority.MEDIUM;
-  reject: Function = () => {};
-  resolve: Function = () => {};
-  retry: number | undefined = 0;
-  timestamp: number = 0;
+  reject: Function = () => {}; // wrapped "reject" of "fn"
+  resolve: Function = () => {}; // wrapped "resolve" of "fn"
+  retry: number = Infinity; // number of retries for rejecting Promises
+  timestamp: number = 0; // time when the item has been added to the queue
```
ffflorian:
```js
/** time when the item has been added to the queue */
timestamp: number = 0;
```

packages/priority-queue/src/main/Item.ts
```diff
@@ -15,14 +15,19 @@
...
+  label: string | undefined;
   priority: number = Priority.MEDIUM;
+  /** wrapped `reject` of `fn` */
   reject: Function = () => {};
+  /** wrapped `resolve` of `fn` */
   resolve: Function = () => {};
-  retry: number | undefined = 0;
+  /** number of retries for rejecting Promises */
+  retry: number = Infinity;
+  /** time when the item has been added to the queue */
```
lipis: Is that some sort of special comment? Can we just use `//`?

ffflorian: It's JSDoc, hover over the variable in VSCode to see the description.

PR #350, opened 2018-03-08
fix(core): Update cookie if it already exists

bennyn:


packages/core/src/demo/sender.js
```diff
@@ -9,30 +9,33 @@ const path = require('path');
...
-      const timeoutInMillis = 2000;
-      setTimeout(async () => {
-        account.service.conversation.sendTextMessage(CONVERSATION_ID, 'Hello World!').then(() => sendMessage());
-      }, timeoutInMillis);
-    }
+  function sendMessage() {
+    const timeoutInMillis = 2000;
+    setTimeout(() => {
+      account.service.conversation.sendTextMessage(CONVERSATION_ID, 'Hello World!').then(() => sendMessage());
+    }, timeoutInMillis);
```
ffflorian:
```js
setTimeout(async () => {
  await account.service.conversation.sendTextMessage(CONVERSATION_ID, 'Hello World!');
  sendMessage();
}, timeoutInMillis);
```

packages/core/src/demo/sender.js
```diff
@@ -9,30 +9,33 @@ const path = require('path');
...
+  }
 
-    sendMessage();
-  })
-  .catch(error => {
-    console.error(error);
-    process.exit(1);
-  });
```
ffflorian:
```js
try {
...
} catch(error) {
  console.error(error);
}
```

PR #361, opened 2018-03-13
chore(cryptobox): Add benchmark with ping-pong scenario

bennyn:


packages/cryptobox/src/demo/benchmark.js
```diff
@@ -4,16 +4,23 @@ const Proteus = require('@wireapp/proteus');
...
 };
 
+function createSessionId(receiver) {
+  return `session-with-${receiver.identity.public_key.fingerprint()}`;
+}
+
+function numbersInArray(count) {
+  return Array(count)
+    .fill(null)
+    .map((value, key) => key + 1);
```
ffflorian: To avoid ESLint's `no-magic-numbers` rule, we could also write:
```js
.map((value, key) => ++key);
```

PR #374, opened 2018-03-15
feat(core): Add support for generic messages

ffflorian:


packages/core/package.json
```diff
@@ -1,4 +1,7 @@
 {
+  "browser": {
+    "./dist/commonjs/shims/node/assetCryptography": "./dist/commonjs/shims/browser/assetCryptography.js"
```
bennyn: Let's organize shims by context instead of placing them all in a `shims` directory.

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -19,44 +40,106 @@ export default class ConversationService {
...
+
+    return this.sendMessage(this.clientID, conversationId, payload);
   }
 
   public setClientID(clientID: string) {
     this.clientID = clientID;
   }
+
+  private shouldSendAsExternal(plainText: Buffer, preKeyBundles: UserPreKeyBundleMap): boolean {
+    const EXTERNAL_MESSAGE_THRESHOLD = 200 * 1024;
```
bennyn: Let's comment the unit of that threshold (Bytes, Bits, ...)

packages/core/src/main/shims/browser/assetCryptography.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {EncryptedAsset} from '../../crypto/root';
+const {crypto} = window;
+
+const equalHashes = (bufferA: Buffer, bufferB: Buffer): boolean => {
```
bennyn: `equalHashes` -> `isEqual`

packages/core/src/test/node/spec/conversation/ConversationService.test.js
```diff
@@ -0,0 +1,96 @@
...
+};
+
+describe('ConversationService', () => {
+  let account;
+
+  beforeAll(async done => {
+    const engine = new MemoryEngine();
+    await engine.init('');
+
+    const client = new APIClient(new Config(engine, APIClient.BACKEND.STAGING));
```
bennyn: In the test we can init our API client with an object: `new APIClient({store: engine, urls: APIClient.BACKEND.STAGING);`

We should use the object initialization because on the long run we want to get rid off `new Config()`.

packages/core/src/test/node/spec/conversation/ConversationService.test.js
```diff
@@ -0,0 +1,96 @@
...
+    done();
+  });
+
+  describe("'shouldSendAsExternal'", () => {
+    it('returns true for a big payload', async done => {
+      const {conversation} = account.service;
+      const preKeyBundles = generatePreKeyBundles(128, 4);
+
+      const longMessage =
+        'massive external message massive external message massive external message massive external message massive external message massive external message massive external message massive external message massive external messagemassive external message massive external message massive external message massive external message massive external message massive external message massive external message massive external message massive external messagemassive external message massive external message massive external message massive external message massive external message massive external message massive external message massive external message massive external messagemassive external message massive external message massive external message massive external message massive external message massive external message massive external message massive external message massive external message';
```
bennyn: http://www.loremipsum.de 💃

packages/core/src/main/shims/node/assetCryptography.ts
```diff
@@ -0,0 +1,77 @@
...
+  cipherText,
+  keyBytes,
+  sha256: referenceSHA256,
+}: EncryptedAsset): Promise<Buffer> => {
+  const computedSHA256 = crypto
+    .createHash('SHA256')
+    .update(cipherText)
+    .digest();
+
+  if (!equalHashes(computedSHA256, referenceSHA256)) {
```
bennyn: We should always consider the positive case first: 

```ts
if (equalHashes(computedSHA256, referenceSHA256)) { 
  ...
} else { 
 throw new Error('Encrypted asset does not match its SHA-256 hash'); 
}
```

packages/core/src/main/shims/browser/assetCryptography.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export const decryptAsset = async ({
+  cipherText,
+  keyBytes,
+  sha256: referenceSHA256,
+}: EncryptedAsset): Promise<Buffer> => {
+  const computedSHA256 = await crypto.subtle.digest('SHA-256', cipherText);
+
+  if (!equalHashes(new Buffer(computedSHA256), referenceSHA256)) {
```
bennyn: Always consider the positive case first because our brain doesn't like to think in boolean negation.

packages/core/src/main/shims/browser/assetCryptography.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ */
+
+import {EncryptedAsset} from '../../crypto/root';
+const {crypto} = window;
+
+const equalHashes = (bufferA: Buffer, bufferB: Buffer): boolean => {
+  const arrayA = new Uint32Array(bufferA);
+  const arrayB = new Uint32Array(bufferB);
+  return arrayA.length === arrayB.length && arrayA.every((value, index) => value === arrayB[index]);
```
bennyn:
```ts
return arrayA.length === arrayB.length && arrayA.every((value, index) => value === arrayB[index]);
```

->

```ts
const hasSameLength = arrayA.length === arrayB.length;
const hasSameValues = arrayA.every((value, index) => value === arrayB[index])
return hasSameLength && hasSameValues 
```

packages/core/src/main/shims/browser/assetCryptography.ts
bennyn: `bufferA` and `bufferB` can be just named `a` and `b` because TypeScript tells you already that these are Buffers. That's also common sense: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

packages/core/src/main/shims/node/assetCryptography.ts
ffflorian: :-1: :-1: :-1:

bennyn:
Let's apply Michael's `BackendRepository` & `DatabaseRepository` pattern here: https://github.com/wireapp/wire-web-packages/tree/%40wireapp/travis-bot%400.1.3/packages/core/src/main/notification

packages/core/src/main/Account.ts
```diff
@@ -67,7 +67,7 @@ class Account extends EventEmitter {
       switch (event.type) {
         case ConversationEventType.OTR_MESSAGE_ADD: {
           const otrMessage: OTRMessageAdd = event as OTRMessageAdd;
-          const sessionId: string = this.service.crypto.constructSessionId(otrMessage.from, otrMessage.data.sender);
+          const sessionId: string = CryptographyService.constructSessionId(otrMessage.from, otrMessage.data.sender);
```
bennyn: In `Account.ts` we defined the following:

```js
        this.service = {
          conversation: conversationService,
          crypto: cryptographyService,
          notification: notificationService,
        };
```

It should become:

```js
        this.service = {
          conversation: conversationService,
          cryptography: cryptographyService,
          notification: notificationService,
        };
```

PR #376, opened 2018-03-16
fix(core): Validate local client with backend

Yserz:


packages/core/src/main/Account.ts
```diff
@@ -351,14 +352,23 @@ class Account extends EventEmitter {
...
+      .loadClient()
+      .then(client => (loadedClient = client))
+      .then(() => this.apiClient.client.api.getClient(loadedClient.id))
+      .then(() => {
+        this.service!.conversation.setClientID(<string>this.context!.clientId);
+        return loadedClient;
+      })
+      .catch(error => {
+        // There was no client so we need to "create" and "register" a client
+        if (error instanceof cryptobox.error.CryptoboxError || error.code === StatusCode.NOT_FOUND) {
```
bennyn:
```ts
if (error instanceof cryptobox.error.CryptoboxError || error.code === StatusCode.NOT_FOUND)
```  
->
```ts
if (notFoundInDatabase || notFoundOnBackend)
```

PR #378, opened 2018-03-16
 fix(core): Initialize notification stream on client creation

Yserz:


packages/core/src/main/notification/NotificationService.ts
```diff
@@ -0,0 +1,59 @@
+import {CRUDEngine} from '@wireapp/store-engine/dist/commonjs/engine/index';
```
ffflorian:
```js
import {CRUDEngine} from '@wireapp/store-engine/dist/commonjs/engine/';
```

packages/core/src/main/notification/NotificationService.ts
```diff
@@ -0,0 +1,59 @@
...
+import APIClient = require('@wireapp/api-client');
+import {Notification} from '@wireapp/api-client/dist/commonjs/notification/index';
```
ffflorian:
```js
import {Notification} from '@wireapp/api-client/dist/commonjs/notification/';
```

packages/core/src/main/notification/NotificationService.ts
```diff
@@ -0,0 +1,59 @@
...
+  };
+
+  public static KEYS = {
+    PRIMARY_KEY_LAST_EVENT: 'z.storage.StorageKey.EVENT.LAST_DATE',
+    PRIMARY_KEY_LAST_NOTIFICATION: 'z.storage.StorageKey.NOTIFICATION.LAST_ID',
+  };
+
+  constructor(private apiClient: APIClient, private storeEngine: CRUDEngine) {}
+
+  public initializeNotificationStream(clientId: string): Promise<any> {
```
ffflorian:
```js
public initializeNotificationStream(clientId: string): Promise<string> {
```

packages/core/src/main/Account.ts
```diff
@@ -20,6 +20,7 @@ const pkg = require('../package.json');
 import {IncomingNotification} from '@wireapp/api-client/dist/commonjs/conversation/index';
```
ffflorian:
```js
import {IncomingNotification} from '@wireapp/api-client/dist/commonjs/conversation/';
```

packages/core/src/main/Account.ts
```diff
@@ -20,6 +20,7 @@ const pkg = require('../package.json');
...
 import * as cryptobox from '@wireapp/cryptobox';
 import {CryptographyService, GenericMessageType, PayloadBundle} from './crypto/root';
+import {NotificationService} from './notification/root';
 import {Context, LoginData, PreKey} from '@wireapp/api-client/dist/commonjs/auth/index';
```
ffflorian:
```js
import {Context, LoginData, PreKey} from '@wireapp/api-client/dist/commonjs/auth/';
```

packages/core/src/main/notification/NotificationService.ts
```diff
@@ -0,0 +1,59 @@
+import {CRUDEngine} from '@wireapp/store-engine/dist/commonjs/engine/index';
+import APIClient = require('@wireapp/api-client');
+import {Notification} from '@wireapp/api-client/dist/commonjs/notification/index';
+import {RecordNotFoundError} from '@wireapp/store-engine/dist/commonjs/engine/error';
+
+export default class NotificationService {
+  public static STORES = {
+    AMPLIFY: 'amplify',
```
bennyn: Looking back it was a terrible decission to name the table `amplify`. Do we need that name for backward compatibility or do we have a chance now to change it?

bennyn: Looking back it was a terrible decission to name the table "`amplify`". Do you see a chance that we can change this name or do we need to keep it to be backward compatible?

Yserz: Unfortunately we need it to be backward compatible

packages/core/src/main/notification/NotificationService.ts
Yserz: We decided to explicitly import `index` because of compatibility issues. Later on we will move to `root` instead of `index` to enforce it.

bennyn: With our new export syntax we might be able to use `index.ts`. We should check that once more because this is common sense: https://basarat.gitbooks.io/typescript/docs/tips/barrel.html

PR #379, opened 2018-03-18
refactor(cbor): Rewrite in TypeScript

ffflorian:
:boom:


packages/cbor/src/main/cbor/BaseError.ts
```diff
@@ -17,21 +17,15 @@
...
+class BaseError extends Error {
+  constructor(public message: string) {
+    super(message);
+    Object.setPrototypeOf(this, BaseError.prototype);
 
-  BaseError.prototype = new Error();
-  BaseError.prototype.constructor = BaseError;
+    this.message = message;
+    this.name = this.constructor.name;
+    //this.stack = new Error().stack;
```
lipis: Use or delete commented code.

packages/cbor/src/main/cbor/Types.ts
```diff
@@ -0,0 +1,96 @@
...
+      case Types.NULL:
+        return 7;
+      case Types.OBJECT:
+        return 5;
+      case Types.TAGGED:
+        return 6;
+      case Types.TEXT:
+        return 3;
+      case Types.UNDEFINED:
+        return 7;
```
lipis: Can we name these magic numbers?! What are these 3, 6, 7?

ffflorian: They are already named with `TEXT`, `INT32`, etc. ...

lipis: Then let's replace them with `Types.FLOAT64`, `Types.TEXT`

packages/cbor/package.json
```diff
@@ -1,31 +1,34 @@
...
-    "jsdoc": "3.5.5",
     "rimraf": "2.6.2",
-    "tsd-jsdoc": "dcodeio/tsd-jsdoc#fe76c4dc071b00bb549d3af7a3bad8c555999e81",
     "webpack": "3.11.0"
   },
   "files": [
-    "dist/",
-    "src/"
+    "dist",
+    "src/demo"
```
bennyn: No need to upload demo code to npmjs

packages/cbor/src/main/cbor/Types.ts
```diff
@@ -0,0 +1,96 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+/* eslint no-magic-numbers: "off" */
+
+class Types {
```
bennyn: Class names should be singular.

packages/cbor/src/main/cbor/Decoder.ts
```diff
@@ -35,12 +35,14 @@ export interface DecoderConfig {
   max_text_length: number;
 }
 
+const MEGABYTE_IN_BYTES = 0x100000;
```
lipis: Sounds a bit confusing TBH.. Isn't the other way around makes more sense (or just `MEGABYTE`)?

(SECONDS_IN_MINUTE, DAYS_IN_WEEK, etc..)

ffflorian: Yeah, you're right. First I wanted to write `ONE_MEGABYTE_IN_BYTES` but it makes more sense the other way round.

PR #383, opened 2018-03-19
fix(core): Register new client on RecordNotFoundError while logging in

Yserz:


packages/core/src/main/Account.ts
```diff
@@ -41,6 +41,7 @@ import {ConversationService} from './conversation/root';
 import Client = require('@wireapp/api-client');
 import EventEmitter = require('events');
 import {StatusCode} from '@wireapp/api-client/dist/commonjs/http/index';
+import {RecordNotFoundError} from '@wireapp/store-engine/dist/commonjs/engine/error';
```
ffflorian:
```js
import {StatusCode} from '@wireapp/api-client/dist/commonjs/http/';
import {RecordNotFoundError} from '@wireapp/store-engine/dist/commonjs/engine/error/';
```
Let's always add a `/` when omitting `/index` (or add `/index`). Otherwise we will get problems on Windows, please ask @bennyn about that. :wink:

PR #393, opened 2018-03-21
feat(store-engine): Add "purge" to delete the database

bennyn:


packages/store-engine/src/main/engine/IndexedDBEngine.ts
```diff
@@ -23,6 +23,14 @@ export default class IndexedDBEngine implements CRUDEngine {
     return Promise.resolve(this.db);
   }
 
+  purge(): Promise<void> {
+    if (this.db) {
+      return this.db.delete();
+    } else {
+      return Dexie.delete(this.storeName);
```
gregor: @bennyn No else needed after return. Our linter used to catch this. Seems like something is missing in the config. In this particular case I would even go as far and change to:

`return this.db() ?  this.db.delete() : Dexie.delete(this.storeName);`

bennyn: We don't have a linter for TypeScript files. That's why it was not catched.

PR #394, opened 2018-03-21
fix(core): Verification cleanup

Yserz:


packages/core/src/main/Account.ts
```diff
@@ -378,9 +378,19 @@ class Account extends EventEmitter {
...
-        if (notFoundInDatabase || notFoundOnBackend) {
+        if (notFoundInDatabase) {
           return this.registerClient(loginData, clientInfo);
         }
+        if (notFoundOnBackend) {
+          const shouldDeleteWholeDatabase = loadedClient.type === ClientType.TEMPORARY;
+          if (shouldDeleteWholeDatabase) {
+            return this.service!.cryptography.purgeDb().then(() => this.registerClient(loginData, clientInfo));
+          } else {
+            return this.service!.cryptography.deleteCryptographyStores().then(() =>
```
gregor: No else after return needed

PR #405, opened 2018-03-22
fix(core): Synchronize clients after client registration

Yserz:


packages/core/src/main/client/ClientDatabaseRepository.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  public createLocalIdentityClient(client: RegisteredClient): Promise<MetaClient> {
+    const transformedClient = this.transformLocalIdentityClient(client);
+    return this.storeEngine
+      .create(ClientDatabaseRepository.STORES.CLIENTS, ClientDatabaseRepository.KEYS.LOCAL_IDENTITY, transformedClient)
+      .then(() => transformedClient);
+  }
+
+  public updateLocalIdentityClient(client: RegisteredClient): Promise<MetaClient> {
```
gregor: It is not really a local identity. We just named the database key at some point. Let's name it `updateLocalClient` instead. And make use of this name everywhere.

packages/core/src/main/client/ClientService.ts
```diff
@@ -0,0 +1,64 @@
...
+  public createLocalIdentityClient(client: RegisteredClient): Promise<MetaClient> {
+    return this.database.createLocalIdentityClient(client);
+  }
+
+  public synchronizeClients() {
+    return this.backend
+      .getClients()
+      .then((registeredClients: RegisteredClient[]) =>
+        registeredClients.filter(client => client.id !== this.apiClient.context!.clientId)
+      )
```
gregor: If it doesn't fit into a single like please do:

````
.then(() => {
  return ....;
})
```

packages/core/src/main/client/ClientService.ts
```diff
@@ -0,0 +1,64 @@
...
+      .then((registeredClients: RegisteredClient[]) =>
+        this.database.createClientList(this.apiClient.context!.userId, registeredClients)
+      );
```
gregor: see above

PR #415, opened 2018-03-26
fix(api-client,store-engine): Open IndexedDB after purge

Yserz:


packages/store-engine/src/main/engine/IndexedDBEngine.test.browser.js
```diff
@@ -25,12 +25,18 @@ describe('StoreEngine.IndexedDBEngine', () => {
...
 
-  async function initEngine() {
-    const storeEngine = new IndexedDBEngine();
+  async function initEngine(shouldCreateNewEngine = true) {
+    let storeEngine;
+    if (shouldCreateNewEngine) {
+      storeEngine = new IndexedDBEngine();
+    } else {
+      storeEngine = engine;
+    }
```
ffflorian:
```js
const storeEngine = shouldCreateNewEngine ? new IndexedDBEngine() : engine;
```

Yserz: done

PR #445, opened 2018-04-03
refactor(api-client): Keep token transient

Yserz:


packages/api-client/src/main/auth/AccessTokenStore.ts
```diff
@@ -31,54 +29,19 @@ class AccessTokenStore extends EventEmitter {
...
-      .read<AccessTokenData>(AUTH_TABLE_NAME, AUTH_ACCESS_TOKEN_KEY)
-      .catch((error: Error) => {
-        if (error.name === RecordNotFoundError.name) {
-          return undefined;
-        }
-
-        throw error;
-      })
-      .then((accessToken: AccessTokenData | undefined) => (this.accessToken = accessToken));
-  }
```
ffflorian: This is a major change, removing this will break a lot of other packages.

Yserz: It's not because the `AccessTokenStore` is private in the `Client`. As long as the api-client itself is not using this function it's not a breaking change (and we never use it in the api-client).

PR #446, opened 2018-04-03
chore(core): Extract registerClient from initClient

Yserz:


packages/core/src/main/Account.ts
```diff
@@ -389,38 +378,41 @@ class Account extends EventEmitter {
...
+      .then(() => {
+        this.logger.info('Client is created');
+        this.apiClient.context!.clientId = registeredClient.id;
+        this.service!.conversation.setClientID(registeredClient.id);
+        return this.service!.notification.initializeNotificationStream(registeredClient.id);
+      })
+      .then(() => this.service!.client.synchronizeClients())
+      .then(() => {
+        return {isNewClient: true, localClient: registeredClient};
+      });
```
gregor: no need for three lines: `.then(() => ({isNewClient: true, localClient: registeredClient}));`

PR #455, opened 2018-04-04
fix(api-client): Remove obsolete initEngine call

Yserz:


packages/api-client/src/main/Client.ts
```diff
@@ -175,14 +175,20 @@ class Client {
...
+        .then(() => this.auth.api.postRegister(userAccount))
+        /** Note:
+         * It's necessary to initialize the context (Client.createContext()) and the store (Client.initEngine())
+         * for saving the retrieved cookie from POST /access (Client.init()) in a Node environment.
+         */
+        .then((user: User) =>
+          this.createContext(user.id, undefined, persist ? ClientType.PERMANENT : ClientType.TEMPORARY)
+        )
+        .then((context: Context) => this.initEngine(context))
+        .then(() => this.init(persist ? ClientType.PERMANENT : ClientType.TEMPORARY))
```
Yserz: Here's the actual fix

PR #457, opened 2018-04-05
chore: Only execute tests for updated projects

Yserz:


bin/testUpdated.js
```diff
@@ -0,0 +1,40 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+const {execSync} = require('child_process');
+
+let output;
+try {
+  output = execSync(`npx lerna updated`);
+} catch (error) {}
```
bennyn: What's the idea behind "swallowing" that error? Should the application stop to work in that case?

PR #458, opened 2018-04-05
refactor: Use catch block

bennyn:


bennyn:
Incorporated into https://github.com/wireapp/wire-web-packages/pull/463

PR #466, opened 2018-04-06
fix(cli-client): Access user context from API client

bennyn:


packages/cli-client/src/main/index.ts
```diff
@@ -71,7 +71,9 @@ storeEngine.init('', {fileExtension: '.json'}).then(() => {
       }
     })
     .then(() =>
-      console.log(`Connected to Wire — User ID "${account.context.userId}" — Client ID "${account.context.clientId}"`)
+      console.log(
+        `Connected to Wire — User ID "${apiClient!.context!.userId}" — Client ID "${apiClient!.context!.clientId}"`
+      )
```
gregor:
````
.then(() => {
  const {clientId, userId} =  account.context!;
  console.log(`Connected to Wire — User ID "${userId}" — Client ID "${clientId}"`);
})
````

bennyn: 👍

PR #492, opened 2018-04-18
feat(core): Send images

ffflorian:


packages/core/src/main/Account.ts
```diff
@@ -299,14 +299,21 @@ class Account extends EventEmitter {
     };
     return Promise.resolve(Root.fromJSON(proto))
       .then((root: Root) => {
+        this.protocolBuffers.Asset = root.lookup('Asset');
         this.protocolBuffers.External = root.lookup('External');
         this.protocolBuffers.GenericMessage = root.lookup('GenericMessage');
         this.protocolBuffers.Text = root.lookup('Text');
       })
```
gregor: Looks like we should move this to it's own function we can put in the chain and rely on {} or Object assign instead of writing the same over and over again.

packages/core/src/test/node/spec/conversation/AssetService.test.js
```diff
@@ -0,0 +1,87 @@
...
+    await engine.init('');
+
+    const client = new APIClient({store: engine, urls: APIClient.BACKEND.STAGING});
+    account = new Account(client);
+    await account.init();
+
+    done();
+  });
+
+  describe("'uploadImageAsset'", () => {
```
bennyn: `describe('"uploadImageAsset"',`

PR #509, opened 2018-04-23
feat(core): Receive images

ffflorian:


packages/api-client/src/main/asset/AssetUtil.ts
```diff
@@ -1,21 +1,22 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
 
 export const isValidKey = (key: string): boolean => /^[A-Za-z0-9-]+$/.test(key);
 
-export const isValidToken = (token: string): boolean => /^[A-Za-z0-9+/=\-]+$/.test(token);
+export const isValidToken = (token: string): boolean => /^[A-Za-z0-9+/=_\-]+$/.test(token);
```
bennyn: Please add a test case with in input value which makes use of that updated RegEx.

PR #514, opened 2018-04-24
refactor(react-ui-kit): Replace "btoa" with "bazinga64"

bennyn:
`btoa` is not available in Node.js, so we use `bazinga64` because we need to run the React UI Kit inside Node.js to be able to create server-side rendered webpages.


ffflorian:
What about using
```js
Buffer.from(invalidDot).toString('base64')
```
?

PR #518, opened 2018-04-25
feat(store-engine): Add "UnsupportedError"

bennyn:


packages/store-engine/src/main/engine/FileEngine.ts
```diff
@@ -12,6 +18,10 @@ export default class FileEngine implements CRUDEngine {
   constructor(private baseDirectory: string = '') {}
 
   init(storeName: string = '', options: {fileExtension: string}): Promise<any> {
+    if (typeof window !== 'undefined') {
```
atomrc: Since I am seeing this test (or it negative version) multiple times, maybe we can export this into a utility function.
Something like `isNodeEnvironment`?

bennyn: Fair point

PR #520, opened 2018-04-25
feat(core): Send and receive confirmations

ffflorian:


packages/core/src/demo/receiver.js
```diff
@@ -15,8 +15,7 @@ const APIClient = require('@wireapp/api-client');
 const fs = require('fs');
 const {promisify} = require('util');
 const {Config} = require('@wireapp/api-client/dist/commonjs/Config');
-const {MemoryEngine} = require('@wireapp/store-engine/dist/commonjs/engine');
-const engine = new MemoryEngine();
+const {FileEngine} = require('@wireapp/store-engine');
```
bennyn: Why switching to `FileEngine`?

packages/core/src/main/Account.ts
```diff
@@ -232,7 +234,13 @@ class Account extends EventEmitter {
...
   }
 
-  private decodeEvent(event: ConversationEvent): Promise<string> {
+  private decodeEvent(
+    event: ConversationEvent
+  ): Promise<{
+    content?: string;
+    id: string;
+    type?: GenericMessageType;
+  }> {
```
bennyn: Let's create a type for:

```js
{
    content?: string;
    id: string;
    type?: GenericMessageType;
}
```

packages/core/src/main/conversation/MessageStatusType.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+enum MessageStatusType {
+  DELIVERED = 3,
+  FAILED = 0,
+  SEEN = 4,
+  SENDING = 1,
+  SENT = 2,
+  UNSPECIFIED = -1,
+}
```
gregor: These are webapp internal types. The protocol buffer supports these https://github.com/wireapp/generic-message-proto/blob/master/proto/messages.proto#L123-L124

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -53,7 +53,33 @@ export default class ConversationService {
     });
   }
 
-  public async sendExternalGenericMessage(
+  public async sendConfirmation(conversationId: string, messageId: string): Promise<ClientMismatch> {
+    const confirmation = this.protocolBuffers.Confirmation.create({
+      type: MessageStatusType.DELIVERED,
```
gregor: Not the right type in this case. See comment below.

packages/core/src/demo/receiver.js
ffflorian: Because after testing the receiver 10 times while developing I created 10 new clients :D

packages/core/src/main/Account.ts
```diff
@@ -248,13 +250,25 @@ class Account extends EventEmitter {
             const genericMessage = this.protocolBuffers.GenericMessage.decode(decryptedMessage);
             switch (genericMessage.content) {
               case GenericMessageType.TEXT: {
-                resolve(genericMessage.text.content);
+                resolve({
+                  content: genericMessage.text.content,
+                  id: genericMessage.messageId,
+                });
```
gregor: Message type is missing from the return value

packages/core/src/main/Account.ts
```diff
@@ -268,22 +282,21 @@ class Account extends EventEmitter {
...
-          } else {
-            this.emit(Account.INCOMING.ASSET, data);
+        if (typeof data.content === 'string') {
+          this.emit(Account.INCOMING.TEXT_MESSAGE, data);
+        } else if (data.type) {
+          switch (data.type) {
+            case GenericMessageType.ASSET:
+              this.emit(Account.INCOMING.ASSET, data);
+            case GenericMessageType.CONFIRMATION:
+              this.emit(Account.INCOMING.CONFIRMATION, data);
```
gregor: Let's use a switch case on the type for all of them

packages/core/src/main/conversation/DecodedEvent.ts
```diff
@@ -0,0 +1,28 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {GenericMessageType} from '../conversation/root';
+
+interface DecodedEvent {
+  content?: string;
+  id: string;
+  type?: GenericMessageType;
```
gregor: I think `type` should never be optional

packages/core/src/main/cryptography/PayloadBundle.ts
```diff
@@ -17,10 +17,14 @@
...
 
+import {GenericMessageType} from '../conversation/root';
+
 interface PayloadBundle {
-  content: string;
+  content?: string;
   conversation: string;
   from: string;
+  id: string;
+  type?: GenericMessageType;
```
gregor: Same here. Type should not be optional.

PR #522, opened 2018-04-26
fix(api-client): Add expires_at to Self type

Yserz:


packages/api-client/src/main/self/Self.ts
```diff
@@ -29,6 +30,7 @@ interface Self {
   name: string;
   phone?: string;
   team?: string;
+  expires_at?: string;
```
bennyn: `expires_at` shpuld be below `email`

Yserz: 😒

PR #534, opened 2018-04-27
refactor(core): Return messageId after sending message

Yserz:


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -144,35 +80,34 @@ export default class ConversationService {
...
+      messageId,
       asset: imageAsset,
     });
 
     const preKeyBundles = await this.getPreKeyBundles(conversationId);
     const plainTextBuffer: Buffer = this.protocolBuffers.GenericMessage.encode(genericMessage).finish();
     const payload: EncryptedAsset = await AssetCryptography.encryptAsset(plainTextBuffer);
 
-    return this.sendExternalGenericMessage(this.clientID, conversationId, payload, <UserPreKeyBundleMap>preKeyBundles);
+    await this.sendExternalGenericMessage(this.clientID, conversationId, payload, <UserPreKeyBundleMap>preKeyBundles);
```
atomrc: We could return the messageId right away. Here it's waiting for the promise to end before returning it.
Is this the intended behavior?

PR #536, opened 2018-04-29
feat(core): Support sending messages to multiple conversations

bennyn:


bennyn:
@gregor The use case is a status bot (like our Wire Webapp Bot or Wire News Bot) which could send the same information to multiple people/conversations. This way we could support something like a broadcast. Supporting sending a message to multiple conversations would also make our core more scalable because we will provide a "new feature" without losing the current feature of sending a message only so a single conversation.

bennyn:
Closing this for now as code became incompatible with the current implementation.

PR #546, opened 2018-05-02
feat: Implement isSupported

Yserz:


gregor:
Let's make sure we handle the exceptions to expect in Chrome and Firefox with disabled cookies and private mode properly.

atomrc:
@gregor @Yserz Sorry I missed this PR, but I would argue that instead of throwing, the method should always return a boolean. 
I believe it would make the API much clearer (as `isSupported` usually has the semantic of "I will give you a boolean"*).

* or a boolean wrapped in a promise for that matter

PR #561, opened 2018-05-07
chore(store-engine): Fail build if test coverage threshold is not met

bennyn:


bennyn:
Thoughts have been incorporated in https://github.com/wireapp/wire-web-packages/pull/638

PR #562, opened 2018-05-07
feat(core): Get and set username

ffflorian:


packages/core/src/main/self/SelfService.ts
```diff
@@ -0,0 +1,47 @@
...
+  public async getName(): Promise<string> {
+    const {name} = await this.apiClient.self.api.getName();
+    return name;
+  }
+
+  public getSelf(): Promise<Self> {
+    return this.apiClient.self.api.getSelf();
+  }
+
+  public async setUserName(userName: string): Promise<{}> {
```
bennyn: Let's also add `getUserName`

packages/core/src/main/self/SelfService.ts
```diff
@@ -0,0 +1,56 @@
...
+    const [availableHandle] = await this.apiClient.user.api.postHandles({
+      handles: [userName],
+    });
+
+    if (availableHandle) {
+      return this.apiClient.self.api.putHandle({handle: userName});
+    }
+
+    throw new Error(`Username "${userName}" is not available.`);
+  }
```
gregor: @ffflorian Why do you check availability when setting? `PUT` on the handle will also fail in this case. No need for the roundtrip before. The check of availability is used for UI purposes so we can show checkmarks and user feedback that the entered handle could be selected and set.

ffflorian: Thanks! Will add a PR for the fix.

ffflorian: https://github.com/wireapp/wire-web-packages/pull/568

PR #566, opened 2018-05-07
lint: Format CHANGELOG.md with Prettier

lipis:


bennyn:
We need @Yserz's input here to automatically format changelogs on new releases with our lerna release process.

lipis:
I think because of the hooks it will happen on it's own! We can give it a shot.. and worst case scenario we can roll back one PR..

PR #568, opened 2018-05-08
fix(core): Additional username checks

ffflorian:


gregor:
I suppose the core should still have the capability of checking the availability of usernames. I would thus propose to not remove but move the `POST` request to an individual call `checkUsernames` which takes an array of strings and a `checkUsername` which takes a string only and the uses the first call.

ffflorian:
But `getName` gets you the display name, not the username - should I rename it to `getDisplayName` then?

bennyn:
I also thought of "getDisplayName" but this can be confusing when working on the UI part because in our message list we had some logic that the app shows just the first name (Benny instead of Benny Neugebauer) but if there are two Bennys, then it shows something like "Benny N" and "Benny B". So I would argue that "Benny N" and "Benny B" are the display names (because that's what you see) but there is still a "name" which is "Benny Neugebauer".

packages/core/src/main/self/SelfService.ts
```diff
@@ -24,33 +24,36 @@ import {Self} from '@wireapp/api-client/dist/commonjs/self/index';
...
-    const [availableHandle] = await this.apiClient.user.api.postHandles({
-      handles: [userName],
-    });
-
-    if (availableHandle) {
-      return this.apiClient.self.api.putHandle({handle: userName});
-    }
-
-    throw new Error(`Username "${userName}" is not available.`);
+  public setUserName(userName: string): Promise<{}> {
```
bennyn: Let's merge wireapp/wire-web-packages#565 first and then we can remove the carp return type. 🎣

PR #571, opened 2018-05-08
fix: Cleanup TSLint rules

lipis:


ffflorian:
Why did you remove exactly these rules?

bennyn:
Is `false` the default of TSLint?

lipis:
Because having them false is like don't declare them at all..

bennyn:
Ok, so `false` seems to be the default for every setting which is nice. 👍

PR #579, opened 2018-05-09
feat(api-client,core): Send and receive typing notifications

ffflorian:


packages/core/src/main/Account.ts
```diff
@@ -238,46 +239,45 @@ class Account extends EventEmitter {
...
-          });
-          break;
-        }
-      }
-    });
+    const sessionId = CryptographyService.constructSessionId(otrMessage.from, otrMessage.data.sender);
+    const ciphertext = otrMessage.data.text;
+    const decryptedMessage = await this.service.cryptography.decrypt(sessionId, ciphertext);
+    const genericMessage = this.protocolBuffers.GenericMessage.decode(decryptedMessage);
+    console.log('genericMessage', genericMessage);
```
lipis: do we allow `console.log` around here?

ffflorian: No, will remove it.

packages/core/src/demo/receiver.js
```diff
@@ -59,6 +59,20 @@ const {MemoryEngine} = require('@wireapp/store-engine/dist/commonjs/engine');
     await account.service.conversation.sendPing(conversationId);
   });
 
+  account.on(Account.INCOMING.TYPING, async data => {
+    const {
+      conversation: conversationId,
+      from,
+      data: {status},
+    } = data;
+    console.log(`Typing in "${conversationId}" from "${from}".`, data);
```
lipis: there are more :)

packages/core/src/demo/receiver.js
```diff
@@ -59,6 +59,20 @@ const {MemoryEngine} = require('@wireapp/store-engine/dist/commonjs/engine');
...
+    if (status === 'started') {
+      await account.service.conversation.sendTypingStart(conversationId);
+    } else {
+      await account.service.conversation.sendTypingStop(conversationId);
+    }
+  });
+
   try {
     console.log('Logging in ...');
```
lipis: not part of this PR.. but maybe it should go too..

ffflorian: No, this is fine for the demo.

packages/core/src/demo/receiver.js
ffflorian: Fine for demo

PR #593, opened 2018-05-14
chore: Disallow unused variables and imports using TSLint

bennyn:


packages/react-ui-kit/src/Form/Form.jsx
```diff
@@ -20,7 +20,7 @@
 import styled from 'styled-components';
 
 const Form = styled.form`
-  /* appearance */
+  /* n */
```
lipis: n

bennyn: Good eye sight!

PR #611, opened 2018-05-22
refactor: Rename npm script "clear" to "clean"

bennyn:
We have a task that handles project cleaning which is named `clear` but it is common practice to name that task `clean`. 

Here are some references to confirm my statement:
- https://www.gnu.org/software/make/manual/html_node/Cleanup.html (`make`, C/C++)
- https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html (`mvn`, Java)
- https://t3n.de/news/npm-build-tool-612400/2/ (`npm`, JavaScript)


Yserz:
I remember that we did it that way because at some point it was reserved from yarn. Is that resolved?

package.json
```diff
@@ -30,8 +30,8 @@
   "private": true,
   "scripts": {
     "boot": "lerna bootstrap --reject-cycles",
-    "clear": "lerna run clear",
-    "dist": "yarn clear && lerna run dist",
+    "clean": "lerna run clean",
+    "dist": "yarn clean && lerna run dist",
```
chenrui333: I think yarn does not have `clean` command anymore, you may want to update to use `autoclean`

PR #623, opened 2018-05-25
feat: Commons package

Yserz:


packages/commons/src/main/index.ts
```diff
@@ -0,0 +1,25 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import * as CommonConfig from './config/CommonConfig';
+import * as RandomUtil from './util/RandomUtil';
+import * as StringUtil from './util/StringUtil';
+import * as UrlUtil from './util/UrlUtil';
+
+export = {RandomUtil, StringUtil, UrlUtil, CommonConfig};
```
atomrc: Why do we use this notation (`export =`) and not just `export {...}` for this file?

Yserz: Uh, thx

PR #649, opened 2018-06-01
feat(proteus): Add error code 214 for QA purposes

bennyn:


bennyn:
Not needed anymore because fix was made in "wire-web-ets" by receiving messages with `await account.listen()`. PR: https://github.com/wireapp/wire-web-ets/pull/45

PR #660, opened 2018-06-04
refactor(api-client,core): ClientType NONE

Yserz:


packages/api-client/src/main/Client.ts
```diff
@@ -162,19 +162,15 @@ class Client {
       .then((response: AxiosResponse<any>) => {
         cookieResponse = response;
         accessToken = response.data;
-        context = this.createContext(
-          accessToken.user,
-          undefined,
-          loginData.persist ? ClientType.PERMANENT : ClientType.TEMPORARY
-        );
+        context = this.createContext(accessToken.user, loginData.clientType, undefined);
```
AndyLnd: ne need for an explicit ```undefined``` here

packages/api-client/src/main/Client.ts
```diff
@@ -137,13 +137,13 @@ class Client {
...
 
-  public init(clientType?: ClientType): Promise<Context> {
+  public init(clientType: ClientType = ClientType.NONE): Promise<Context> {
     let context: Context;
     let accessToken: AccessTokenData;
     return this.transport.http
       .postAccess()
       .then((createdAccessToken: AccessTokenData) => {
-        context = this.createContext(createdAccessToken.user, undefined, clientType);
+        context = this.createContext(createdAccessToken.user, clientType, undefined);
```
AndyLnd: no need for an explicit ```undefined``` here

packages/api-client/src/main/Client.ts
```diff
@@ -184,11 +180,9 @@ class Client {
          * It's necessary to initialize the context (Client.createContext()) and the store (Client.initEngine())
          * for saving the retrieved cookie from POST /access (Client.init()) in a Node environment.
          */
-        .then((user: User) =>
-          this.createContext(user.id, undefined, persist ? ClientType.PERMANENT : ClientType.TEMPORARY)
-        )
+        .then((user: User) => this.createContext(user.id, clientType, undefined))
```
AndyLnd: see above

PR #684, opened 2018-06-08
feat(cryptobox): Serialization & Deserialization of cryptographic items

bennyn:
Cryptobox now has a `serialize` and a `deserialize` function which makes it easy to export a Cryptobox which later than can be imported into another instance.

I wrote a test case which shows that a serialized Cryptobox can be loaded by another Cryptobox instance to prove the compatibility.

In my test case there is Alice, Bob and Eve. Alice writes to Bob and to Eve. Then Alice exports her data and Eve will load (deserialize) the data of Alice. This gives Eve the possibility to continue messaging to Bob because Eve became Alice when deserializing Alice's data.


packages/cryptobox/src/main/Cryptobox.ts
```diff
@@ -418,6 +419,85 @@ class Cryptobox extends EventEmitter {
...
+  private async importSessions(payload: string[]): Promise<void> {
+    this.logger.log(`Importing "${payload.length}" sessions...`);
+
+    const proteusSessions = payload.map(session => {
+      const sessionBuffer = Decoder.fromBase64(session).asBytes.buffer;
+      return ProteusSession.Session.deserialise(this.identity!, sessionBuffer);
+    });
+
+    const saveSessions = proteusSessions.map(session => {
+      const cryptoBoxSession = new CryptoboxSession(undefined, this.store, session);
```
ffflorian: For better readability:
```js
const saveSessions = proteusSessions.map(proteusSession => {
  const cryptoBoxSession = new CryptoboxSession(undefined, this.store, proteusSession);
```

packages/cryptobox/src/main/Cryptobox.ts
```diff
@@ -418,6 +419,85 @@ class Cryptobox extends EventEmitter {
...
+  public async deserialize(payload: SerializedCryptobox) {
+    await this.deleteData();
+    await this.importIdentity(payload.identity);
+    await this.importPreKeys(payload.prekeys);
+    await this.importSessions(payload.sessions);
+    await this.refill_prekeys(true);
+  }
+
+  public async serialize(): Promise<SerializedCryptobox> {
+    const toBase64 = (array: ArrayBuffer) => Encoder.toBase64(new Uint8Array(array)).asString;
```
ffflorian: Why do we need `new Uint8Array(array)`? bazinga64's `toBase64()` [also accepts `ArrayBuffer`](https://github.com/wireapp/wire-web-packages/blob/master/packages/bazinga64/src/main/bazinga64.ts#L195).

packages/cryptobox/src/main/Cryptobox.ts
```diff
@@ -418,6 +419,85 @@ class Cryptobox extends EventEmitter {
...
+    const identity = await this.store.load_identity();
+
+    if (identity) {
+      data.identity = toBase64(identity.serialise());
+      const sessions = await this.store.read_sessions(identity);
+      data.sessions = sessions.map(session => toBase64(session.serialise()));
+    }
+
+    const prekeys = await this.store.load_prekeys();
+    data.prekeys = prekeys.map(prekey => toBase64(prekey.serialise()));
```
ffflorian: For better readability:
```js
const storedPrekeys = await this.store.load_prekeys();
  data.prekeys = storedPrekeys.map(storedPrekey => toBase64(storedPrekey.serialise()));
```

packages/cryptobox/src/main/Cryptobox.ts
```diff
@@ -418,6 +419,85 @@ class Cryptobox extends EventEmitter {
...
+      sessions: [],
+    };
+
+    const identity = await this.store.load_identity();
+
+    if (identity) {
+      data.identity = toBase64(identity.serialise());
+      const sessions = await this.store.read_sessions(identity);
+      data.sessions = sessions.map(session => toBase64(session.serialise()));
+    }
```
ffflorian: For better readability:
```js
const storedIdentity = await this.store.load_identity();

 if (storedIdentity) {
  data.identity = toBase64(storedIdentity.serialise());
  const storedSessions = await this.store.read_sessions(storedIdentity);
  data.sessions = storedSessions.map(storedSession => toBase64(storedSession.serialise()));
}
```

packages/cryptobox/src/main/CryptoboxSession.ts
```diff
@@ -26,10 +26,10 @@ class CryptoboxSession {
   public pk_store: CryptoboxCRUDStore;
   public session: ProteusSession.Session;
 
-  constructor(id: string, pk_store: CryptoboxCRUDStore, session: ProteusSession.Session) {
-    this.id = id;
+  constructor(id: string | undefined, pk_store: CryptoboxCRUDStore, session: ProteusSession.Session) {
```
ffflorian: Shouldn't we set `id` as last parameter if it's optional?

packages/cryptobox/src/main/CryptoboxSession.ts
```diff
@@ -48,6 +48,12 @@ class CryptoboxSession {
     });
   }
 
+  public generateSessionId(): string {
+    const from = this.fingerprint_local();
+    const to = this.fingerprint_remote();
+    return `${from}@${to}`;
```
ffflorian: For better readability:
```js
return from + '@' + to;
```

packages/cryptobox/src/main/store/CryptoboxCRUDStore.ts
```diff
@@ -154,6 +153,20 @@ class CryptoboxCRUDStore implements ProteusSession.PreKeyStore {
       });
   }
 
+  public read_sessions(identity: ProteusKeys.IdentityKeyPair): Promise<Array<ProteusSession.Session>> {
```
ffflorian:
```ts
public read_sessions(identity: ProteusKeys.IdentityKeyPair): Promise<ProteusSession.Session[]> {
```

packages/cryptobox/src/main/store/CryptoboxCRUDStore.ts
```diff
@@ -154,6 +153,20 @@ class CryptoboxCRUDStore implements ProteusSession.PreKeyStore {
...
+    return this.engine.readAll(CryptoboxCRUDStore.STORES.SESSIONS).then((records: Array<any>) => {
```
ffflorian:
```ts
return this.engine.readAll(CryptoboxCRUDStore.STORES.SESSIONS).then((records: PersistedRecord[]) => {
```

packages/cryptobox/test/CryptoboxSpec.node.js
```diff
@@ -73,4 +73,75 @@ describe('Cryptobox', () => {
...
+      const eveSessions = await eve.store.read_sessions(eve.identity);
+      expect(eveSessions.length).toBe(expectedSessionsOfAlice);
+
+      // Test that Eve's Cryptobox can be serialized
+      const serializedEve = await eve.serialize();
+      expect(Object.keys(serializedEve.prekeys).length).toBe(amountOfAlicePreKeys);
+
+      // Test that Eve can write to Bob because Alice had a session with Bob
+      const bobId = bob.identity.public_key.fingerprint();
+      const sessionId = `${aliceId}@${bobId}`;
```
ffflorian:
```js
const sessionId = aliceId + '@' + bobId;
```

packages/cryptobox/src/main/CryptoboxSession.ts
bennyn: Yes, we should but I didn't want to break the API.

packages/cryptobox/src/main/CryptoboxSession.ts
bennyn: We devided to prefer string templates.

packages/cryptobox/test/CryptoboxSpec.node.js
bennyn: A good IDEA supports colors, so string templates are easy to read:

image

packages/cryptobox/src/main/store/CryptoboxCRUDStore.ts
bennyn: We tried to this in the past but we received:

> src/main/store/CryptoboxCRUDStore.ts:157:73 - error TS2345: Argument of type '(records: PersistedRecord[]) => Session[]' is not assignable to parameter of type '((
value: {}[]) => Session[] | PromiseLike<Session[]>) | null | undefined'.
  Type '(records: PersistedRecord[]) => Session[]' is not assignable to type '(value: {}[]) => Session[] | PromiseLike<Session[]>'.
    Types of parameters 'records' and 'value' are incompatible.
      Type '{}[]' is not assignable to type 'PersistedRecord[]'.
157     return this.engine.readAll(CryptoboxCRUDStore.STORES.SESSIONS).then((records: PersistedRecord[]) => {
                                                                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

packages/cryptobox/src/main/CryptoboxSession.ts
bennyn: But I will do a refactoring. Otherwise this will never change. 😎  👍

bennyn: It gave me some very nice ideas so I refactored it. Thanks Florian!!

PR #696, opened 2018-06-13
fix(api-client): Add optional expires_at to User entity

Yserz:


bennyn:
> fix: Add optional expires_at to User entity

->

> fix(api-client): Add optional expires_at to User entity

PR #697, opened 2018-06-13
test(cryptobox): Accept PreKey message even if there is already a session

bennyn:


packages/cryptobox/src/main/Cryptobox.ts
```diff
@@ -158,6 +158,16 @@ class Cryptobox extends EventEmitter {
     return this.store.load_prekey(prekey_id);
   }
 
+  public async get_prekey_bundle(
+    prekey_id: number = ProteusKeys.PreKey.MAX_PREKEY_ID
+  ): Promise<ProteusKeys.PreKeyBundle | undefined> {
+    const preKey = await this.get_prekey(prekey_id);
+    if (this.identity && preKey) {
```
atomrc: Shouldn't we handle the case where `identity` is not set differently than the case where there is no prekey?

The `identity` problem seems to be linked to a temporal coupling which is always hard to debug. Maybe throwing an error in that case would be a good idea to ease debugging.

bennyn: `identity` can only be `undefined` in this case when you call that method before you call `init`. But yeah... I can rethrow that error. Also thought about that. Good point.

PR #698, opened 2018-06-13
feat(core): Send and receive session reset messages

bennyn:


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -68,6 +68,16 @@ export default class ConversationService {
     });
   }
 
+  public async sendSessionReset(conversationId: string) {
+    const messageId = new UUID(4).format();
+    const sessionReset = this.protocolBuffers.GenericMessage.create({
+      clientAction: ClientAction.RESET_SESSION,
+    });
```
ffflorian: `messageId` is missing in `sessionReset`.

packages/core/src/main/Account.ts
```diff
@@ -260,8 +262,10 @@ class Account extends EventEmitter {
 
     switch (event.type) {
       case CONVERSATION_EVENT.OTR_MESSAGE_ADD: {
-        const decodedMessage = await this.decodeGenericMessage(event as ConversationOtrMessageAddEvent);
-        return {...decodedMessage, from, conversation};
+        const otrMessage = event as ConversationOtrMessageAddEvent;
+        const sessionId = CryptographyService.constructSessionId(from, otrMessage.data.sender);
+        const decodedMessage = await this.decodeGenericMessage(otrMessage);
+        return {...decodedMessage, from, conversation, sessionId};
```
ffflorian: Returning `sessionId` is not needed anymore, is it?

bennyn: It is, because it is now part of the contract of `PayloadBundle`.

PR #700, opened 2018-06-13
fix(react-ui-kit): Grid gutter updates

lipis:
The master version is visible here:
https://wire-react-ui-kit.netlify.com

and each pull request you can check from the Netlify check below. It should look something like this:

https://deploy-preview-700--wire-react-ui-kit.netlify.com


bennyn:
> Grid gutter updates

->

> style(react-ui-kit): Update Gutter values

PR #715, opened 2018-06-15
fix(store-engine): Allow dots in directory names

bennyn:


packages/store-engine/src/main/engine/FileEngine.ts
```diff
@@ -12,6 +12,7 @@ import {
 
 export default class FileEngine implements CRUDEngine {
   public storeName: string = '';
+  public storeNameOriginal: string = '';
```
bennyn: Instead of having both (`storeName` & `storeNameOriginal`) we shoul decide on the one to keep because I doubt that we really need to keep both.

packages/store-engine/src/main/engine/FileEngine.ts
```diff
@@ -36,32 +38,36 @@ export default class FileEngine implements CRUDEngine {
...
-          typeof testPath !== 'undefined' &&
-          (testPath.includes('.') || testPath.includes('/') || testPath.includes('\\'))
-        ) {
-          return true;
-        }
-      }
-      return false;
-    };
+  static enforcePathRestrictions(givenTrustedRoot: string, givenPath: string, forceWindows: boolean = false): void {
+    const pathApi: any = process.platform === 'win32' || forceWindows === true ? path.win32 : path;
```
bennyn: `any` is not a good thing to have

packages/store-engine/src/main/engine/FileEngine.ts
```diff
@@ -36,32 +38,36 @@ export default class FileEngine implements CRUDEngine {
...
+    }
+  }
+
+  private resolvePath(tableName: string, primaryKey: string = ''): Promise<string> {
     return new Promise((resolve, reject) => {
-      if (isPathTraversal(tableName, primaryKey || '')) {
-        const message = `Path traversal has been detected on "${path.join(tableName, String(primaryKey))}".`;
-        return reject(new PathValidationError(message));
-      }
+      FileEngine.enforcePathRestrictions(this.baseDirectory, this.storeNameOriginal);
```
bennyn: Executing `FileEngine.enforcePathRestrictions` three times on every `resolvePath` sounds too much to me. Keep in mind that `resolvePath` is executed on every CRUD operation. So every "read" will cause such excessive checks which will make our code slow. I think we can come up with something more performant.

packages/store-engine/src/main/engine/FileEngine.test.node.js
```diff
@@ -45,23 +45,56 @@ describe('FileEngine', () => {
...
+    });
+
+    it('allows empty strings.', () => {
+      const tableName = 'amplify';
+      const primaryKey = '';
+      const actual = FileEngine.enforcePathRestrictions(path.join(unixFolder, tableName), primaryKey);
+      expect(actual).toBeUndefined();
+    });
+
+    it('allows navigation within the restricted folder.', () => {
```
bennyn: Why should we allow navigation within the restriced folder? Do we have a use-case for that?

pwnsdx: We can now have tables and primary keys that contains dots but also slashes now.

bennyn: 👍

bennyn: We already have the `it('allows dots inside of primary keys.')` test case. We should add another one for `it('allows slashes inside of primary keys.')` to highlight this use case. I think its a really important one. Thanks for bringing that up, Sabri!

PR #719, opened 2018-06-18
feat(core): Ephemeral messages

ffflorian:


packages/core/src/demo/sender.js
```diff
@@ -27,16 +27,19 @@ const {FileEngine} = require('@wireapp/store-engine');
 
   const engine = new FileEngine(path.join(__dirname, '.tmp', 'sender'));
   await engine.init(undefined, {fileExtension: '.json'});
-  const apiClient = new APIClient(new Config(engine, APIClient.BACKEND.PRODUCTION));
+  const apiClient = new APIClient(new Config(engine, APIClient.BACKEND.STAGING));
```
bennyn: Let's prefer `APIClient.BACKEND.PRODUCTION` because external reviewers don't have a staging account by default.

PR #722, opened 2018-06-18
feat(core): Delete messages

ffflorian:


packages/core/src/demo/sender.js
```diff
@@ -34,22 +41,42 @@ const {FileEngine} = require('@wireapp/store-engine');
   await account.login(login);
   await account.listen();
 
+  const name = await account.service.self.getName();
+
+  logger.log('Name:', name);
```
bennyn: We should either do `logger.log('Name', name);` or 

```
logger.log(`Name: ${name}`);
```

PR #725, opened 2018-06-18
 refactor(core): Implement global send method

ffflorian:


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -350,10 +277,131 @@ export default class ConversationService {
...
+  }
+
+  public async send(
+    conversationId: string,
+    payloadBundle: PayloadBundleOutgoingUnsent
+  ): Promise<PayloadBundleOutgoing> {
+    switch (payloadBundle.type) {
+      case GenericMessageType.ASSET: {
+        if (payloadBundle.content) {
+          const ctn = payloadBundle.content as ImageAsset;
```
bennyn: Can't we use `context` instead of `ctn` here?

PR #726, opened 2018-06-18
test(store-engine): Add path validation tests

bennyn:


packages/store-engine/src/main/engine/FileEngine.test.node.js
```diff
@@ -82,30 +82,54 @@ describe('FileEngine', () => {
...
+        'readAll',
+        'readAllPrimaryKeys',
+        'update',
+        'updateOrCreate',
+      ];
+
+      for (operation of functionNames) {
+        try {
+          await engine[operation]('../etc', 'primary-key', {});
+          done.fail('Expected error');
```
ffflorian:
```js
done.fail(`Expected error "${expectedError}"`);
```

PR #731, opened 2018-06-19
feat(core): Support timed messages in send method

bennyn:


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -370,13 +375,14 @@ export default class ConversationService {
 
   public async send(
     conversationId: string,
-    payloadBundle: PayloadBundleOutgoingUnsent
+    payloadBundle: PayloadBundleOutgoingUnsent,
+    expireAfterMillis?: number
```
Yserz: Should we check for negative/too big values?

PR #735, opened 2018-06-20
feat(api-client): Message timer update

ffflorian:


packages/api-client/src/main/conversation/ConversationMessageTimer.ts
```diff
@@ -0,0 +1,24 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+interface ConversationMessageTimer {
```
atomrc: Not sure about the naming here. 
Since this is the payload for the update request (not actually used in business code), I would rather call it `ConversationMessageTimerUpdatePayload` 
WDYT?

ffflorian: It's a bit too long, I renamed it to `ConversationMessageTimerUpdate`, keeping it in line with the spec.

PR #759, opened 2018-06-25
feat(react-ui-kit): Default to no icon if unset

Yserz:


gregor:
This turned out to be a breaking change. Let's be a little more cautious with semantic versioning and reviews. It changed existing and used behavior.

PR #765, opened 2018-06-25
fix(core): Set 'from' to selfId when sending delete message

Yserz:


ffflorian:
What about `messageId`? Do you want to change it to `deletedMessageId`?

Yserz:
I'm not sure about that currently

PR #775, opened 2018-06-28
feat(changelog-bot): Add repo slug to message header

bennyn:


packages/changelog-bot/src/main/start.ts
```diff
@@ -31,10 +32,12 @@ export async function start(parameters: {[index: string]: string}): Promise<Chan
     password,
   };
 
-  const changelog = await ChangelogBot.generateChangelog(String(TRAVIS_REPO_SLUG), String(TRAVIS_COMMIT_RANGE));
+  const repoSlug = String(TRAVIS_REPO_SLUG);
```
atomrc: Do you really need to wrap it in an object? There are very few (if any) scenarios where it's needed

PR #783, opened 2018-06-29
chore(api-client): Increase upper bound for notification fetch size

bennyn:


packages/api-client/src/main/notification/NotificationAPI.ts
```diff
@@ -53,7 +53,7 @@ class NotificationAPI {
    */
   public getNotifications(
     clientId?: string,
-    size: number = 1000,
+    size: number = 10000,
```
gregor: Let's not set the default value to equal the max value.

bennyn: As discussed, streamlining this with our web app implementation: https://github.com/wireapp/wire-webapp/blob/2018-06-19-08-04/app/script/event/EventRepository.js#L213

packages/api-client/src/main/notification/NotificationAPI.ts
```diff
@@ -53,7 +55,7 @@ class NotificationAPI {
    */
   public getNotifications(
     clientId?: string,
-    size: number = 10000,
+    size: number = NOTIFICATION_SIZE_MAXIMUM,
```
bennyn: This one is for you @lipis. ❤️

PR #789, opened 2018-07-02
test(core): Initial nock setup

ffflorian:


packages/core/src/main/Account.test.node.js
```diff
@@ -41,4 +53,115 @@ describe('Account', () => {
...
+          new RegExp(
+            `${ConversationAPI.URL.CONVERSATIONS}/.*/${ConversationAPI.URL.OTR}/${ConversationAPI.URL.MESSAGES}`
+          )
+        )
+        .query({ignore_missing: false})
+        .reply(StatusCode.OK)
+        .persist();
+
+      nock(BASE_URL_HTTPS)
+        .get(NotificationAPI.URL.NOTIFICATION + '/' + NotificationAPI.URL.LAST)
```
bennyn: string template can be used here

packages/core/src/main/Account.test.node.js
```diff
@@ -41,4 +53,115 @@ describe('Account', () => {
...
+        const apiClient = new APIClient(config);
+        const account = new Account(apiClient);
+
+        await account.init();
+        await account.login({
+          clientType: ClientType.TEMPORARY,
+          email: 'hello@example.com',
+          password: 'my-secret',
+        });
+
```
bennyn: we need to define expectations

packages/core/src/main/Account.test.node.js
```diff
@@ -41,4 +53,115 @@ describe('Account', () => {
...
+        await account.init();
+        await account.login({
+          clientType: ClientType.TEMPORARY,
+          email: 'hello@example.com',
+          password: 'wrong',
+        });
+
+        done.fail('Should not be logged in');
+      } catch (error) {
+        done();
```
bennyn: let's do an assertion on the `error` object

packages/core/src/main/Account.test.node.js
```diff
@@ -41,4 +53,115 @@ describe('Account', () => {
...
+          password: 'my-secret',
+        });
+
+        done();
+      } catch (error) {
+        done.fail(error);
+      }
+    });
+
+    fit('fails with incorrect credentials', async done => {
```
bennyn: `fit`

packages/core/src/main/Account.test.node.js
```diff
@@ -41,4 +53,115 @@ describe('Account', () => {
...
+        .get(NotificationAPI.URL.NOTIFICATION + '/' + NotificationAPI.URL.LAST)
+        .query({client: clientId})
+        .reply(StatusCode.OK, {});
+
+      nock(BASE_URL_HTTPS)
+        .get(ClientAPI.URL.CLIENTS)
+        .reply(StatusCode.OK, [{id: clientId}]);
+    });
+
+    fit('logs in with correct credentials', async done => {
```
bennyn: `fit`

PR #803, opened 2018-07-05
feat(core): Add and remove conversation participants

bennyn:
This PR adds the functionality in `@wireapp/core` to add and remove conversation participants but it also adds a concept for mapping backend error messages because in the future we are not going to work with plain `AxiosError` objects.


packages/api-client/src/main/http/BackendErrorMapper.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ */
+
+import {UnknownConversationError} from '../conversation/';
+import {BackendError, BackendErrorLabel, StatusCode} from '../http/';
+import {UnknownUserError} from '../user/';
+
+class BackendErrorMapper {
+  public static mapBackendError(error: BackendError): Error {
+    const errors: {
```
atomrc: We could probably extract this config object outside the scope of the method, so that it's not created at every call of the function

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -475,6 +475,21 @@ export default class ConversationService {
     });
   }
 
+  public async addUser(conversationId: string, userId: string): Promise<string[]>;
+  public async addUser(conversationId: string, userIds: string[]): Promise<string[]>;
+  public async addUser(conversationId: string, userIds: string | string[]): Promise<string[]> {
```
gregor: Isn't this a duplication?

ffflorian: Usually that's what [function overloading](https://www.typescriptlang.org/docs/handbook/functions.html) looks like but in this case it's duplication because they all return the same.

packages/api-client/src/main/conversation/ConversationAPI.ts
```diff
@@ -451,6 +451,24 @@ class ConversationAPI {
...
+  public async postMembers(conversationId: string, userIds: string[]) {
+    const config: AxiosRequestConfig = {
+      data: {
+        users: userIds,
+      },
+      method: 'put',
+      url: `${ConversationAPI.URL.CONVERSATIONS}/${conversationId}/${ConversationAPI.URL.MEMBERS}`,
+    };
+
+    await this.client.sendJSON(config);
```
ffflorian: Why not return the result like in all the other functions?

packages/api-client/src/main/conversation/ConversationError.ts
```diff
@@ -0,0 +1,37 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export class ConversationError extends Error {
+  public name: string = 'ConversationError';
```
ffflorian:
```ts
public name = 'ConversationError';
```

Doesn't need to be defined as `string` since it's obvious.

packages/api-client/src/main/conversation/ConversationError.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  constructor(message: string) {
+    super(message);
+    Object.setPrototypeOf(this, ConversationError.prototype);
+    this.stack = new Error().stack;
+  }
+}
+
+export class UnknownConversationError extends ConversationError {
+  public name: string = 'UnknownConversationError';
```
ffflorian: Same here

packages/api-client/src/main/user/UserError.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  constructor(message: string) {
+    super(message);
+    Object.setPrototypeOf(this, UserError.prototype);
+    this.stack = new Error().stack;
+  }
+}
+
+export class UnknownUserError extends UserError {
+  public name: string = 'UnknownUserError';
```
ffflorian: `string` is obvious.

packages/api-client/src/main/user/UserError.ts
```diff
@@ -0,0 +1,37 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export class UserError extends Error {
+  public name: string = 'UserError';
```
ffflorian: `string` is obvious.

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -474,6 +475,33 @@ export default class ConversationService {
...
+    return this.apiClient.conversation.api
+      .postMembers(conversationId, ids)
+      .then(() => ids)
+      .catch((error: AxiosError) => {
+        if (error && error.response) {
+          throw BackendErrorMapper.map(error.response.data as BackendError);
+        }
+        throw error;
+      });
+  }
```
ffflorian: The function is `async`, so please also use `await` :slightly_smiling_face: 
```ts
try {
  await this.apiClient.conversation.api.postMembers(conversationId, ids);
  return ids;
} catch(error: AxiosError) => {
  if (error && error.response) {
    throw BackendErrorMapper.map(error.response.data as BackendError);
  }
  throw error;
}

File: packages/core/src/main/conversation/ConversationService.ts [hunk 9]
```diff
@@ -474,6 +475,33 @@ export default class ConversationService {
...
+  public async removeUser(conversationId: string, userId: string): Promise<string> {
+    return this.apiClient.conversation.api
+      .deleteMember(conversationId, userId)
+      .then(() => userId)
+      .catch((error: AxiosError) => {
+        if (error && error.response) {
+          throw BackendErrorMapper.map(error.response.data as BackendError);
+        }
+        throw error;
+      });
```
ffflorian: Function is `async`, too

packages/core/src/main/conversation/ConversationService.ts
ffflorian: So yes, I suggest to remove the overloading.

bennyn: Ok, I will make `addUser` return just a single `string` then.

packages/api-client/src/main/conversation/ConversationAPI.ts
```diff
@@ -451,6 +451,24 @@ class ConversationAPI {
     return this.client.sendJSON(config).then((response: AxiosResponse) => response.data);
   }
 
+  /**
+   * Add users to an existing conversation.
+   * @param conversationId The conversation ID to add the users to
+   * @param userIds List of user IDs to add to a conversation
+   * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/conversations/addMembers
+   */
+  public async postMembers(conversationId: string, userIds: string[]): Promise<ConversationMemberJoinEvent> {
```
ffflorian: Doesn't need to be async.

packages/core/src/main/conversation/ConversationService.ts
bennyn: async/await is not always the most readable option: https://vote4code.com/5646874153320448/asyncawait

PR #810, opened 2018-07-06
feat(core): Add "timestamp" to payload bundle

bennyn:


bennyn:
We will use `ms` instead of ISO strings.

PR #814, opened 2018-07-09
feat(bot-api): Add Bot API (WEBFOUND-13)

bennyn:
Building an echo bot is now as easy as this:

```bash
yarn add @wireapp/bot-api
```

This package should become the base layer for our "changelog-bot" and our "travis-bot". In the future we will exchange the current bot authentication (at the moment they are normal users) with our bot API authentication mechanism (https://github.com/wireapp/bot-sdk-node).

**index.ts**

```ts
const bot = new Bot({
  email: String(process.env.WIRE_EMAIL), 
  password: String(process.env.WIRE_PASSWORD)
});

bot.addHandler(new EchoHandler());

bot.start();
```

**EchoHandler.ts**

```ts
class EchoHandler extends MessageHandler {
  async handleEvent(payload: PayloadBundleIncoming): Promise<void> {
    if (payload.type === PayloadBundleType.TEXT) {
      const content = payload.content as TextContent;
      await this.sendText(payload.conversation, content.text);
    }
  }
}

export {EchoHandler};
```


packages/bot-api/src/main/EchoHandler.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {MessageHandler} from './MessageHandler';
+
+class EchoHandler extends MessageHandler {
+  async handleText(conversationId: string, _: string, text: string): Promise<void> {
```
ffflorian: Why `_:`?

bennyn: `noUnusedParameters` legacy 🙈

packages/bot-api/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "build:browser": "exit 0",
+    "build:node": "tsc",
+    "clean": "rimraf dist",
+    "dist": "yarn clean && yarn build",
+    "test": "yarn test:node && yarn test:browser",
+    "test:browser": "exit 0",
+    "test:node": "exit 0",
+    "test:project": "yarn dist && yarn test"
+  },
+  "version": "0.1.0-beta.9"
```
ffflorian: Will we keep the beta tag?

packages/bot-api/src/main/Bot.ts
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {APIClient} from '@wireapp/api-client';
+import {ClientType} from '@wireapp/api-client/dist/commonjs/client/';
+import {Config} from '@wireapp/api-client/dist/commonjs/Config';
+import {Connection, ConnectionStatus} from '@wireapp/api-client/dist/commonjs/connection';
```
ffflorian: Missing `/` at the end

packages/bot-api/src/main/Bot.ts
```diff
@@ -0,0 +1,123 @@
...
+import {Account} from '@wireapp/core';
+import {PayloadBundleIncoming, PayloadBundleType} from '@wireapp/core/dist/conversation/root';
+import {MemoryEngine} from '@wireapp/store-engine';
+import UUID from 'pure-uuid';
+import {BotConfig} from './BotConfig';
+import {MessageHandler} from './MessageHandler';
+
+const logdown = require('logdown');
```
ffflorian: Can be imported with
```ts
import * as logdown from 'logdown';
```

packages/bot-api/src/main/Bot.ts
```diff
@@ -0,0 +1,123 @@
...
+            this.logger.error(`An error occurred during connection request handling: ${error.message}`, error);
+          }
+        }
+      }
+    });
+    await this.account.login(login);
+    await this.account.listen();
+    this.handlers.forEach(handler => (handler.account = this.account));
+    return true;
+  }
```
ffflorian: This means the bot API can handle only `PayloadBundleType.TEXT` and `PayloadBundleType.CONNECTION_REQUEST`?

bennyn: You are right, we should think of a generic way to get all the events possible.

packages/bot-api/src/main/Bot.ts
bennyn: I will add it but since `forceConsistentCasingInFileNames` is turned on in our `tsconfig.json` file we don't need to pay attention to it anymore. 😃

packages/bot-api/src/main/Bot.ts
```diff
@@ -0,0 +1,140 @@
...
+import {BotConfig} from './BotConfig';
+import {MessageHandler} from './MessageHandler';
+
+class Bot {
+  public account: Account | undefined;
+
+  private readonly credentials: {email: string; password: string};
+  private readonly config: BotConfig;
+  private readonly handlers: Map<string, MessageHandler>;
+  private readonly logger: any = logdown('@wireapp/standup-bot/StandupBot', {
```
ffflorian: Doesn't need to be `any`.

packages/bot-api/src/main/Bot.ts
```diff
@@ -0,0 +1,140 @@
...
+        if (this.validateMessage(String(conversation), userId)) {
+          this.logger.info('Processing connection request ...');
+          try {
+            this.handlers.forEach(handler => handler.handleEvent(payload));
+          } catch (error) {
+            this.logger.error(`An error occurred during connection request handling: ${error.message}`, error);
+          }
+        }
+      }
+    });
```
ffflorian: Why is this not handled by `handlePayload`?

PR #822, opened 2018-07-11
fix(react-ui-kit): Pass MenuModal & MenuItem props

Yserz:


packages/react-ui-kit/demo/Demo.jsx
```diff
@@ -135,12 +135,22 @@ class Demo extends React.PureComponent {
...
         )}
         {this.state.isMenuModalOpen && (
-          <MenuModal onBackgroundClick={() => this.setState({isMenuModalOpen: false})}>
-            <MenuItem onClick={() => this.setState({isMenuModalOpen: false})}>Like</MenuItem>
-            <MenuItem onClick={() => this.setState({isMenuModalOpen: false})}>Edit</MenuItem>
-            <MenuItem onClick={() => this.setState({isMenuModalOpen: false})}>Delete for me...</MenuItem>
-            <MenuItem onClick={() => this.setState({isMenuModalOpen: false})}>Delete for everyone...</MenuItem>
-            <MenuItem onClick={() => this.setState({isMenuModalOpen: false})}>Cancel</MenuItem>
+          <MenuModal data-uie-name="should-be-there" onBackgroundClick={() => this.setState({isMenuModalOpen: false})}>
+            <MenuItem data-uie-name="should-be-there" onClick={() => this.setState({isMenuModalOpen: false})}>
```
atomrc: How about extracting this method into a constant and just give the constant to the click handler.

```
const closeModal = () => this.setState({isMenuModalOpen: false});

//...

<MenuItem onClick={closeModal}>..</MenuItem>
```
This would avoid creating 5 times the same function

packages/react-ui-kit/src/Menu/MenuModal.jsx
```diff
@@ -91,16 +91,20 @@ const MenuModalBackground = styled.div`
...
     <MenuModalBody>
       <MenuModalContent>{children}</MenuModalContent>
     </MenuModalBody>
     <MenuModalBackground onClick={onBackgroundClick} data-uie-name="menu-background" />
   </MenuModalWrapper>
 );
 
-const MenuItem = ({children, onClick}) => <MenuItemContent onClick={onClick}>{children}</MenuItemContent>;
+const MenuItem = ({children, onClick, ...props}) => (
+  <MenuItemContent onClick={onClick} {...props}>
```
atomrc: I believe you don't need to specify the `onClick={onClick}` since it can be handled by the `{...props}` spreading.

PR #829, opened 2018-07-12
feat(core): Export conversation event content types

bennyn:


Yserz:
Are there any imports that need adjustments?

bennyn:
No but I need the exports in another project which is still in beta.

PR #831, opened 2018-07-12
fix(cryptobox): Add missing dependency

ffflorian:
When installing `@wireapp/cryptobox` from npm, `bazinga64` is not installed since none of the sub-dependencies have it as dependency but it is required in [`src/main/Cryptobox.ts`](https://github.com/wireapp/wire-web-packages/blob/8b522bd2e0917438f4aa54d0872654a89700de95/packages/cryptobox/src/main/Cryptobox.ts#L24).


ffflorian:
Same with `logdown`, it is only loaded as dev dependency, but required in the same file.

PR #835, opened 2018-07-13
fix(core): Add conversation ID to payload bundle

bennyn:


bennyn:
As discussed with @Yserz, we will solve the current problem by emitting specified payload bundles (like `TextBundle`, `AssetBundle` & Co.) instead of always emitting a `PayloadBundleIncoming`.

PR #852, opened 2018-07-18
chore(react-ui-kit): Rename Link to ActionLink (Breaking)

lipis:


bennyn:
@lipis & @Yserz @AndyLnd: What's the status with this PR? Also... There is a guideline on how to announce breaking changes: https://github.com/wireapp/wire-web-packages/wiki/Releases

lipis:
This will break a few things, but I'm up for it for a greater good! We already discussed the reasons behind it, while I was in Berlin :)

After updating all the affected projects, we can introduce the `<Link/>` that will be used as the [default hyperlink](https://en.wikipedia.org/wiki/Hyperlink)

PR #864, opened 2018-07-20
chore: no-unused-variable is deprecated in TSLint

ffflorian:
See https://github.com/palantir/tslint/issues/1481.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,6 +8,7 @@
     "noEmitOnError": true,
     "noImplicitReturns": true,
     "noUnusedLocals": true,
+    "noUnusedParameters": true,
     "removeComments": true,
     "sourceMap": true,
     "strict": true,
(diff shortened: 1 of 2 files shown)


bennyn:
We should revert this rule. It forces us to change things like this:

```ts
const binaryString = escapedString.replace(/%([0-9A-F]{2})/g, (match, position) => {
  const code = parseInt(`0x${position}`, 16);
  return String.fromCharCode(code);
});
```

into:

```ts
const binaryString = escapedString.replace(/%([0-9A-F]{2})/g, (_, position) => {
  const code = parseInt(`0x${position}`, 16);
  return String.fromCharCode(code);
});
```

ffflorian:
Oh, right, then we should just set `noUnusedParameters: false`.

ffflorian:
https://github.com/wireapp/wire-web-packages/pull/868

PR #873, opened 2018-07-23
refactor: Export LRUCache and APIClient as objects (breaking)

ffflorian:
```ts
import APIClient = require('@wireapp/api-client');
```
is now
```ts
import {APIClient} from '@wireapp/api-client';
```
and
```js
const LRUCache = require('@wireapp/lru-cache').default;
```
is now
```js
const {LRUCache} = require('@wireapp/lru-cache');
```
---

Importing `package.json` like 

```ts
import {version} from 'package.json';
```
can't be used yet because https://github.com/Microsoft/TypeScript/pull/25457 is not in production yet and it would have several unfavorable implications like duplicating the `package.json` file and removing the `rootDir` entry from the `tsconfig.json` file.


bennyn:
@ffflorian Can we try with "3.1.0-dev.20180721"? :)

ffflorian:
@bennyn It works with TypeScript `3.1.0-dev.20180721` but has several implications which I would consider unfavorable. For now we need to stick with 
```ts
const {version}: {version: string} = require('../package.json');
```

PR #883, opened 2018-07-24
refactor: Use brackets instead of "Array"

ffflorian:
image


packages/core/src/main/client/ClientDatabaseRepository.ts
```diff
@@ -55,7 +55,7 @@ class ClientDatabaseRepository {
   }
 
   public createClientList(userId: string, clientList: RegisteredClient[]): Promise<MetaClient[]> {
-    const createClientTasks: Promise<MetaClient>[] = [];
+    const createClientTasks: Array<Promise<MetaClient>> = [];
```
gregor: Why are you doing the reverse here then?

packages/cryptobox/src/main/store/CryptoboxCRUDStore.ts
```diff
@@ -126,7 +126,7 @@ class CryptoboxCRUDStore implements ProteusSession.PreKeyStore {
    * Saves the serialised formats from a batch of PreKeys.
    */
   public save_prekeys(pre_keys: ProteusKeys.PreKey[]): Promise<ProteusKeys.PreKey[]> {
-    const promises: Promise<ProteusKeys.PreKey>[] = pre_keys.map(pre_key => this.save_prekey(pre_key));
+    const promises: Array<Promise<ProteusKeys.PreKey>> = pre_keys.map(pre_key => this.save_prekey(pre_key));
```
gregor: And here...

packages/core/src/main/client/ClientDatabaseRepository.ts
ffflorian: Because it's not a simple type.

bennyn: Is `Array<Promise<MetaClient>>` different from `Promise<MetaClient>[]`?

ffflorian: No, it's just a a different style of writing.

bennyn: Ok, then let's use `[]` for everything (no matter if complex type or simple type).

ffflorian: Okay!

ffflorian: Will you update the PR then?

bennyn: yes :)

PR #888, opened 2018-07-26
chore(store-engine): Add types for bro-fs

ffflorian:
@bennyn: This breaks at [L136](https://github.com/wireapp/wire-web-packages/compare/chore/store-engine/types?expand=1#diff-2196fc1328142ef35679b5e910e101c2R136) because the `FileReader` will always return a `string` and not `T` when called with `{type: 'Text'}`.

In the `FileEngine` we don't return the data at all if it couldn't be parsed as JSON (see [L141](https://github.com/wireapp/wire-web-packages/pull/888/files#diff-333c4b7db5b6087797f2608a9df38dc5L141)).

We have now three options:
* define `data` as `any` in the `FileSystemEngine`
* don't return the data if it couldn't be parsed as JSON in the `FileSystemEngine` (would be my favorite since it's not a `T` then anyway)
* define the return value of `read()` as `Promise<string>` (need to check the other implementations of this function then, too)


bennyn:
@ffflorian The `FileReader` implements the [read method from CRUDEngine](https://github.com/wireapp/wire-web-packages/blob/%40wireapp/travis-bot%401.0.100/packages/store-engine/src/main/engine/CRUDEngine.ts#L59). Will your favorite option (not returning the data he data if it couldn't be parsed as JSON) break the contract of that function?

ffflorian:
No it would not break the contract. But we should throw an error then.

bennyn:
> No it would not break the contract. But we should throw an error then.

If it does not break the contract, then **go for it**. 👍

packages/store-engine/src/test/shared/append.js
```diff
@@ -6,11 +6,10 @@ module.exports = {
 
     const text = 'Hello';
     const textExtension = '\r\nWorld';
-
     const primaryKey = await engine.create(TABLE_NAME, PRIMARY_KEY, text);
     await engine.append(TABLE_NAME, primaryKey, textExtension);
     const record = await engine.read(TABLE_NAME, primaryKey);
-    expect(record).toBe(`${text}${textExtension}`);
+    expect(record).toBe(text + textExtension);
```
bennyn: Why do you use string concatenation over ES6 string templates here?

PR #889, opened 2018-07-26
fix(react-ui-kit): prevent jumpy window when opening menu modal

AndyLnd:


bennyn:
`fix:` → `fix(react-ui-kit):`

PR #891, opened 2018-07-26
fix: button behaviour, select element style

AndyLnd:


packages/react-ui-kit/src/Form/Select.jsx
```diff
@@ -27,33 +26,38 @@ const StyledContainerSelect = styled.div`
   /* appearance */
   /* positioning */
   display: flex;
-  margin-top: 3px;
   overflow: hidden;
   position: relative;
+ // background: ${COLOR.GRAY_LIGHTEN_92};
```
gregor: Please remove commented code. You set it a few lines below

PR #896, opened 2018-07-30
feat(api-client): Check SSO code endpoint

Yserz:


packages/api-client/src/main/auth/AuthAPI.ts
```diff
@@ -109,6 +111,15 @@ class AuthAPI {
 
     return this.client.sendJSON(config).then((response: AxiosResponse) => response.data);
   }
+
+  public async headInitiateLogin(ssoCode: string): Promise<void> {
+    const config: AxiosRequestConfig = {
+      method: 'head',
+      url: `${AuthAPI.URL.INITIATE_LOGIN}/${AuthAPI.URL.INITIATE_LOGIN}/${ssoCode}`,
```
ffflorian: `initiate-login/initiate-login`?

packages/api-client/src/main/auth/AuthAPI.ts
```diff
@@ -109,6 +111,15 @@ class AuthAPI {
...
+    };
+
+    await this.client.sendJSON(config);
```
ffflorian: The server doesn't return any data?

packages/api-client/src/main/auth/AuthAPI.ts
```diff
@@ -109,6 +111,15 @@ class AuthAPI {
 
     return this.client.sendJSON(config).then((response: AxiosResponse) => response.data);
   }
+
+  public async headInitiateLogin(ssoCode: string): Promise<void> {
+    const config: AxiosRequestConfig = {
+      method: 'head',
```
atomrc: Usually, we use uppercase for HTTP method names

Yserz: Doch :P

PR #897, opened 2018-07-30
fix(core): Receive images

ffflorian:


packages/api-client/src/main/asset/AssetAPI.ts
```diff
@@ -23,20 +23,20 @@ import {base64MD5FromBuffer, concatToBuffer} from '../shims/node/buffer';
...
   private static readonly ASSET_URL = '/assets/v3';
 
   constructor(private readonly client: HttpClient) {}
 
-  getAsset(key: string, token?: string): Promise<ArrayBuffer> {
-    if (!isValidKey(key)) {
-      throw TypeError('Expected key to only contain alphanumeric values and dashes.');
+  getAsset(assetId: string, token?: string): Promise<ArrayBuffer> {
+    if (!isValidAssetId(assetId)) {
+      throw new TypeError('Expected asset ID to only contain alphanumeric values and dashes.');
```
bennyn: Let's log the asset ID for debugging purposes: `Expected asset ID "${assetId}" to only contain alphanumeric values and dashes.`

PR #901, opened 2018-07-30
fix(api-client): Always store the cookie

ffflorian:
Open questions:

* Should we really always save the cookie?
* How can we resolve the `instanceof` problem?


packages/api-client/src/main/shims/node/cookie.ts
```diff
@@ -24,18 +24,16 @@ import {Cookie as ToughCookie} from 'tough-cookie';
...
+}
 
-const loadExistingCookie = async (engine: CRUDEngine): Promise<Cookie> => {
+const loadExistingCookie = (engine: CRUDEngine): Promise<Cookie> => {
   return engine
     .read<PersistedCookie>(AUTH_TABLE_NAME, AUTH_COOKIE_KEY)
-    .catch(error => {
-      if (error instanceof StoreEngineError.RecordNotFoundError) {
+    .catch((error: Error) => {
+      if (error instanceof StoreEngineError.RecordNotFoundError || error.constructor.name === 'RecordNotFoundError') {
```
bennyn: When `error instanceof StoreEngineError.RecordNotFoundError` is not working at all then we should always take `error.constructor.name === 'RecordNotFoundError'`.

ffflorian: Actually we do the same in Account.ts:
https://github.com/wireapp/wire-web-packages/blob/3d2effa274e186d0c54b0c969f20c2b64b1736da/packages/core/src/main/Account.ts#L141-L146

So I vote for checking both with `instanceof` and `error.constructor.name`.

bennyn: But there one check should be sufficient, no?

bennyn:
@ffflorian No downsides that I know of.

PR #902, opened 2018-07-30
fix(react-ui-kit): Fix paddings of header menu

lipis:


packages/react-ui-kit/src/Layout/HeaderMenu.jsx
```diff
@@ -28,19 +28,22 @@ const MenuWrapper = styled.div`
...
     props.open &&
     `
+  left: 0;
+  padding: 0 ${GUTTER};
   position: fixed;
   width: 100%;
-  z-index: 10000;
-  left: 0;
-  padding: 0;`};
+  z-index: 10000;`};
```
lipis: just sorted them all

PR #905, opened 2018-07-30
feat(react-ui-kit): Large text

lipis:
Will be used for showing the price (https://wire.com/pricing/)


bennyn:
If  @AndyLnd  & @Yserz are okay with it, then I am too. 👍

PR #919, opened 2018-08-01
refactor(core): Reduce amount of return types for "handleEvent"

bennyn:
To make our API a bit neater, this PR turns the following signature:

```ts
private async handleEvent(event: IncomingEvent): Promise<PayloadBundleIncoming | IncomingEvent | void>
```

Into:

```ts
private async handleEvent(event: IncomingEvent): Promise<PayloadBundleIncoming | void>
```

### DIFF
--- a/packages/core/src/main/Account.test.node.js
+++ b/packages/core/src/main/Account.test.node.js
@@ -150,7 +150,7 @@ describe('Account', () => {
     });
   });
 
-  fdescribe('"handleEvent"', () => {
+  describe('"handleEvent"', () => {
     it('maps unencrypted events', async done => {
       const account = new Account();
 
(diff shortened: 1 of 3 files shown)


packages/core/src/main/Account.test.node.js
```diff
@@ -124,6 +150,134 @@ describe('Account', () => {
...
+      expect(incomingEvent.state).toBe(PayloadBundleState.INCOMING);
+      expect(incomingEvent.timestamp).toBe(new Date(event.time).getTime());
+      expect(incomingEvent.type).toBe(CONVERSATION_EVENT.MEMBER_JOIN);
+    });
+
+    it('maps "conversation.rename" events', () => {
+      const event = {
+        conversation: 'ed5e4cd5-85ab-4d9e-be59-4e1c0324a9d4',
+        data: {
+          name: 'Tiny Timed Messages',
```
atomrc: What a cute message

PR #929, opened 2018-08-02
refactor(core): Unify mapped message types

bennyn:
## Pull Request Checklist

## Description

Instead of using `Account.INCOMING` we now rely on `PayloadBundleType`.


packages/core/src/main/Account.ts
```diff
@@ -429,31 +437,31 @@ class Account extends EventEmitter {
...
             break;
-          case GenericMessageType.REACTION:
-            this.emit(Account.INCOMING.REACTION, data);
+          case Account.INCOMING.REACTION:
+            this.emit(data.type, event);
             break;
-          case GenericMessageType.TEXT:
-            this.emit(Account.INCOMING.TEXT_MESSAGE, data);
+          case Account.INCOMING.TEXT_MESSAGE:
+            this.emit(data.type, event);
```
ffflorian: Since they all have the same emit data, you can just write them as 
```ts
case Account.INCOMING.ASSET:
case Account.INCOMING.CLIENT_ACTION:
case Account.INCOMING.CONFIRMATION:
...
  this.emit(data.type, event);
  break;
```

packages/core/src/main/conversation/PayloadBundle.ts
```diff
@@ -70,7 +69,7 @@ interface PayloadBundle {
   id: string;
   state: PayloadBundleState;
   timestamp: number;
-  type: GenericMessageType | CONVERSATION_EVENT | USER_EVENT;
+  type: string;
```
ffflorian: Why not `Account.INCOMING`?

bennyn: Because `Account.INCOMING` cannot be used as a type (TypeScript error).

packages/core/src/main/Account.ts
bennyn: Even better, I can use the `default` to do this. :)

packages/core/src/main/conversation/PayloadBundle.ts
ffflorian: Then we will need to refactor `INCOMING` as an enum.

packages/core/src/main/Account.ts
```diff
@@ -434,71 +432,51 @@ class Account extends EventEmitter {
...
-          case GenericMessageType.ASSET: {
+          case PayloadBundleType.CLIENT_ACTION:
+          case PayloadBundleType.CONFIRMATION:
+          case PayloadBundleType.MESSAGE_DELETE:
+          case PayloadBundleType.MESSAGE_HIDE:
+          case PayloadBundleType.ASSET_IMAGE:
+          case PayloadBundleType.PING:
+          case PayloadBundleType.REACTION:
+          case PayloadBundleType.TEXT:
+            this.emit(data.type, data);
```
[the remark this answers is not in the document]
ffflorian: Yeah, @bennyn, take a `break`! :slightly_smiling_face:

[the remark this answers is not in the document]
bennyn: `break;` just got added. Thanks guys! 😎

PR #932, opened 2018-08-02
feat(core): Send and receive files (WEBFOUND-4)

ffflorian:
Breaking changes:
* Incoming images will now be emitted as `Account.INCOMING.IMAGE`.
* Incoming asset metadata will be emitted as `Account.INCOMING.ASSET_META`.
* Incoming asset data will be emitted as `Account.INCOMING.ASSET`.
* `getImage()` in the conversation service was renamed to `getAsset()`.

## Pull Request Checklist


packages/core/src/main/Account.ts
```diff
@@ -322,6 +331,20 @@ class Account extends EventEmitter {
         };
       }
       case GenericMessageType.ASSET: {
+        const {uploaded, original} = genericMessage.asset;
+        const isImage = !!uploaded && !!uploaded.assetId && !!original && !!original.image;
+
+        if (!isImage) {
```
bennyn: Let's start with the positive case first...

packages/core/src/main/Account.ts
```diff
@@ -85,6 +93,7 @@ class Account extends EventEmitter {
     notification: NotificationService;
     self: SelfService;
   };
+  private readonly assetOriginalCache: {[messageId: string]: Original} = {};
```
bennyn: `assetOriginalCache` can be removed. Instead we can emit `Account.INCOMING.IMAGE` & `Account.INCOMING.IMAGE_META`.

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -623,6 +697,10 @@ export default class ConversationService {
     });
   }
 
+  public getImage(data: RemoteData): Promise<Buffer> {
```
bennyn: We can remove this legacy method and run a beta release because with my PR (#929) we will need a major release anyway.

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -641,14 +719,18 @@ export default class ConversationService {
...
             return this.sendImage(conversationId, payloadBundle);
           }
-          throw new Error(`No send method implemented for sending other assets than images.`);
+          if ((payloadBundle.content as FileAssetContent).asset) {
+            return this.sendFile(conversationId, payloadBundle);
+          }
+          throw new Error(`No send method implemented for this kind of asset.`);
         }
-        throw new Error(`No send method implemented for "${payloadBundle.type}" without content".`);
+        throw new Error(`No send method implemented for "${payloadBundle.type}" without content.`);
```
bennyn: Maybe the text `Payload of type "${payloadBundle.type}" does not have a "content" property"` is more specific?

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -641,14 +719,18 @@ export default class ConversationService {
...
+      case GenericMessageType.IMAGE: {
         if (payloadBundle.content) {
           if ((payloadBundle.content as ImageAssetContent).image) {
             return this.sendImage(conversationId, payloadBundle);
           }
-          throw new Error(`No send method implemented for sending other assets than images.`);
+          if ((payloadBundle.content as FileAssetContent).asset) {
+            return this.sendFile(conversationId, payloadBundle);
+          }
+          throw new Error(`No send method implemented for this kind of asset.`);
```
bennyn: Here we should also log a `payloadBundle.type` or something similar to specify "this kind of asset".

ffflorian: And `payloadBundle.type` will be just `asset`.

PR #939, opened 2018-08-03
feat(core): Targeted message encryption (WEBFOUND-17)

ffflorian:
With this PR a message can now be encrypted for certain users only.

## Pull Request Checklist


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -144,6 +146,28 @@ class ConversationService {
...
+  ): Promise<PayloadBundleOutgoing> {
+    const confirmationContent = payloadBundle.content as ConfirmationContent;
+
+    const confirmationObject: PayloadBundleOutgoingUnsent = {
+      ...payloadBundle,
+      type: GenericMessageType.CONFIRMATION,
+    };
+
+    await this.send(conversationId, confirmationObject);
+    await this.deleteMessageEveryone(conversationId, confirmationContent.confirmMessageId);
```
atomrc: Humm there seems to be a flaw in the process here. 
Basically what this code means is that the first that receives the ephemeral message will actually try to delete it for everyone.

In the webapp, what we do is sending this deletion message (in the same conversation) but only targeted (/encrypted) for the person who actually sent the message. 

(see https://github.com/wireapp/wire-webapp/blob/e1357df97249e328cb24e6977444af7417f0513c/app/script/conversation/ConversationRepository.js#L2551)

As you can see, in the webapp, we have a parameter for the `delete_message_everyone` where we can specify a bunch of user ids to send the message to.

Maybe it's the flow to use in that case, maybe we should invent something new, but we should not send the `delete` message to everyone in the convo. 

ping @gregor @bennyn

ffflorian: That's technically correct, but we never implemented encrypting messages for certain people. So for now we have to send the `MessageDelete` to everyone in the group. That's a bit of overhead but every client needs to be able to handle this anyway.

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -159,6 +161,38 @@ class ConversationService {
...
+    payloadBundle: PayloadBundleOutgoingUnsent
+  ): Promise<PayloadBundleOutgoing> {
+    const confirmationContent = payloadBundle.content as ConfirmationContent;
+
+    const confirmationObject: PayloadBundleOutgoingUnsent = {
+      ...payloadBundle,
+      type: PayloadBundleType.CONFIRMATION,
+    };
+
+    await this.send(conversationId, confirmationObject);
```
atomrc: Here the method sends the `confirmation` and the `ephemeralConfirmation` at the same time.
The timeline should be this way:
- send `confirmation` right away ;
- wait until timer is out ;
- send `ephemeralConfirmation`.

Also, if you use `timeout` for that, you are going to run into issues if you have timers > a few days (don't remember exactly the number, something like 20)

packages/core/src/demo/echo.js
```diff
@@ -41,6 +41,18 @@ const messageEchoCache = {};
...
 
+  const sendConfirmation = async (messageId, conversationId, messageTimer, userId) => {
+    const confirmationPayload = account.service.conversation.createConfirmation(messageId);
+    await account.service.conversation.send(conversationId, confirmationPayload);
+
+    if (messageTimer && userId) {
+      setTimeout(
+        () => account.service.conversation.deleteMessageEveryone(conversationId, messageId, [userId]),
+        messageTimer
+      );
```
gregor: Still don't get why we would send this confirmation on messageTimer timeout even in the demo case.

packages/core/src/main/Account.ts
```diff
@@ -251,13 +251,12 @@ class Account extends EventEmitter {
...
-          const expireAfterMillis = genericMessage.ephemeral.expireAfterMillis;
-          unwrappedMessage.messageTimer =
-            typeof expireAfterMillis === 'number' ? expireAfterMillis : (expireAfterMillis as Long).toNumber();
-        }
+        unwrappedMessage.id = genericMessage.messageId;
+        const expireAfterMillis = genericMessage.ephemeral.expireAfterMillis;
+        unwrappedMessage.messageTimer =
+          typeof expireAfterMillis === 'number' ? expireAfterMillis : (expireAfterMillis as Long).toNumber();
         return unwrappedMessage;
       } else {
```
gregor: No need for else after return.

packages/core/src/demo/echo.js
ffflorian: But we send a confirmation in [line 46](https://github.com/wireapp/wire-web-packages/pull/939/files#diff-d497294ce653e28d25a71b3bf9dfb025R46).

packages/core/src/demo/echo.js
```diff
@@ -41,6 +41,15 @@ const messageEchoCache = {};
   const apiClient = new APIClient({store: engine, urls: backend});
   const account = new Account(apiClient);
 
+  const sendConfirmation = async (messageId, conversationId, messageTimer, userId) => {
```
gregor: I still don't think this is the appropriate name. Confirmation is used within the scope of the app to confirm the delivery of the message.

What you are doing here is a message deletion after ephemeral timer run out. Confirmation is not the correct term to be used.

packages/core/src/demo/echo.js
```diff
@@ -41,36 +41,62 @@ const messageEchoCache = {};
...
-      `Message "${messageId}" in "${conversationId}" from "${from}":`,
-      content.text,
+      `Receiving: "${type}" ("${messageId}") in "${conversationId}" from "${from}":`,
+      content,
       messageTimer ? `(ephemeral message, ${messageTimer} ms timeout)` : ''
     );
 
-    const confirmationPayload = account.service.conversation.createConfirmation(messageId);
-    await account.service.conversation.send(conversationId, confirmationPayload);
+    if (messageId) {
```
gregor: In the long run a message ID should not be the trigger of for a confirmation. There are specific content messages that need to be confirmed. Call messages won't be for example.

PR #945, opened 2018-08-06
refactor(core): Get pre keys for the conversation

ffflorian:
This PR maps missing and deleted pre keys of a `payloadBundle` and adds / removes them if needed.
This means we don't have to rely on a first initial failed message sending anymore but actually get the pre keys before sending.

## Pull Request Checklist


packages/core/src/main/conversation/ConversationService.test.node.js
```diff
@@ -33,21 +33,17 @@ const createMessage = (conversationRepository, content) => {
...
-    .forEach(userId => {
-      preKeyBundles[userId] = {};
-      new Array(devicesPerUser)
-        .fill()
-        .map(() => new UUID(4).format())
-        .forEach(deviceId => (preKeyBundles[userId][deviceId] = {}));
-    });
-
-  return preKeyBundles;
+const generatePreKeyBundle = (userCount, devicesPerUser) => {
```
bennyn: Please specify return types for functions.

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -108,16 +107,27 @@ class ConversationService {
...
   }
 
-  // TODO: The correct functionality of this function is heavily based on the case that it always runs into the catch
-  // block
-  private getPreKeyBundles(conversationId: string): Promise<ClientMismatch | UserPreKeyBundleMap> {
-    return this.apiClient.conversation.api.postOTRMessage(this.clientID, conversationId).catch((error: AxiosError) => {
-      if (error.response && error.response.status === 412) {
-        const recipients: UserClients = error.response.data.missing;
-        return this.apiClient.user.api.postMultiPreKeyBundles(recipients);
+  private async getPreKeyBundles(
```
bennyn: Since `generatePreKeyBundles` has been renamend to `generatePreKeyBundle`, the `getPreKeyBundles` function should be renamed too into something like `getPreKeyBundle`?

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -433,16 +437,55 @@ class ConversationService {
...
+    }
+  }
+
+  private async onClientMismatch(
+    error: AxiosError,
+    message: NewOTRMessage,
+    plainTextArray: Uint8Array
+  ): Promise<NewOTRMessage> {
+    if (error.response && error.response.status === StatusCode.PRECONDITION_FAILED) {
+      const {missing: missingDevices, deleted: deletedDevices} = error.response.data;
```
bennyn: - `missingDevices` -> `missingClients`
- `deletedDevices` -> `deletedClients`

packages/core/src/main/conversation/ConversationService.test.node.js
ffflorian: That's just a mock function in JavaScript, I can't specify return types directly here.

packages/core/src/main/conversation/ConversationService.ts
ffflorian: The renaming just happened in a test file. I'm not sure if we should rename this function.

PR #955, opened 2018-08-07
feat(core): Send conversation clearing (WEBFOUND-23)

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/core/src/main/conversation/ConversationService.ts
+++ b/packages/core/src/main/conversation/ConversationService.ts
@@ -45,6 +45,7 @@ import {
 
 import {
   Asset,
+  Cleared,
   ClientAction,
   Confirmation,
   Ephemeral,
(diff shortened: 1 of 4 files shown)


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -574,6 +575,41 @@ class ConversationService {
...
+
+    const {id: selfConversationId} = await this.getSelfConversation();
+
+    await this.sendGenericMessage(this.clientID, selfConversationId, genericMessage);
+
+    return {
+      conversation: conversationId,
+      from: this.apiClient.context!.userId,
+      id: messageId,
+      messageTimer: this.messageTimer.getMessageTimer(conversationId),
```
gregor: Why would this message have a timer?

ffflorian: Good question. It wouldn't.

ffflorian: https://github.com/wireapp/wire-web-packages/pull/956

PR #956, opened 2018-08-07
fix(core): Remove timer from conversation clearing

ffflorian:
## Pull Request Checklist


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -603,7 +603,7 @@ class ConversationService {
       conversation: conversationId,
       from: this.apiClient.context!.userId,
       id: messageId,
-      messageTimer: this.messageTimer.getMessageTimer(conversationId),
+      messageTimer: 0,
```
gregor: Why does the payload need a timer? There is a very specific set of messages that can be ephemeral. Thus this should be optional.

PR #961, opened 2018-08-08
feat(commons): Add runtime detection

lipis:
## Pull Request Checklist

### DIFF
--- a/packages/commons/src/main/util/Runtime.ts
+++ b/packages/commons/src/main/util/Runtime.ts
@@ -18,117 +18,134 @@
  */
 
 import * as platform from 'platform';
-import {BROWSER} from '../config/CommonConfig';
+import {BROWSER, WEBAPP_SUPPORTED_BROWSERS} from '../config/CommonConfig';
 
 class Runtime {
+  public static ELECTRON_APP = {
+    FRANZ: 'franz',
+    WIRE: 'wire',
+  };
+
+  public static OS = {
+    DESKTOP: {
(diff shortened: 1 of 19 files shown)


packages/commons/src/main/util/Runtime.ts
```diff
@@ -0,0 +1,134 @@
...
+    return (Runtime.getPlatform().name || 'unknown').toLowerCase();
+  }
+
+  public static getBrowserVersion(): {major: number; minor: number} {
+    const [majorVersion, minorVersion] = (Runtime.getPlatform().version || 'unknown').split('.');
+    return {major: parseInt(majorVersion, 10), minor: parseInt(minorVersion, 10)};
+  }
+
+  public static getUserAgent(): string {
+    return (Runtime.getPlatform().ua || 'unknown').toLowerCase();
```
gregor: This will crash if platform was not properly loaded.

packages/commons/src/main/util/Runtime.ts
```diff
@@ -0,0 +1,134 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import * as platform from 'platform';
+import {BROWSER, WEBAPP_SUPPORTED_BROWSERS} from '../config/CommonConfig';
+
+class Runtime {
+  private static getPlatform(): Platform {
+    return platform;
```
gregor: Return an empty object if instantiating platform failed otherwise all the later safeguard can be obsolete.

Yserz:
💯

packages/commons/src/main/util/Runtime.ts
Yserz: Let's add a test for that case

PR #967, opened 2018-08-09
fix(api-client): Refresh access token (WEBFOUND-25)

ffflorian:
When the backend sends us an error because the access token is expired, it always has code 403, not 401 (see [Error.hs#L336](https://github.com/wireapp/wire-server/blob/fcf0f3123c07e3459b3ff9b391270f884fa4efb0/services/brig/src/Brig/API/Error.hs#L336)).

## Pull Request Checklist


packages/api-client/src/main/http/HttpClient.ts
```diff
@@ -129,8 +130,8 @@ class HttpClient extends EventEmitter {
           return Promise.reject(networkError);
         }
 
-        if (isUnauthorized) {
-          return this.refreshAccessToken().then(() => this._sendRequest(config, tokenAsParam));
+        if ((isUnauthorized || isForbidden) && isBackendError && error.response.data.message === 'Token expired') {
```
bennyn: I am concerned about the `Token expired` message... What happens if our backend team changes that message?

ffflorian: Understandable. But this is the only difference between an error caused by an expired token and an error with invalid credentials.

bennyn: If you entered invalid credentials, then you don't have a `context` but when entering valid credentials and later on dropping because of an expired access token, then the `context` is set. Can we work with this state?

ffflorian: I am now checking if an access token exists.

bennyn: That's great! 😃

PR #980, opened 2018-08-10
feat(api-client): Provider endpoints

Yserz:
Added a TSLint rule in order to resolve a ordering conflict between two linters.


tslint.json
```diff
@@ -12,9 +12,14 @@
     "no-object-literal-type-assertion": true,
     "no-return-await": true,
     "no-sparse-arrays": true,
+    "ordered-imports": [
+      true,
+      {
+        "named-imports-order": "any"
```
bennyn: Let's use `case-insensitive` here.

PR #992, opened 2018-08-13
fix(react-ui-kit): Adjust modal paddings

Yserz:


packages/react-ui-kit/demo/Demo.jsx
```diff
@@ -244,7 +244,11 @@ class Demo extends React.PureComponent {
 
         {this.state.isModalOpen && (
           <Modal onClose={() => this.setState({isModalOpen: false})}>
-            <Text>Normal Modal</Text>
+            <H2 style={{margin: 0}}>Normal Modal</H2>
```
bennyn: Can we use some kind of CSS reset here instead of defining the defaults ourselves? I am thinking of something like this: https://github.com/zacanger/styled-reset

PR #1007, opened 2018-08-15
fix(api-client): Update Provider type

Yserz:


packages/api-client/src/main/team/service/Provider.ts
```diff
@@ -19,6 +19,10 @@
 
 interface Provider {
   id: string;
+  name: string;
+  email: string;
```
bennyn: Sorting `ʕ·͡ᴥ·ʔ﻿ `

PR #1014, opened 2018-08-16
fix(core): Emit ping

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -331,6 +331,17 @@ class Account extends EventEmitter {
           type: PayloadBundleType.MESSAGE_HIDE,
         };
       }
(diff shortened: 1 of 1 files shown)


packages/core/src/main/Account.ts
```diff
@@ -331,6 +331,17 @@ class Account extends EventEmitter {
           type: PayloadBundleType.MESSAGE_HIDE,
         };
       }
+      case GenericMessageType.KNOCK: {
+        return {
+          conversation: event.conversation,
+          from: event.from,
+          id: genericMessage.messageId,
+          messageTimer: 0,
```
gregor: Let's change this structure and remove make the timer optional at some point.

PR #1017, opened 2018-08-16
fix(core): Emit confirmation

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -320,13 +320,13 @@ class Account extends EventEmitter {
       }
       case GenericMessageType.EDITED: {
         const {
-          text: {content: editedContent},
+          text: {content: editedText},
           replacingMessageId,
         } = genericMessage[GenericMessageType.EDITED];
 
         const content: EditedTextContent = {
           originalMessageId: replacingMessageId,
-          text: editedContent,
+          text: editedText,
         };
 
         return {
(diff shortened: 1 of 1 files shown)


packages/core/src/main/Account.ts
```diff
@@ -270,9 +271,12 @@ class Account extends EventEmitter {
   private mapGenericMessage(genericMessage: any, event: ConversationOtrMessageAddEvent): PayloadBundleIncoming {
     switch (genericMessage.content) {
       case GenericMessageType.TEXT: {
+        const {content: textContent} = genericMessage[GenericMessageType.TEXT];
```
gregor: No need to use destructuring if you are only accessing a single property.

packages/core/src/main/Account.ts
```diff
@@ -284,10 +288,31 @@ class Account extends EventEmitter {
           type: PayloadBundleType.TEXT,
         };
       }
+      case GenericMessageType.CONFIRMATION: {
+        const {firstMessageId} = genericMessage[GenericMessageType.CONFIRMATION];
```
gregor: see above:

````
const confirmMessageId = genericMessage[GenericMessageType.CONFIRMATION].firstMessageId;

const content: ConfirmationContent = {confirmMessageId};
````

File: packages/core/src/main/Account.ts [hunk 3]
```diff
@@ -284,10 +288,31 @@ class Account extends EventEmitter {
...
+          from: event.from,
+          id: genericMessage.messageId,
+          messageTimer: 0,
+          state: PayloadBundleState.INCOMING,
+          timestamp: new Date(event.time).getTime(),
+          type: PayloadBundleType.CONFIRMATION,
+        };
+      }
       case GenericMessageType.DELETED: {
+        const {messageId} = genericMessage[GenericMessageType.DELETED];
```
gregor: see above

PR #1020, opened 2018-08-16
feat(bot-api): Send image

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/bot-api/src/main/MessageHandler.test.node.js
+++ b/packages/bot-api/src/main/MessageHandler.test.node.js
@@ -1,0 +1,49 @@
+/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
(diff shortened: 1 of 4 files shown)


packages/bot-api/src/main/MessageHandler.ts
```diff
@@ -37,6 +38,13 @@ abstract class MessageHandler {
     }
   }
 
+  public async sendImage(conversationId: string, image: ImageContent): Promise<void> {
```
atomrc: How about, writing a test for this one ;) ?

ffflorian: What should I test here? The parameters are ensured by TypeScript. We should rather write tests in the core for `createImage()` and `send()`. :wink:

atomrc: Wow! I see some logic here. If there is logic, then there is room for tests!
You can check that you don't call the `send` when there is no account or no service.

ffflorian: Sure...

packages/bot-api/src/main/MessageHandler.test.node.js
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  describe('"sendImage"', () => {
+    it('just returns without account or service', async () => {
+      bot.account = new Account();
+      await bot.account.init();
+
+      bot.addHandler(mainHandler);
+
+      spyOn(bot.account.service.conversation, 'send').and.returnValue(Promise.resolve());
```
atomrc: You don't need to return anything, because the value is never used

packages/bot-api/src/main/MessageHandler.test.node.js
```diff
@@ -0,0 +1,49 @@
...
+    bot = new Bot({
+      email: 'email@example.com',
+      password: 'my-secret-password',
+    });
+
+    mainHandler = new MessageHandler();
+  });
+
+  describe('"sendImage"', () => {
+    it('just returns without account or service', async () => {
```
atomrc: How about the happy path where you actually call `send`?

ffflorian: Not possible since for that we would need to call `bot.start()` and then the bot tries to log in.

PR #1022, opened 2018-08-17
feat(core,api-client): Send link previews (WEBFOUND-33)

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/test/node/spec/asset/AssetAPI.test.js
+++ b/packages/api-client/src/test/node/spec/asset/AssetAPI.test.js
@@ -1,0 +1,57 @@
+/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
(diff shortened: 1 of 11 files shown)


packages/core/src/demo/echo.js
```diff
@@ -108,6 +150,7 @@ const messageEchoCache = {};
     const cacheOriginal = assetOriginalCache[messageId];
     if (!cacheOriginal) {
       logger.warn(`Uploaded data for message ID "${messageId} was received before the metadata."`);
```
atomrc: I think this is a case we handle in the webapp. If the preview comes before the actual text message, we save it as is. (I believe there was some stories like this in the past with mobile client + bad networking or something).
// ping @gregor

packages/core/src/demo/echo.js
```diff
@@ -91,11 +91,53 @@ const messageEchoCache = {};
...
 
-    await handleIncomingMessage(data);
+      await handleIncomingMessage(data);
+
+      const messageIdOriginal = messageIdCache[messageId];
+
+      if (messageIdOriginal) {
+        textPayload = account.service.conversation.createText(content.text, linkPreviews, messageIdOriginal);
+      } else {
+        logger.warn(`Link preview for message ID "${messageId} was received before the original message."`);
```
atomrc: I think this is a case we handle in the webapp. If the preview comes before the actual text message, we save it as is. (I believe there was some stories like this in the past with mobile client + bad networking or something).
// ping @gregor

atomrc: Ok nevermind, this is just demo code.

packages/api-client/src/test/node/spec/asset/AssetAPI.test.js
```diff
@@ -0,0 +1,57 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+const {APIClient} = require('@wireapp/api-client');
+
+describe('"AssetAPI"', () => {
```
atomrc: :+1:

ffflorian: :smirk:

PR #1026, opened 2018-08-20
test(api-client): Add cases for AuthAPI

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/main/Client.ts
+++ b/packages/api-client/src/main/Client.ts
@@ -63,17 +63,11 @@ class APIClient {
   public notification: {api: NotificationAPI};
   public self: {api: SelfAPI};
   public teams: {
-    invitation: {api?: TeamInvitationAPI};
-    member: {api?: MemberAPI};
-    payment: {api?: PaymentAPI};
-    service: {api?: ServiceAPI};
-    team: {api?: TeamAPI};
-  } = {
-    invitation: {api: undefined},
-    member: {api: undefined},
-    payment: {api: undefined},
-    service: {api: undefined},
-    team: {api: undefined},
+    invitation: {api: TeamInvitationAPI};
+    member: {api: MemberAPI};
+    payment: {api: PaymentAPI};
+    service: {api: ServiceAPI};
+    team: {api: TeamAPI};
   };
   public user: {api: UserAPI};
 
@@ -193,7 +187,6 @@ class APIClient {
     const context = await this.createContext(user.id, clientType);
 
     await this.initEngine(context);
-
     return this.init(clientType);
   }
 
--- a/packages/api-client/src/main/Client.ts
+++ b/packages/api-client/src/main/Client.ts
@@ -18,7 +18,6 @@
  */
 
 import {MemoryEngine} from '@wireapp/store-engine/dist/commonjs/engine';
-import {AxiosResponse} from 'axios';
 import * as logdown from 'logdown';
 
 import {AssetAPI} from './asset/';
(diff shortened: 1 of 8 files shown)


packages/api-client/src/main/auth/AuthAPI.ts
```diff
@@ -74,15 +74,14 @@ class AuthAPI {
   public postLogin(loginData: LoginData): Promise<AxiosResponse<any>> {
     const login = {
       ...loginData,
-      clientType: undefined,
```
ffflorian: This is the actual fix.

packages/api-client/src/main/http/HttpClient.ts
```diff
@@ -129,27 +130,27 @@ class HttpClient extends EventEmitter {
...
 
-        if (isForbidden && this.accessTokenStore && this.accessTokenStore.accessToken && !retry) {
-          return this.refreshAccessToken().then(() => this._sendRequest(config, tokenAsParam, true));
-        }
-
         if (isBackendError) {
           error = BackendErrorMapper.map(error.response.data);
         }
 
+        if ((isForbidden || isUnauthorized) && this.accessTokenStore && this.accessTokenStore.accessToken && !retry) {
```
ffflorian: Let's map the backend error first and then check if it could be a 403 or 401.

packages/api-client/src/main/Client.ts
```diff
@@ -150,70 +142,67 @@ class APIClient {
...
-        } else {
-          throw error;
-        }
-      })
-      .then(() => this.disconnect('Closed by client logout'))
-      .then(() => this.accessTokenStore.delete())
-      .then(() => {
-        delete this.context;
-      });
+  public async logout(options = {ignoreError: false}): Promise<void> {
```
bennyn: As long as it is just a single parameter, we don't need a `options` object.

packages/api-client/src/main/Client.ts
```diff
@@ -150,70 +142,67 @@ class APIClient {
...
-         * It's necessary to initialize the context (Client.createContext()) and the store (Client.initEngine())
-         * for saving the retrieved cookie from POST /access (Client.init()) in a Node environment.
-         */
-        .then((user: User) => this.createContext(user.id, clientType))
-        .then((context: Context) => this.initEngine(context))
-        .then(() => this.init(clientType))
-    );
+  public async register(userAccount: RegisterData, clientType: ClientType = ClientType.PERMANENT): Promise<Context> {
+    if (this.context) {
+      await this.logout({ignoreError: true});
```
bennyn: In the webapp we keep track of a [logout reason](https://github.com/wireapp/wire-webapp/blob/bd8d7d07bd91f80391d33da207071ed9013d627b/app/script/auth/route.js#L30). Maybe it's better to pass a logout reason (`register`, `login`, ...) instead of a boolean value.

packages/api-client/src/test/node/spec/auth/AuthAPI.test.js
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+const {APIClient} = require('@wireapp/api-client');
+const {ClientType} = require('@wireapp/api-client/dist/commonjs/client/');
+
+describe('"AuthAPI"', () => {
```
bennyn: We don't do `"` on top-level block. Only on function level describes.

PR #1044, opened 2018-08-24
fix(api-client): Throw on backend logout, add test case for cookie refresh

ffflorian:
## Pull Request Checklist


packages/api-client/src/test/node/spec/Client.test.js
```diff
@@ -128,43 +128,50 @@ describe('Client', () => {
...
 
-    it('can login after a logout', done => {
+    it('can login after a logout', () => {
       const client = new APIClient();
-      client
-        .login(loginData)
-        .then(() => client.logout())
-        .then(done)
-        .catch(done.fail);
+      return client.login(loginData).then(() => client.logout());
```
atomrc: Isn't there an assertion we can make here?

atomrc: Also just reading the test, it doesn't seem we try to login after a logout but rather the other way around (trying to logout after a login). Am I wrong?

atomrc: Since it's not new code, not a blocker, but having a look at it could be cool :)

PR #1055, opened 2018-08-28
feat(api-client): GET supported currencies

Yserz:


packages/api-client/src/main/team/payment/PaymentAPI.ts
```diff
@@ -129,6 +130,15 @@ class PaymentAPI {
...
   }
+
+  public getSupportedCurrencies(teamId: string): Promise<string[]> {
+    const config: AxiosRequestConfig = {
+      method: 'get',
+      url: `${PaymentAPI.URL.TEAMS}/${teamId}/${PaymentAPI.URL.BILLING}/${PaymentAPI.URL.CURRENCY}`,
+    };
+
+    return this.client.sendJSON(config).then((response: AxiosResponse) => response.data);
+  }
```
ffflorian: Can you add links to Swagger?

Yserz: There is no swagger :P

PR #1067, opened 2018-08-31
 refactor(react-ui-kit): New MatchMedia components

AndyLnd:


packages/react-ui-kit/src/Layout/matchMedia.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+MatchMedia.propTypes = {
+  children: PropTypes.node.isRequired,
+  not: PropTypes.bool,
+  query: PropTypes.string.isRequired,
+};
+
+MatchMedia.defaultProps = {
+  not: false,
+};
```
ffflorian: What about `children` and `query` propTypes? They are marked as required..?

PR #1088, opened 2018-09-07
fix(api-client): Paginated invoices endpoint

Yserz:


packages/api-client/src/main/team/payment/PaymentStripeInvoices.ts
```diff
@@ -0,0 +1,28 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {PaymentStripeInvoice} from './PaymentStripeInvoice';
+
+// https://stripe.com/docs/api#invoice_object
```
bennyn: Instead of defining our own, could we use: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/stripe/index.d.ts#L1888 ?

Yserz: Unfortunately these types are just a subset of stripes data types.

PR #1099, opened 2018-09-12
refactor(core): Add TextContentBuilder [BREAKING]

ffflorian:
Old:
```ts
const textMessage = conversation.createText(text, linkPreviews, messageId);
```

New:
```ts
const textMessage = conversation.createText(text, messageId).withLinkPreviews(linkPreviews);
```

## Pull Request Checklist

### DIFF
--- a/packages/core/src/main/conversation/TextContentBuilder.ts
+++ b/packages/core/src/main/conversation/TextContentBuilder.ts
@@ -49,7 +49,14 @@ class TextContentBuilder {
     return this.payloadBundle.type;
   }
 
-  public withLinkPreviews(linkPreviews: LinkPreviewUploadedContent[]): TextContentBuilder {
+  public withLinkPreviews(linkPreview: LinkPreviewUploadedContent): TextContentBuilder {
+    const content = this.payloadBundle.content as TextContent;
+    content.linkPreviews = [linkPreview];
+
+    return this;
+  }
+
+  public withLinkPreview(linkPreviews: LinkPreviewUploadedContent[]): TextContentBuilder {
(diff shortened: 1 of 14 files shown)


packages/core/src/main/conversation/TextContentBuilder.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  public get timestamp(): number {
+    return this.payloadBundle.timestamp;
+  }
+
+  public get type(): PayloadBundleType {
+    return this.payloadBundle.type;
+  }
+
+  public withLinkPreviews(linkPreview: LinkPreviewUploadedContent): TextContentBuilder {
```
bennyn: `withLinkPreviews(linkPreview: LinkPreviewUploadedContent)` -> `withLinkPreviews(linkPreview: LinkPreviewUploadedContent[])`

packages/core/src/main/conversation/TextContentBuilder.ts
```diff
@@ -0,0 +1,67 @@
...
+    const content = this.payloadBundle.content as TextContent;
+    content.linkPreviews = [linkPreview];
+
+    return this;
+  }
+
+  public withLinkPreview(linkPreviews: LinkPreviewUploadedContent[]): TextContentBuilder {
```
bennyn: Let's drop the single version for now. Less convenience but also less maintenance.

packages/core/src/main/conversation/TextContentBuilder.ts
ffflorian: Yes 😆

PR #1103, opened 2018-09-13
chore(core): Add mentions in sender demo

bennyn:


packages/core/src/demo/sender.js
```diff
@@ -125,7 +125,28 @@ const {FileEngine} = require('@wireapp/store-engine');
     await account.service.conversation.clearConversation(CONVERSATION_ID);
   }
 
-  const methods = [sendAndDeleteMessage, sendAndEdit, sendEphemeralText, sendFile, sendImage, sendPing, sendText];
+  async function sendMention() {
+    const mentionEveryone = {
+      end: 14,
```
ffflorian: This is now `length`: https://github.com/wireapp/generic-message-proto/blob/master/proto/messages.proto#L93

PR #1125, opened 2018-09-18
feat(core): Emit "cleared" messages

ffflorian:
## Pull Request Checklist


packages/core/src/demo/echo.js
```diff
@@ -254,6 +254,10 @@ const messageIdCache = {};
     await sendMessageResponse(data, imagePayload);
   });
 
+  account.on(PayloadBundleType.CLEARED, async data => {
```
atomrc: You can simplify like this

```
account.on((PayloadBundleType.CLEARED, handleIncomingMessage)
```

PR #1126, opened 2018-09-18
refactor(lru-cache): Implement iterator and getAll returns object (BREAKING)

ffflorian:
Given we have a message cache like:
```ts
const messageCache = new LRUCache();
messageCache.add(...);
```

Instead of using:
```ts
const messages = messageCache.getAll();
for (const messageObject of messages) {
  const objectId = Object.keys(message)[0];
  const message = messageObject[objectId];
  // do something with the message
}
```

we can now iterate over the messages:
```ts
for (const message of messageCache) {
  // do something with the message
}
```

## Pull Request Checklist

### DIFF
--- a/packages/lru-cache/test/specs/LRUCacheSpec.js
+++ b/packages/lru-cache/test/specs/LRUCacheSpec.js
@@ -102,6 +102,19 @@ describe('LRUCache', () => {
     });
   });
 
+  describe('"iterator"', () => {
+    it('iterates over the nodes', () => {
+      const cache = new LRUCache(4);
+      cache.set('1', 'Apple');
+      cache.set('2', 'Orange');
+      cache.set('3', 'Tomato');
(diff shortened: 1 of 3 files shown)


packages/lru-cache/test/specs/LRUCacheSpec.js
```diff
@@ -102,6 +102,19 @@ describe('LRUCache', () => {
...
   });
 
+  describe('"iterator"', () => {
+    it('iterates over the nodes', () => {
+      const cache = new LRUCache(4);
+      cache.set('1', 'Apple');
+      cache.set('2', 'Orange');
+      cache.set('3', 'Tomato');
+      cache.set('4', 'Plum');
+      for (const value of cache) {
```
bennyn: Can you also do `let [key, value] of` here?

PR #1142, opened 2018-09-20
fix: Make notification services public

bennyn:
## Pull Request Checklist


bennyn:
Looks like the test "updates last event date if lesser database entry exists" is flaky.

FYI: I changed the access modifiers from `private` to `public` because services are stateless and are the public entry point for people to use our code. We also have been using the code already but from JS which disrespects access modifiers so we were not aware that we should modify the `private` state.

PR #1186, opened 2018-10-01
feat(api-client, core): Implement OTR Muted Status

ffflorian:
## Pull Request Checklist


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -1151,20 +1152,36 @@ class ConversationService {
...
   }
 
+  public async toggleMuteConversation(
+    conversationId: string,
+    muted: MutedStatus,
+    muteTimestamp?: number | Date
+  ): Promise<void>;
   public async toggleMuteConversation(
     conversationId: string,
     muted: boolean,
```
bennyn: If `muted: boolean` is here for legacy reasons, then we should document that.

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -1151,7 +1152,24 @@ class ConversationService {
     this.clientID = clientID;
   }
 
-  public async toggleArchiveConversation(
+  public setConversationMutedStatus(
+    conversationId: string,
+    status: MutedStatus,
+    muteTimestamp: number | Date = new Date()
```
gregor: This should never default to the current date. It has to be explicitly set to the latest known server timestamp in the conversation.

bennyn: Let's remove the default value then and continue.

PR #1212, opened 2018-10-09
chore: Replace explicit type castings if possible

bennyn:
Over time the TypeScript definitions have evolved so we don't need some of our legacy type castings anymore. There is a TSLint rule called "no-unnecessary-type-assertion" which helps us keeping track of that.


bennyn:
Also testing it here: https://github.com/wireapp/wire-desktop-packages/pull/7/commits/e9e2e63bb7ac0c486b270c6d14f9e7c21dba5b9e

ffflorian:
@bennyn Maybe https://github.com/palantir/tslint/issues/3540 is related?

PR #1223, opened 2018-10-11
feat(api-client,core): Set availability status (WEBFOUND-50)

ffflorian:
## Pull Request Checklist


packages/core/src/main/broadcast/BroadcastService.ts
```diff
@@ -0,0 +1,85 @@
...
+    const recipients = await this.cryptographyService.encrypt(plainTextArray, preKeyBundle);
+
+    return this.sendOTRBroadcastMessage(clientId, recipients, plainTextArray);
+  }
+
+  private async sendOTRBroadcastMessage(
+    sendingClientId: string,
+    recipients: OTRRecipients,
+    plainTextArray: Uint8Array,
+    data?: any
```
bennyn: Can this be specified?

ffflorian: Unfortunately not, we also have this in [ConversationService.ts#L479](https://github.com/wireapp/wire-web-packages/blob/master/packages/core/src/main/conversation/ConversationService.ts#L479).

packages/core/src/main/user/UserService.test.node.js
```diff
@@ -1,29 +1,120 @@
...
       })
       .persist();
+
+    nock(Backend.PRODUCTION.rest)
+      .post(BroadcastAPI.URL.BROADCAST, body => {
+        hasState[body.sender] = true;
+        return true;
+      })
+      .query(() => true)
+      .reply(StatusCode.OK, '{}')
```
bennyn: Let's return proper data here.

packages/core/src/main/user/UserService.test.node.js
```diff
@@ -24,6 +51,71 @@ describe('UserService', () => {
...
+          ],
+          user: requestedUserId,
+        };
+
+        return [StatusCode.OK, JSON.stringify(data)];
+      })
+      .persist();
+  });
+
+  beforeEach(() => {
```
bennyn: one-liner

packages/core/src/main/user/UserService.test.node.js
```diff
@@ -1,18 +1,45 @@
...
 
 const PayloadHelper = require('../test/PayloadHelper');
 
-const HTTPS_URL = Backend.PRODUCTION.rest;
-
 describe('UserService', () => {
+  let hasState;
+
+  afterAll(() => {
+    nock.cleanAll();
```
bennyn: one-liner

PR #1234, opened 2018-10-17
feat(commons): Add AccentColor

Yserz:


packages/commons/src/main/util/AccentColor.ts
```diff
@@ -0,0 +1,75 @@
...
+  STRONG_LIME_GREEN,
+  VIVID_RED,
+  BRIGHT_ORANGE,
+  SOFT_PINK,
+  VIOLET,
+];
+
+export const accentColorById = (id: number): AccentColor | undefined => ACCENT_COLORS.find(color => color.id === id);
+
+export const random = (): AccentColor => RandomUtil.randomArrayElement(ACCENT_COLORS);
```
AndyLnd: Can we have a more descriptive name here?
```suggestion
export const randomColor = (): AccentColor => RandomUtil.randomArrayElement(ACCENT_COLORS);
```

gregor:
We should still support the removed yellow accent color but flag it as no longer selectable

PR #1241, opened 2018-10-22
feat(core): Add quote to text messages (WEBFOUND-53)

ffflorian:
## Pull Request Checklist


packages/core/src/main/Account.ts
```diff
@@ -380,6 +384,10 @@ class Account extends EventEmitter {
           content.mentions = editedMentions;
         }
 
+        if (editedQuote) {
+          content.quote = editedQuote;
```
gregor: Add comment that while this is possible on protobuf level it should be ignored on receiver side according to spec

ffflorian: I added it in `EditedTextContent.ts`.

PR #1297, opened 2018-11-12
fix: Token refresh

Yserz:


packages/api-client/src/test/node/spec/http/HttpClient.test.js
```diff
@@ -0,0 +1,85 @@
...
+            'iJCRCjc8oROO-dkrkqCXOade997oa8Jhbz6awMUQPBQo80VenWqp_oNvfY6AnU5BxEsdDPOBfBP-uz_b0gAKBQ==.v=1.k=1.d=1498600993.t=a.l=.u=aaf9a833-ef30-4c22-86a0-9adc8a15b3b4.c=15037015562284012115',
+          expires_in: 900,
+          token_type: 'Bearer',
+          user: 'aaf9a833-ef30-4c22-86a0-9adc8a15b3b4',
+        },
+      };
+      const errorMessage = 'cookie invalid';
+
+      const client = new HttpClient('https://test.zinfra.io', mockedAccessTokenStore, undefined);
+      axios.request = () =>
```
ffflorian: Let's use a `spy` here.

PR #1391, opened 2018-12-17
fix: Improved isUUIDv4 check (WEBAPP-5835)

pwnsdx:


packages/commons/src/main/util/ValidationUtil.ts
```diff
@@ -18,6 +18,6 @@
  */
 
 export function isUUIDv4(candidate: string): boolean {
-  const uuidv4Regex = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
-  return uuidv4Regex.test(candidate.toLowerCase());
+  const uuidv4Regex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
```
ffflorian: Since we are here, let's refine the regex:
```suggestion
  const uuidv4Regex = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
```

See https://www.cryptosys.net/pki/uuid-rfc4122.html.

PR #1470, opened 2019-01-23
chore(api-client): Deprecate signaling keys

bennyn:
Signaling keys got deprecated with:
- https://github.com/wireapp/wire-server/releases/tag/v2019-01-10


ffflorian:
Maybe rename the PR to "remove signaling keys"?

PR #1496, opened 2019-02-04
fix(copy-config): Support paths from Windows systems

bennyn:
## Pull Request Checklist

### DIFF
--- a/packages/copy-config/src/main/CopyConfig.test.js
+++ b/packages/copy-config/src/main/CopyConfig.test.js
@@ -111,4 +111,16 @@ describe('CopyConfig', () => {
     delete process.env.WIRE_CONFIGURATION_EXTERNAL_DIR;
     delete process.env.WIRE_CONFIGURATION_FILES;
   });
+
+  it('is compatible with Window paths', () => {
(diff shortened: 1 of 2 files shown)


packages/copy-config/src/main/CopyConfig.test.js
```diff
@@ -95,7 +94,7 @@ describe('CopyConfig', () => {
     expect(copiedResult.length).toBe(1);
   });
 
-  it('reads environment variables', async () => {
+  it('can be configures using environment variables', async () => {
```
AndyLnd:
```suggestion
  it('can be configured using environment variables', async () => {
```

packages/copy-config/src/main/CopyConfig.test.js
```diff
@@ -112,4 +111,16 @@ describe('CopyConfig', () => {
     delete process.env.WIRE_CONFIGURATION_EXTERNAL_DIR;
     delete process.env.WIRE_CONFIGURATION_FILES;
   });
+
+  it('is compatible with Window paths', () => {
```
Yserz:
```suggestion
  it('is compatible with Windows paths', () => {
```

PR #1555, opened 2019-02-25
chore: Allow default imports from modules with no default export

bennyn:
@Yserz made me aware that `esModuleInterop` is a new flag (introduced with [TypeScript 2.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html)) to support importing CommonJS modules with ES6 syntax. Using this flag brings the output generated by TypeScript closer to that generated by Babel. It will also enable us to [update the reconnecting-websocket](https://github.com/pladaria/reconnecting-websocket/issues/95) library.

You can read here why `esModuleInterop` should be used:
- https://itnext.io/great-import-schism-typescript-confusion-around-imports-explained-d512fc6769c2

The `esModuleInterop` flag is also enabled when creating new TypeScript projects with `tsc --init`.


packages/travis-bot/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "declaration": true,
```
ffflorian: Why did you remove the declaration files?

PR #1605, opened 2019-03-06
refactor(bazinga64): Organize classes in separate files

bennyn:
## Pull Request Checklist

### DIFF
--- a/packages/bazinga64/tsconfig.json
+++ b/packages/bazinga64/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "outDir": "dist",
-    "rootDir": "src/main"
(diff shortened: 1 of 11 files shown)


packages/bazinga64/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "outDir": "dist",
-    "rootDir": "src/main"
+    "rootDir": "src"
```
ffflorian: :+1:

bennyn: Were `src/main` came from: https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html 😃

PR #1621, opened 2019-03-11
fix(api-client): Return type of PUT billing

Yserz:


packages/api-client/src/main/team/payment/PaymentDataUpdate.ts
```diff
@@ -22,6 +22,7 @@ import {PaymentInterval} from '../payment/';
 interface PaymentDataUpdate {
   interval?: PaymentInterval;
   source?: string;
+  coupon?: string;
```
ffflorian: Sorting...

PR #1641, opened 2019-03-14
feat(commons): Add Permission & Role

Yserz:


Yserz:
Might be better to move that to the api-client...

packages/commons/src/main/permission/Permission.test.ts
```diff
@@ -0,0 +1,113 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {Permissions, hasPermissions} from './Permission';
+
+describe('Permission', () => {
+  describe('hasPermission true for', () => {
+    it('CREATE_CONVERSATION and 1', () => expect(hasPermissions(1, Permissions.CREATE_CONVERSATION)).toBe(true));
```
bennyn: Can the spec descriptions be made more descriptive?

PR #1647, opened 2019-03-15
feat(copy-config): Add basedir configuration and cleanup basedir after copying files

ffflorian:
## Pull Request Checklist


ffflorian:
@Yserz what do you think? Do you need the copied configuration after all files were copied or can we delete it?

packages/copy-config/src/main/CopyConfig.ts
```diff
@@ -202,6 +201,15 @@ export class CopyConfig {
...
     }
 
+    if (!this.noCleanup) {
+      await this.removeBasedir();
+    }
+
     return copiedFiles.sort();
   }
+
+  public async removeBasedir(): Promise<void> {
```
ffflorian: Should be private.

PR #1652, opened 2019-03-18
fix(api-client): Add role property to NewTeamInvitation

Yserz:


packages/api-client/src/main/team/invitation/NewTeamInvitation.ts
```diff
@@ -17,8 +17,11 @@
  *
  */
 
+import {MemberRole} from '../MemberRole';
+
 interface NewTeamInvitation {
   email: string;
+  role: MemberRole;
```
ffflorian: Sorting

PR #1667, opened 2019-03-21
chore: Enforce modern versions of node and yarn

bennyn:
Currently when you try to run `yarn && yarn boot` with yarn v0.15 it will fail because we need a very recent version of yarn. 

@ffflorian Do you know which version of yarn enforced https links in `yarn.lock` file? Then we could even make this our lower end requirement.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "engines": {
     "node": ">= 10",
-    "yarn": ">= 1.10.0"
+    "yarn": ">= 1"
   },
   "lint-staged": {
     "*.{js,jsx}": [
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "tslint-react-hooks": "2.0.0"
   },
   "engines": {
-    "node": "10.x.x",
+    "node": ">= 10",
     "yarn": "1.x.x"
   },
   "lint-staged": {
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "engines": {
     "node": ">= 10",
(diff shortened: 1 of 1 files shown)


ffflorian:
@bennyn

> Do you know which version of yarn enforced https links in `yarn.lock` file?

Yes, it was fixed in https://github.com/yarnpkg/yarn/pull/6353 and released with [v1.10.0](https://github.com/yarnpkg/yarn/releases/tag/v1.10.0).

package.json
```diff
@@ -17,6 +17,10 @@
     "tslint-react": "3.6.0",
     "tslint-react-hooks": "2.0.0"
   },
+  "engines": {
+    "node": "10.x.x",
```
ffflorian:
```suggestion
    "node": ">= 10",
```

package.json
```diff
@@ -17,6 +17,10 @@
...
+    "yarn": "1.x.x"
```
ffflorian:
```suggestion
    "yarn": ">= 1.10.0"
```

ffflorian:
If we are setting the minimum version of Node to 10, we could also let TypeScript compile all code to ES6.

bennyn:
Thanks for checking the yarn version. We can think of ES6 compilation since the webapp uses Babel now. So far we also have been pre-compiling code for the webapp to ES5 but with the upgraded build system on webapp that might not be needed anymore.

ffflorian:
Netlify doesn't work with this anymore since it's using yarn 1.3.2.

PR #1670, opened 2019-03-21
feat(bot-api): Expose send text method

bennyn:
## Pull Request Checklist


ffflorian:
Btw, why don't you use `MessageHandler.sendText()`?

bennyn:
Because then I need to implement a handler first and provide a method to reply to message which I don't need for a fire-and-forget bot (like a status bot).

PR #1673, opened 2019-03-22
refactor(commons): Migrate browser tests to TypeScript

Yserz:


packages/commons/src/main/util/Runtime.test.browser.ts
```diff
@@ -17,13 +17,15 @@
  *
  */
 
+/* tslint:disable:no-object-literal-type-assertion */
+
 import * as platform from 'platform';
-import {Runtime} from '@wireapp/commons';
+import {OS, Runtime} from './Runtime';
 
 describe('Runtime', () => {
```
bennyn: Let's kill this root describe block

packages/commons/src/main/util/Runtime.test.browser.ts
```diff
@@ -17,13 +17,15 @@
...
   describe('"isAndroid"', () => {
```
bennyn: Time to remove `'"`. Can you please do that with search & replace?

PR #1694, opened 2019-03-27
chore: Run selective tests for private packages

bennyn:
We had an issue where tests from packages (which have `"private": true` defined) were not executed during the `testUpdated.js` run. This is now fixed. :)

## Pull Request Checklist


bin/testUpdated.js
```diff
@@ -20,23 +20,22 @@
 const {execSync} = require('child_process');
 
 let output;
-console.info('Checking for updated packages');
+console.info('Checking for changed packages...');
+
 try {
-  output = execSync(`npx lerna updated`);
+  output = execSync(`npx lerna changed --all --json`);
```
ffflorian: :+1:

bennyn: Good call! I opened a ticket to get information about their API: https://github.com/lerna/lerna/issues/2013

PR #1702, opened 2019-03-29
fix(commons): Don't add a default namespace

bennyn:
At the beginning I thought it is a good idea to use `process.env.npm_package_name` as the default namespace for loggers in the Node.js world but then I encountered the following:

- There is package "A" with a default logger namespace of `process.env.npm_package_name`
- There is package "B" with a default logger namespace of `process.env.npm_package_name`
- Your application uses "A" and "B" and gets started; now the namespace of "A" and "B" is actually the package name of your application 🙈 

So it is better not have such a default namespace behaviour. The user is now in charge to set a namespace.

I also fixed a small bug where it was necessary to set a namespace in order to get timestamps next to log statements.

## Pull Request Checklist


atomrc:
Ahahah, good to know :+1:

PR #1717, opened 2019-04-02
fix(api-client): Replace unsupported console.info statements to support MS Edge 15

Yserz:


bennyn:
Problem with MS Edge 15 seem to get created by `core-js`. 

I just saw in MS Edge 15 that the following file is throwing an error: [host-report-errors.js](https://github.com/zloirock/core-js/blob/v3.0.0/packages/core-js/internals/host-report-errors.js)

PR #1738, opened 2019-04-08
feat(commons): Pattern calculation when setting new passwords

Yserz:


ffflorian:
What about adding some test? :wink:

Yserz:
It's likely that this will change soon since not everything is spec'ed out. For now it's more important to actually use it everywhere IMHO.

PR #1776, opened 2019-04-23
feat: Remove default exports

ffflorian:


packages/tslint-config/tslint.json
```diff
@@ -15,6 +15,7 @@
     "jsx-boolean-value": false,
     "jsx-no-lambda": false,
     "jsx-no-multiline-js": false,
+    "no-default-export": true,
```
lipis: 👍

lipis: This will give bunch of errors in other projects.. but that's a good thing!

ffflorian: Yes, I will mark it as breaking

bennyn:
**Note:** I restored the `feat/no-default` branch because tests were running without errors on Chrome 73 and I wanted to check how it looks like with Chrome 74. Unfortunately our Store Engine tests for the IndexedDBEngine fail with Chrome 74. 😢

PR #1786, opened 2019-04-26
feat(tslint-config): add typedef

ffflorian:


packages/tslint-config/tslint.json
```diff
@@ -26,7 +26,6 @@
     "no-return-await": true,
     "no-sparse-arrays": true,
     "no-this-assignment": true,
-    "no-unused-variable": true,
```
lipis: :(

ffflorian: It's deprecated

lipis: but there is no alternative at the moment.. and it's still under discussion: https://github.com/palantir/tslint/issues/4100 

My vote is to keep it for now..

PR #1811, opened 2019-05-08
test(store-engine): Fix flaky LocalStorageEngine tests

bennyn:
I adjusted the test cases because they expected a fixed order of items from `readAllPrimaryKeys` but that order is not guaranteed (and never was).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ node_js:
   - '10'
 
 addons:
-  chrome: beta
+  chrome: stable
 
 # http://docs.travis-ci.com/user/build-lifecycle/
 before_install:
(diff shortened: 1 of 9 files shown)


bennyn:
Problem sits in `createSpec`.

bennyn:
Problematic test case: `throws an error when attempting to overwrite a record`

bennyn:
Chrome 74 seems to be the issue: https://github.com/wireapp/wire-web-packages/pull/1776#event-2327792722

bennyn:
Failures in IndexedDBEngine seem to be just noise. The real trouble maker seems to be the `LocalStorageEngine`.

bennyn:
The order of `Object.keys(localStorage)` in `readAllPrimaryKeys` seem to not be preserved anymore with Chrome v74.

PR #1866, opened 2019-05-27
feat(store-engine-fs): Distribute FileEngine in separate package

bennyn:
Our `@wireapp/store-engine` will be based on an adapter technique. This means we will move all implementations of `CRUDEngine` (except the `MemoryEngine`) into single packages. 

We will start with the `FileEngine`. Once it is released, I will remove the `FileEngine` from `@wireapp/store-engine`, so it must be consumed from `@wireapp/store-engine-fs`.

## Pull Request Checklist

### DIFF
--- a/packages/store-engine-fs/src/index.ts
+++ b/packages/store-engine-fs/src/index.ts
@@ -25,8 +25,8 @@ import {
   RecordTypeError,
   UnsupportedError,
 } from '@wireapp/store-engine/dist/commonjs/engine/error/';
-import * as fs from 'fs-extra';
-import * as path from 'path';
+import fs from 'fs-extra';
+import path from 'path';
 
 export class FileEngine implements CRUDEngine {
   [index: string]: any;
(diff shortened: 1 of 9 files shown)


packages/store-engine-fs/src/index.ts
```diff
@@ -0,0 +1,252 @@
...
+
+export class FileEngine implements CRUDEngine {
+  [index: string]: any;
+
+  public storeName = '';
+  public options: {fileExtension: string} = {
+    fileExtension: '.dat',
+  };
+  // Using a reference to Node.js' "path" module to influence the platform-specific behaviour in our tests
+  public static path: any = path;
```
ffflorian: Is this still needed?

bennyn: Unfortunately yes. During the tests we need to influence if `path.posix` should be used or not. Since the import is at the beginning of the file we cannot dynamically exchange it so instead of changing the import we change the `path: any` reference.

PR #1873, opened 2019-05-29
feat(core): Add Legal Hold Status (BREAKING)

ffflorian:
## Pull Request Checklist

## Breaking Change
The following functions have a new parameter order:
* `createConfirmation()`
* `createFileMetadata()`
* `createImage()`

All of them now accept the optional `messageId` as **third parameter instead of fourth**. This is due to the growing parameter list. We need an order here and as decided with @Yserz, the `messageId` is more important than other optional parameters like `expectsReadConfirmation` or `legalHoldStatus`.
This change also aligns the parameter order of these functions with all other functions which accept an optional `messageId`.


packages/core/src/main/conversation/ConversationService.test.node.js
```diff
@@ -379,5 +380,28 @@ describe('ConversationService', () => {
       expect(replyMessage.content.text).toEqual(text);
       expect(replyMessage.content.expectsReadConfirmation).toEqual(true);
     });
+
+    it('adds a legal hold status correctly', () => {
```
bennyn:
```suggestion
    it('adds a legal hold status', () => {
```

PR #1912, opened 2019-06-07
fix(store-engine-sqleet): Export SQLeetEngine

bennyn:


packages/store-engine-sqleet/webpack.config.ts
```diff
@@ -42,5 +42,27 @@ const config: webpack.Configuration = {
...
+    output: {
+      filename: 'SQLeetEngine.web.js',
+      library: 'StoreEngine',
+      path: `${__dirname}/dist`,
+    },
+    target: 'web',
+  },
+};
+
+module.exports = [nodeConfig, webConfig];
```
ffflorian: use `export =` or `export default`

packages/store-engine-sqleet/src/index.ts
```diff
@@ -82,7 +84,7 @@ export class SQLeetEngine implements CRUDEngine {
     }
 
     const SQL = await initSqlJs(this.dbConfig);
-    this.db = new SQL.Database(existingDatabase);
+    this.db = new SQL.Database(existingDatabase as Uint8Array);
```
ffflorian: Can be `undefined`, too

PR #1937, opened 2019-06-12
fix(store-engine,store-engine-sqleet): readAll

pwnsdx:


packages/store-engine-sqleet/src/index.ts
```diff
@@ -239,9 +240,18 @@ export class SQLeetEngine implements CRUDEngine {
...
+    let records = this.db.exec(selectRecordStatement);
+
+    // Ensure the record is not empty
+    if (records && records[0]) {
+      records = records[0];
+    } else {
+      return [];
+    }
+
+    return records.values.map((record: any[]) => zipObject(records.columns, record));
```
bennyn: Let's replace `zipObject` with standard ES6: https://lowrey.me/lodash-zipobject-in-es6-javascript/

PR #1950, opened 2019-06-17
feat(api-client,core,commons): Chunked /users requests & types ArrayUtils

Yserz:


packages/api-client/src/main/user/UserAPI.ts
```diff
@@ -38,6 +39,7 @@ import {
 } from '../user/';
 
 export class UserAPI {
+  static DEFAULT_USERS_CHUNK_SIZE = 50;
```
ffflorian:
```suggestion
  static readonly DEFAULT_USERS_CHUNK_SIZE = 50;
```

Or will we change this value?

packages/api-client/src/main/user/UserAPI.ts
```diff
@@ -254,7 +256,28 @@ export class UserAPI {
    * @param parameters Multiple user's handles or IDs
    * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/users/users
    */
-  public getUsers(parameters: {handles?: string[]; ids?: string[]}): Promise<User[]> {
+  public async getUsers(
+    parameters: {handles?: string[]; ids?: string[]},
```
ffflorian:
```suggestion
    parameters: {handles?: string[]; ids?: string[]} = {},
```

packages/api-client/src/main/user/UserAPI.ts
```diff
@@ -254,7 +256,28 @@ export class UserAPI {
...
+      const uniqueHandles = removeDuplicates(handles);
+      const handleChunks = chunk(uniqueHandles, limit);
+      const tasks = handleChunks.map(handleChunk => this._getUsers({handles: handleChunk}));
+      return Promise.all(tasks).then(flatten);
+    } else if (ids && ids.length) {
+      const uniqueIds = removeDuplicates(ids);
+      const idChunks = chunk(uniqueIds, limit);
+      const tasks = idChunks.map(idChunk => this._getUsers({ids: idChunk}));
+      return Promise.all(tasks).then(flatten);
+    } else {
```
ffflorian: `else` is not needed

packages/api-client/src/main/user/UserAPI.ts
Yserz: It's exclusive so it is not optional

ffflorian: Then we should mark this as exclusive with function overloading.

packages/api-client/src/main/user/UserAPI.ts
```diff
@@ -254,7 +256,29 @@ export class UserAPI {
...
+    limit: number = UserAPI.DEFAULT_USERS_CHUNK_SIZE
+  ): Promise<User[]> {
+    const {handles, ids} = parameters;
+
+    if (handles && handles.length) {
+      const uniqueHandles = removeDuplicates(handles);
+      const handleChunks = chunk(uniqueHandles, limit);
+      const tasks = handleChunks.map(handleChunk => this._getUsers({handles: handleChunk}));
+      return Promise.all(tasks).then(flatten);
+    } else if (ids && ids.length) {
```
ffflorian:
```suggestion
    }
    
    if (ids && ids.length) {
```

packages/api-client/src/main/user/UserAPI.ts
```diff
@@ -254,7 +256,29 @@ export class UserAPI {
...
+      const uniqueIds = removeDuplicates(ids);
+      const idChunks = chunk(uniqueIds, limit);
+      const tasks = idChunks.map(idChunk => this._getUsers({ids: idChunk}));
+      return Promise.all(tasks).then(flatten);
+    }
+    return [];
```
ffflorian:
```suggestion

    return [];
```

packages/api-client/src/main/user/UserAPI.ts
```diff
@@ -254,7 +256,29 @@ export class UserAPI {
    * @param parameters Multiple user's handles or IDs
    * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/users/users
    */
-  public getUsers(parameters: {handles?: string[]; ids?: string[]}): Promise<User[]> {
+  public async getUsers(parameters: {ids: string[]}, limit?: number): Promise<User[]>;
+  public async getUsers(parameters: {handles: string[]}, limit?: number): Promise<User[]>;
+  public async getUsers(
+    parameters: {ids?: string[]; handles?: string[]},
+    limit: number = UserAPI.DEFAULT_USERS_CHUNK_SIZE
```
ffflorian: 👍

PR #1972, opened 2019-06-24
feat(tslint-config): Member ordering

lipis:
Alphabetise would be nice.. but there is a bug!

- https://github.com/palantir/tslint/issues/4775
- https://palantir.github.io/tslint/rules/member-ordering/


bennyn:
If "member-ordering" can break the code we should refrain from using it.

lipis:
It's breaking the code when using the `alphabetize` and currently we are not using it.. https://github.com/wireapp/wire-web-packages/pull/1972/files#diff-0544b88677373dc8c9d0e54914d1c14b

PR #1995, opened 2019-06-27
feat(api-client): Precise property typings, cleanup

ffflorian:
## Pull Request Checklist


packages/api-client/src/main/user/UserAPI.ts
```diff
@@ -409,7 +409,7 @@ export class UserAPI {
    * @param propertyData The property data to set
    * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/users/setProperty
    */
-  public async putProperty(propertyKey: string, propertyData: Object): Promise<void> {
+  public async putProperty<T extends string>(propertyKey: T, propertyData: Record<T, any>): Promise<void> {
```
ffflorian: This means that if you call `getProperty('foo')` the resulting object will be `{ foo: ... }`.

PR #2003, opened 2019-06-28
feat(api-client): CallConfig endpoint

Yserz:


packages/api-client/src/main/account/CallConfigData.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export interface ICECandidate {
+  urls: string[];
+  username: string;
+  credential: string;
```
bennyn: sorting

PR #2008, opened 2019-06-28
fix(core): Send call messages

bennyn:


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -902,6 +914,11 @@ export class ConversationService {
     return userId;
   }
 
+  /**
+   * @param {Message} payloadBundle - Outgoing message
```
ffflorian: Types in JSDoc are not necessary in TS.

PR #2051, opened 2019-07-09
feat(core): Fetch unencrypted image v3 assets

bennyn:


packages/api-client/src/main/asset/AssetAPI.ts
```diff
@@ -30,7 +30,7 @@ export class AssetAPI {
 
   constructor(private readonly client: HttpClient) {}
 
-  getAsset(assetId: string, token?: string | null): Promise<ArrayBuffer> {
```
ffflorian: This change won't work since the protobuf will call `getAsset` and there `token` can be `null`.

PR #2078, opened 2019-07-16
refactor(tslint-config): No namespaces

Yserz:


bennyn:
Will the Proteus changes impact the following checks within our webapp? https://github.com/wireapp/wire-webapp/blob/2019-07-11-13-18/src/script/cryptography/CryptographyRepository.js#L488-L490

Yserz:
These should be fine. Only directly checking for `ProteusErrors.DecryptError` requires a change to `ProteusErrors.DecryptError.BaseError`. (I planned to make this a breaking change anyways)

bennyn:
This is a big change in behaviour. There is a test case which states it "assures that specializations are the same type as generics". This PR will change that behaviour so my questions are:

a) Do we need a `BaseError` at all?
b) If "a)" is a no, then should we remove the test case mentioned?

packages/proteus/src/main/errors/DecryptError.ts
```diff
@@ -17,86 +17,84 @@
...
+  CASE_204 = 204,
+  CASE_205 = 205,
+  CASE_206 = 206,
+  CASE_207 = 207,
+  CASE_208 = 208,
+  CASE_209 = 209,
+  CASE_210 = 210,
+  CASE_211 = 211,
+  CASE_212 = 212,
+  CASE_213 = 213, // REMOTE_ENCRYPTION_FAILURE: Created by the receiver when getting a decryption failure symbol (💣)
```
ffflorian:
```suggestion
  /** REMOTE_ENCRYPTION_FAILURE: Created by the receiver when getting a decryption failure symbol (💣) */
  CASE_213 = 213,
```

bennyn: This will help VS Code rendering the info, right?

ffflorian: Exactly :)

lipis:
What's the latest with this one?

packages/tslint-config/tslint.json
```diff
@@ -22,6 +22,7 @@
     "no-duplicate-variable": [true, "check-parameters"],
     "no-floating-promises": true,
     "no-invalid-template-strings": true,
+    "no-namespace": true,
```
lipis: Becomes `@typescript-eslint/no-namespace` in the ESLint

lipis: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md

bennyn:
Will be attacked with https://wearezeta.atlassian.net/browse/WEBAPP-6794

PR #2085, opened 2019-07-18
fix(core): Return image data instead of protobuf

ffflorian:


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -505,7 +503,6 @@ export class ConversationService {
 
     return {
       ...payloadBundle,
-      content: assetMessage as AssetContent,
```
ffflorian: The actual fix

packages/core/src/main/Account.ts
```diff
@@ -521,6 +521,7 @@ export class Account extends EventEmitter {
         } = genericMessage[GenericMessageType.ASSET];
         const isImage = !!uploaded && !!uploaded.assetId && !!original && !!original.image;
 
+        //TODO: isImage ? ImageAssetContent
```
bennyn: What does that mean?

PR #2104, opened 2019-07-22
feat(proteus): Use libsodium for key conversion (WEBAPP-6791)

ffflorian:
This speeds up key conversion 10 times:

```
Setup:
  4x Intel(R) Core(TM) i5-5200U CPU @ 2.20GHz
  15.4 GB RAM
  Linux 4.19.0-8-amd64 (x64)

Running benchmarks with 100,000 iterations each ...
Duration for ed2curve.convertPublicKey: 189035.83 ms
Duration for sodium.crypto_sign_ed25519_pk_to_curve25519: 19818.53 ms
```

Also, this needed some test adjustments as the old public keys for testing were invalid (filled with zeros) and the Test Store was not behaving correctly.

## Pull Request Checklist


packages/proteus/src/main/session/Session.ts
```diff
@@ -259,7 +261,8 @@ export class Session {
 
         if (msg.prekey_id !== PreKey.MAX_PREKEY_ID) {
           // TODO: Zeroize should be tested (and awaited) here!
```
bennyn, line 263: Can the TODO now be removed? I see that we are awaiting it here.

PR #2165, opened 2019-08-07
fix(api-client): Remove possibility to pass URL to configure identity provider

Yserz:


ffflorian:
This will be a breaking change, not a chore ;)

Yserz:
Actually that never worked so it's more like a fix (and non-breaking)

ffflorian:
Alright!

PR #2175, opened 2019-08-08
fix(store-engine,store-engine-sqleet): Verify that strings and objects can be saved

bennyn:
## Pull Request Checklist

- [X] My code is covered by tests
- [X] I will [merge the PR as breaking change](https://github.com/wireapp/wire-web-packages/wiki/Releases#create-major-release), if the API contract changes

### DIFF
--- a/packages/store-engine-sqleet/src/SchemaConverter.ts
+++ b/packages/store-engine-sqleet/src/SchemaConverter.ts
@@ -27,6 +27,7 @@ export enum SQLiteType {
   INTEGER = 'integer',
   // See https://stackoverflow.com/a/8417411
   JSON = 'json',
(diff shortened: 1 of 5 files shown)


packages/store-engine-sqleet/src/SchemaConverter.ts
```diff
@@ -20,20 +20,24 @@
 const uint32 = require('uint32');
 
 export const SQLeetEnginePrimaryKeyName: string = '`key`';
+export const MAGIC_SINGLE_COLUMN = '__sc_value';
```
bennyn: Let's think of a more meaningful name. How about "RESERVED_SINGLE_COLUMN"`? Is "__sc_value" the column name?

PR #2224, opened 2019-08-26
feat(store-engine-dexie): Register persistent storage and call isSupported on initWithDb

ffflorian:
While this seems to work out-of-the-box in Electron, for Chrome there are some requirements, see https://developers.google.com/web/updates/2016/06/persistent-storage.

## Pull Request Checklist


packages/store-engine-dexie/src/index.ts
```diff
@@ -70,13 +70,54 @@ export class IndexedDBEngine implements CRUDEngine {
...
+      if (granted) {
+        console.info('Storage will not be cleared except by explicit user action');
+        return true;
+      }
+
+      console.info('Storage may be cleared by the UA under storage pressure.');
+      return false;
+    } catch (error) {
+      console.error(error);
+      return false;
```
bennyn: Let's fail when it cannot be setup while the code path wants to set it up. Otherwise we might continue in a state we cannot control.

packages/store-engine-dexie/src/index.ts
```diff
@@ -70,13 +70,54 @@ export class IndexedDBEngine implements CRUDEngine {
...
     await this.isSupported();
-    return this.assignDb(new Dexie(storeName));
+    const dexie = await this.assignDb(new Dexie(storeName));
+    if (registerPersisted) {
+      await this.registerPersistentStorage();
+    }
+    return dexie;
+  }
+
+  public async initWithDb(db: Dexie, registerPersisted?: boolean): Promise<Dexie> {
```
bennyn:
```suggestion
  public async initWithDb(db: Dexie, registerPersisted: boolean = false): Promise<Dexie> {
```

PR #2235, opened 2019-08-28
refactor(store-engine-sqleet): Replace forked sql.js with websql

pwnsdx:


packages/store-engine-sqleet/src/index.test.ts
```diff
@@ -302,7 +298,7 @@ describe('SQLeetEngine', () => {
       }
     });
 
-    it('exports and loads a database', async () => {
+    /*it('exports and loads a database', async () => {
```
bennyn: We don't commit commented code.

packages/store-engine-sqleet/src/index.ts
```diff
@@ -120,31 +102,16 @@ export class SQLeetEngine implements CRUDEngine {
...
-    for (let i = 0; i < database.length; i++) {
-      databaseBinary[i] = database.charCodeAt(i);
-    }
-    return databaseBinary;
+    return this.db.export('string');
   }
 
-  async purge(): Promise<void> {
-    // Databases must be closed, when you're finished with them, or the memory consumption will grow forever.
+  public async purge(): Promise<void> {
```
bennyn:
```suggestion
  async purge(): Promise<void> {
```

Public is already the default.

packages/store-engine-sqleet/src/index.test.ts
```diff
@@ -27,25 +26,22 @@ import {readAllSpec} from '@wireapp/store-engine/dist/commonjs/test/readAllSpec'
...
-  const webAssembly = Decoder.fromBase64(SQLeetWebAssembly).asBytes;
   const STORE_NAME = 'wire@production@52c607b1-4362-4b7b-bcb4-5bff6154f8e2@permanent';
   const GENERIC_ENCRYPTION_KEY = 'test';
   let engine: SQLeetEngine | undefined = undefined;
 
-  async function initEngine(scheme: {}, shouldCreateNewEngine = true, rawDatabase?: string): Promise<SQLeetEngine> {
+  async function initEngine(scheme: {}, shouldCreateNewEngine = true): Promise<SQLeetEngine> {
     if (!engine || shouldCreateNewEngine) {
-      engine = new SQLeetEngine(webAssembly, scheme, GENERIC_ENCRYPTION_KEY, rawDatabase);
+      engine = new SQLeetEngine('./base/websql-worker.js', scheme, GENERIC_ENCRYPTION_KEY);
```
bennyn: Let's make the worker path configurable. Please add it as third parameter to the function with a default value of `'./base/websql-worker.js'`.

packages/store-engine-sqleet/src/index.ts
```diff
@@ -276,34 +248,36 @@ export class SQLeetEngine implements CRUDEngine {
...
-    }
+    const statement = await this.db.prepare(selectRecordStatement);
+    const records = (await statement.getAsObject()) as T[];
 
-    return records.values.map((record: any[]) => zipObject(records.columns, record));
+    return records;
   }
 
   async readAllPrimaryKeys(tableName: string): Promise<string[]> {
+    if (!this.db) {
```
bennyn: Please remove the ` if (!this.db)` checks. They are expectional. In most of the cases the database will be instantiated but then every db call makes this check. I guess the database will on `this.db.execute` already when there is no `this.db`, so we don't need to make the extra check for `this.db`.

pwnsdx: Thanks for pointing out. We used to have the this.db assignment in the init() therefore db was not set in the constructor (hence the extra ifs I added if we wanted to enable typescript types). Now, due to the changes in the websql API I was able to move the this.db assignment directly to the constructor (and remove the ifs).

packages/store-engine-sqleet/karma.conf.ts
```diff
@@ -38,14 +38,19 @@ module.exports = (config: Config): void => {
...
+    },
     logLevel: config.LOG_INFO,
     port: 9876,
     preprocessors: {
       '**/*.ts': ['karma-typescript'],
     },
     reporters: ['progress', 'karma-typescript'],
     singleRun: true,
-  });
+  } as unknown) as ConfigOptions);
```
ffflorian: `unknown` is not needed here

packages/store-engine-sqleet/package.json
```diff
@@ -4,8 +4,7 @@
     "@babel/plugin-proposal-class-properties": "7.5.5",
     "@babel/plugin-proposal-object-rest-spread": "7.5.5",
     "@babel/preset-typescript": "7.3.3",
-    "@types/sql.js": "1.0.2",
-    "sql.js": "https://github.com/pwnsdx/sqleet.js#8075c8ed6a95f6fe7ab7a41703b2fcc3a412be2f",
+    "@wireapp/websql": "https://github.com/wireapp/websql#6425db5537739ff106d37c709ab0e1275d71eb18",
```
ffflorian: Why not create a release?

packages/store-engine-sqleet/src/index.ts
```diff
@@ -89,62 +78,29 @@ export class SQLeetEngine implements CRUDEngine {
...
-  }
-
-  private async load(database: string): Promise<Uint8Array | undefined> {
-    const databaseBinary = new Uint8Array(database.length);
-    for (let i = 0; i < database.length; i++) {
-      databaseBinary[i] = database.charCodeAt(i);
-    }
-    return databaseBinary;
+  async export<T>(): Promise<string> {
+    return (this.db.export('utf8') as unknown) as string;
```
ffflorian:
```suggestion
    return this.db.export('utf8') as any;
```

packages/store-engine-sqleet/src/index.ts
```diff
@@ -89,62 +78,29 @@ export class SQLeetEngine implements CRUDEngine {
...
-    return strings.join('');
-  }
-
-  private async load(database: string): Promise<Uint8Array | undefined> {
-    const databaseBinary = new Uint8Array(database.length);
-    for (let i = 0; i < database.length; i++) {
-      databaseBinary[i] = database.charCodeAt(i);
-    }
-    return databaseBinary;
+  async export<T>(): Promise<string> {
```
ffflorian:
```suggestion
  async export(): Promise<string> {
```

packages/store-engine-sqleet/src/index.ts
bennyn: I think here it's ok to cast into `string` since the function returns `Promise<Uint8Array | string>` and `any` would be to broad in that case.

packages/store-engine-sqleet/package.json
bennyn: I created one now: https://www.npmjs.com/package/@wireapp/websql

PR #2241, opened 2019-08-29
feat(core): Return array when sending (BREAKING)

ffflorian:
BREAKING CHANGE: `conversation.send()` now returns an array with the message as first (and only) item.

## Pull Request Checklist

### DIFF
--- a/packages/bot-api/src/main/Bot.ts
+++ b/packages/bot-api/src/main/Bot.ts
@@ -70,9 +70,7 @@ export class Bot {
 
   public async sendText(conversationId: string, message: string): Promise<void> {
     if (this.account && this.account.service) {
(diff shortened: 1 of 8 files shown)


packages/bot-api/src/main/Bot.ts
```diff
@@ -70,9 +70,7 @@ export class Bot {
 
   public async sendText(conversationId: string, message: string): Promise<void> {
     if (this.account && this.account.service) {
-      const textPayload = await this.account.service.conversation.messageBuilder
-        .createText(conversationId, message)
-        .build();
+      const textPayload = this.account.service.conversation.messageBuilder.createText(conversationId, message).build();
```
atomrc: I don't see any signature change for the method `build`. Is this normal?

PR #2273, opened 2019-09-06
refactor(api-client): Add initial notification payloads

bennyn:


packages/api-client/src/main/client/ClientType.ts
```diff
@@ -18,6 +18,7 @@
  */
 
 export enum ClientType {
+  /** 'none' is an artificial client type, not sent by the backend */
```
ffflorian: :+1:

packages/api-client/src/main/event/ConversationEvent.ts
```diff
@@ -125,6 +125,18 @@ export interface ConversationOtrMessageAddEvent extends ConversationEvent {
...
+export interface ConversationOtrMessageAddNotification {
+  conversation: string;
+  time: string;
+  data: {
+    text: string;
+    sender: string;
+    recipient: string;
+  };
+  from: string;
+  type: CONVERSATION_EVENT.OTR_MESSAGE_ADD;
```
ffflorian: Sorting

packages/api-client/src/main/event/UserEvent.ts
```diff
@@ -119,3 +125,12 @@ export interface UserPropertiesSetEvent extends UserEvent {
...
   type: USER_EVENT.UPDATE;
 }
+
+export interface UserUpdateNotification {
+  user: {
+    picture?: Picture[];
+    id: string;
+    assets?: UserAsset[];
+  };
+  type: USER_EVENT.UPDATE;
```
ffflorian: Sorting

packages/api-client/src/main/client/RegisteredClient.ts
```diff
@@ -32,5 +33,10 @@ export interface RegisteredClient {
   model?: string;
   /** An ISO 8601 Date string */
   time: string;
-  type: 'permanent' | 'temporary';
+  type: ClientType.PERMANENT | ClientType.TEMPORARY;
 }
+
+export type RegisteredClient = UpdatedClient & {
+  /** The cookie label */
+  cookie: string;
```
Yserz: I think cookie is defined twice now

bennyn: Good catch! Fixed.

packages/api-client/src/main/event/UserEvent.ts
```diff
@@ -54,6 +55,11 @@ export interface UserClientAddEvent extends UserEvent {
   type: USER_EVENT.CLIENT_ADD;
 }
 
+export interface UserClientAddNotification {
```
Yserz: It seems the naming is inconsistent now. Some interfaces are Events and some are Notifications now. Is there a difference or is it the same?

bennyn: It's on purpose. I put the notifications next to the events because they are similar and it's an indication for us to check if every new event also has a matching notification in the notification stream.

Yserz: We discussed this in 1:1 and came to the conclusion to rename the `*Events` to something else because these are actually no events. Instead they are synchronous responses to REST calls.

Yserz: https://github.com/wireapp/wire-pwa/issues/1321

PR #2281, opened 2019-09-09
fix: Re-add WebSocket ping

Yserz:


packages/api-client/src/main/tcp/ReconnectingWebsocket.ts
```diff
@@ -39,7 +39,15 @@ export enum WEBSOCKET_STATE {
...
 
+export enum PingMessage {
+  PING = 'ping',
+  PONG = 'pong',
+}
+
 export class ReconnectingWebsocket {
+  private static readonly CONFIG = {
+    PING_INTERVAL: TimeUtil.TimeInMillis.SECOND * 20,
+  };
```
ffflorian: Why not move this outside of the class? It's static and private.

packages/api-client/src/main/tcp/ReconnectingWebsocket.ts
```diff
@@ -118,6 +139,30 @@ export class ReconnectingWebsocket {
     }
   }
 
+  private startPinging(): void {
+    this.stopPinging();
```
ffflorian: Stops pinging after the start?

Yserz: No, that's stopping before starting it. There are cases where we had 2 pingers running, that's why the failed ping check was actually failing.

packages/api-client/src/main/tcp/ReconnectingWebsocket.ts
Yserz: Because it's private to `ReconnectingWebsocket` more than to the file.

PR #2284, opened 2019-09-10
test: WebSocket ping

Yserz:


packages/api-client/src/main/tcp/ReconnectingWebsocket.test.node.js
```diff
@@ -50,22 +63,26 @@ function startEchoServer() {
...
+
+  beforeEach(() => {
+    this.server = startEchoServer();
+  });
 
   afterEach(done => {
-    if (server) {
-      server.close(() => {
-        server = undefined;
+    if (this.server) {
```
ffflorian: Let's keep the global variable.

Yserz:
It doesn't connect to a live server. I just missed a `this`.

packages/api-client/src/main/tcp/ReconnectingWebsocket.test.node.js
```diff
@@ -19,15 +19,28 @@
...
-const WEBSOCKET_URL = `ws://127.0.0.1:${WEBSOCKET_PORT}`;
-let server = undefined;
+const reservedPorts = [];
 
 function startEchoServer() {
-  server = new WebSocketServer({port: WEBSOCKET_PORT});
+  const getWebsocketPort = () => {
+    const WEBSOCKET_START_PORT = 8087;
+    let currentPort = WEBSOCKET_START_PORT;
+    while (true) {
```
ffflorian: Can't we also do `while(currentPort++)` here?

ffflorian: https://github.com/wireapp/wire-web-packages/pull/2286/files

PR #2309, opened 2019-09-19
feat(api-client): Extend notification payloads

bennyn:


packages/api-client/src/main/conversation/Conversation.ts
```diff
@@ -48,6 +48,6 @@ export interface Conversation {
   members: ConversationMembers;
   message_timer: number | null;
   name: string;
-  team?: string;
+  team?: string | null;
```
Yserz: optional covers `null` already

Yserz: In general I don't like a distinction between `null` and `undefined` here because there should be none. This statement can lead to unexpected interpretations of null/undefined.

packages/api-client/src/main/event/ConversationEvent.ts
```diff
@@ -80,25 +81,73 @@ export interface ConversationConnectRequestEvent extends ConversationEvent {
...
+
 export interface ConversationMemberJoinEvent extends ConversationEvent {
   data: MemberJoin;
   type: CONVERSATION_EVENT.MEMBER_JOIN;
 }
 
+export interface ConversationMemberJoinNotification {
+  conversation: string;
+  data: {
+    user_ids: ['6f88716b-1383-44da-9d57-45b51cc64d90'];
```
Yserz: Is this fixed id wanted?

packages/api-client/src/main/user/data/UserPushRemoveData.ts
```diff
@@ -0,0 +1,25 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export interface UserPushRemoveData {
+  app: string;
+  client: string;
+  token: string;
+  transport: 'GCM';
```
Yserz: Is it always GCM? what does it mean?

packages/api-client/src/main/conversation/Conversation.ts
bennyn: Why to you think it can led to unexpected interpretations? The typings now reflect 100% the behaviour of the backend because the backend sends this property with a value of `null`. It never sends `undefined` or leaves out the property completely.

packages/api-client/src/main/event/ConversationEvent.ts
bennyn: No, good catch.

packages/api-client/src/main/user/data/UserPushRemoveData.ts
AndyLnd: GCM = Google Cloud Messaging

bennyn: So far I only saw 'GCM' but reading the [PushTokens spec](https://github.com/wearezeta/backend-api-docs/wiki/API-User-PushTokens#transport) there can be also `GCM`, `APNS`, `APNS_SANDBOX`, `APNS_VOIP` & `APNS_VOIP_SANDBOX`. I will add them too. 👍

packages/api-client/src/main/conversation/Conversation.ts
Yserz: The backend doesn't really care about the differences of undefined/null what they want to express is that there is no value. So we should do the same.

bennyn: What we should do is creating an accurate typing of the backend payload. Does the backend return `undefined` for the "team" property? No. Does the backend return `null` for the "team" property? Yes. So creating a typing which allows `undefined` is just wrong.

PR #2324, opened 2019-09-25
refactor(core): Simplify database access for notifications

bennyn:


packages/core/src/main/notification/NotificationService.ts
```diff
@@ -52,29 +52,22 @@ export class NotificationService {
...
+      databaseLastEventDate = await this.database.getLastEventDate();
     } catch (error) {
-      if (
-        error instanceof StoreEngineError.RecordNotFoundError ||
-        error.constructor.name === StoreEngineError.RecordNotFoundError.constructor.name
-      ) {
-        return this.database.createLastEventDate(eventDate);
-      }
-      throw error;
+      return this.database.createLastEventDate(eventDate);
```
Yserz: Wouldn't that mean that we always try to create a lastEventDate no matter what kind of error is thrown?

bennyn: Indeed. I will add the check for the error type back.

PR #2330, opened 2019-09-26
feat(api-client,core): Get notifications from notification service

bennyn:


packages/core/src/main/notification/NotificationBackendRepository.ts
```diff
@@ -18,11 +18,31 @@
  */
 
 import {APIClient} from '@wireapp/api-client';
-import {Notification} from '@wireapp/api-client/dist/commonjs/notification/';
+import {NOTIFICATION_SIZE_MAXIMUM, Notification} from '@wireapp/api-client/dist/commonjs/notification/';
 
 export class NotificationBackendRepository {
   constructor(private readonly apiClient: APIClient) {}
 
+  public async getAllNotifications(clientId?: string, lastNotificationId?: string): Promise<Notification[]> {
```
Yserz: Should only call `apiClient.getAllNotifications...`

PR #2387, opened 2019-10-07
fix: Remove decryption result wrapper

bennyn:
## Pull Request Checklist


packages/core/src/main/cryptography/CryptographyService.ts
```diff
@@ -84,30 +74,10 @@ export class CryptographyService {
...
-      if (isOutdatedMessage || isDuplicateMessage) {
-        return {
-          error,
-          isSuccess: false,
-        };
-      }
-
-      throw error;
-    }
+    return this.cryptobox.decrypt(sessionId, messageBytes.buffer);
```
ffflorian: Note: Behavior is now changed here, since we did `await ... return` before. Now we just do `return`.

PR #2402, opened 2019-10-09
refactor(core): Define call signature for send

bennyn:


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -978,8 +978,7 @@ export class ConversationService {
    * @param userIds - Only send message to specified user IDs
    * @returns Sent message
    */
-  // tslint:disable-next-line:typedef
-  public async send(payloadBundle: Message, userIds?: string[]) {
+  public async send(payloadBundle: Message, userIds?: string[]): Promise<Message> {
```
Yserz: Without the forced `Promise<Message>` typescript can potentially figure out the type of the `Message` based on the input message and the `Message type`.

PR #2416, opened 2019-10-10
fix(store-engine-sqleet): Create Node.js database directory before using it

bennyn:


.github/workflows/main.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           contains(github.event.commits[0].message, '[ci skip]') == false
         uses: actions/setup-node@v1
         with:
-          node-version: 10
+          node-version: 12
```
ffflorian: Let's move this to a second PR.

PR #2488, opened 2019-10-29
refactor(core): Infer message content by message type

bennyn:


packages/core/src/main/conversation/message/TeamMessage.ts
```diff
@@ -0,0 +1,32 @@
...
+import {TeamMemberLeaveData} from '@wireapp/api-client/dist/commonjs/team/data';
+import {BasePayloadBundle, PayloadBundleType} from './PayloadBundle';
+
+export interface TeamMemberLeaveMessage extends BasePayloadBundle {
+  content: TeamMemberLeaveData;
+  type: PayloadBundleType.TEAM_MEMBER_LEAVE;
+}
+
+export interface TeamMemberJoinMessage extends BasePayloadBundle {
+  type: PayloadBundleType.TEAM_MEMBER_JOIN;
```
ffflorian: No content?

bennyn: It's true, this one does not have any content yet: https://github.com/wireapp/wire-web-packages/blob/0dd111f574ac9b91872a601592f1eb0f2eeab42b/packages/api-client/src/main/event/TeamEvent.ts#L82-L85

PR #2503, opened 2019-11-01
feat(store-engine-sqleet): Remove database directory on purge

bennyn:


packages/store-engine-sqleet/src/SQLeetEngine.ts
```diff
@@ -106,6 +106,14 @@ export class SQLeetEngine implements CRUDEngine {
     if (this.db) {
       await this.db.close();
       await this.db.wipe(this.storeName);
+      if (this.nodeDatabaseDir) {
+        fs.rmdirSync(this.nodeDatabaseDir, {recursive: true});
```
ffflorian: Why not use `async` here, too?

```suggestion
        await fs.rmdir(this.nodeDatabaseDir, {recursive: true});
```

AndyLnd: This is not using `fs-extra` so rmdir doesn't return a promise.

ffflorian: Ah sorry, wrong suggested change. Here is the correct one:
```suggestion
        await fs.promises.rmdir(this.nodeDatabaseDir, {recursive: true});
```

bennyn: Why is ` await fs.promises.rmdir` better than `fs.rmdirSync`?

ffflorian: Because `fs.rmdirSync` will block the whole I/O while the `async` solution is non-blocking.

PR #2576, opened 2019-11-21
fix(api-client): Set asset request priority to low

Yserz:


packages/api-client/src/main/asset/AssetAPI.ts
```diff
@@ -17,6 +17,7 @@
  *
  */
 
+import {Priority} from '@wireapp/priority-queue';
 import {AxiosRequestConfig} from 'axios';
 import {HttpClient} from '../http/';
```
ffflorian:
```suggestion

import {HttpClient} from '../http/';
```

PR #2578, opened 2019-11-22
fix(api-client): Legal hold request event

Yserz:


packages/api-client/src/main/user/data/UserClientLegalHoldRequestData.ts
```diff
@@ -19,9 +19,9 @@
...
 import {PreKey} from '../../auth';
 
-export interface UserClientLegalHoldRequestData {
-  client_id: string;
+export interface UserLegalHoldRequestData {
+  client: string;
   last_prekey: PreKey;
-  requester: string;
-  target_user: string;
+  // The ID of the user that is targeted (formally 'target_user')
```
ffflorian:
```suggestion
  /** The ID of the user that is targeted (formally `target_user`) */
```

PR #2589, opened 2019-11-25
docs(api-client): Reuse Cookie in demo script

bennyn:


packages/api-client/demo.ts
```diff
@@ -21,8 +21,17 @@ import {FileEngine} from '@wireapp/store-engine-fs';
 import logdown from 'logdown';
 import path from 'path';
```
ffflorian:
```suggestion
import {CRUDEngine} from '@wireapp/store-engine';

```

packages/api-client/demo.ts
```diff
@@ -21,8 +21,17 @@ import {FileEngine} from '@wireapp/store-engine-fs';
...
 
+import {CRUDEngine} from '@wireapp/store-engine';
```
ffflorian:
```suggestion
```

PR #2647, opened 2019-12-11
refactor(api-client,core,cryptobox,priority-queue): Use Karma with TypeScript

bennyn:


packages/api-client/karma.conf.ts
```diff
@@ -16,16 +16,17 @@
  * along with this program. If not, see http://www.gnu.org/licenses/.
  *
  */
-
+import {Config} from 'karma';
```
ffflorian:
```suggestion

import {Config} from 'karma';
```

PR #2716, opened 2020-01-08
feat(api-client): Add putConversationReceiptMode

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/conversation/ConversationAPI.ts
+++ b/packages/api-client/src/conversation/ConversationAPI.ts
@@ -36,10 +36,16 @@ import {
   ConversationMemberLeaveEvent,
   ConversationMessageTimerUpdateEvent,
   ConversationRenameEvent,
+  ConversationReceiptModeUpdateEvent,
 } from '../event/';
 import {HttpClient} from '../http/';
 import {ValidationError} from '../validation/';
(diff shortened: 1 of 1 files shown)


packages/api-client/src/conversation/ConversationAPI.ts
```diff
@@ -459,13 +467,33 @@ export class ConversationAPI {
...
     };
 
     const response = await this.client.sendJSON<ConversationMessageTimerUpdateEvent>(config);
     return response.data;
   }
 
+  /**
+   * Update the receipt mode for a conversation.
+   * @param conversationId The conversation ID
+   * @param receiptModeData The new receipt mode
```
bennyn: Why not calling it just "receiptMode"?

ffflorian: Because the interface is called `ConversationReceiptModeUpdateData` and includes `receipt_mode`. I also think it could maybe be a bit too much but I wanted to streamline it with other functions which also take an object. Should I update the text / the parameter name?

Yserz: I would argue that this is signaling that it is an object and not just a value. Also a common pattern we already have in the api-client.

PR #2720, opened 2020-01-08
feat(api-client): Add conversation roles and TeamConversationAPI

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/APIClient.ts
+++ b/packages/api-client/src/APIClient.ts
@@ -46,7 +46,7 @@ import * as ObfuscationUtil from './obfuscation/';
 import {SelfAPI} from './self/';
 import {WebSocketClient} from './tcp/';
 import {
-  ConversationAPI as TeamConversationAPI,
+  TeamConversationAPI,
   FeatureAPI,
   IdentityProviderAPI,
   LegalHoldAPI,
--- a/packages/api-client/src/APIClient.ts
+++ b/packages/api-client/src/APIClient.ts
@@ -46,6 +46,7 @@ import * as ObfuscationUtil from './obfuscation/';
 import {SelfAPI} from './self/';
 import {WebSocketClient} from './tcp/';
 import {
+  ConversationAPI as TeamConversationAPI,
   FeatureAPI,
   IdentityProviderAPI,
   LegalHoldAPI,
@@ -90,6 +91,7 @@ export class APIClient extends EventEmitter {
   public notification: {api: NotificationAPI};
   public self: {api: SelfAPI};
   public teams: {
+    conversation: {api: TeamConversationAPI};
     feature: {api: FeatureAPI};
     identityProvider: {api: IdentityProviderAPI};
     invitation: {api: TeamInvitationAPI};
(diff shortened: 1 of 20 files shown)


packages/api-client/src/conversation/ConversationRole.ts
```diff
@@ -0,0 +1,38 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export enum DefaultConversationRole {
+  WIRE_ADMIN = 'wire_admin',
+  WIRE_MEMBER = 'wire_member',
+}
+
+export interface ConversationRole {
```
bennyn:
```suggestion
export interface ConversationPermission {
```

packages/api-client/src/conversation/ConversationRole.ts
```diff
@@ -0,0 +1,38 @@
...
+  /** The set of actions allowed for this role */
+  actions: string[];
+  /**
+   * Role name, between 2 and 128 chars, 'wire_' prefix is reserved for
+   * roles designed by Wire (i.e., no custom roles can have the same prefix).
+   * @see `DefaultConversationRole`
+   */
+  conversation_role: string;
```
bennyn:
```suggestion
  conversation_role: string | DefaultConversationRole;
```

packages/api-client/src/APIClient.ts
```diff
@@ -46,6 +46,7 @@ import * as ObfuscationUtil from './obfuscation/';
 import {SelfAPI} from './self/';
 import {WebSocketClient} from './tcp/';
 import {
+  ConversationAPI as TeamConversationAPI,
```
bennyn: I propose to rename the original `ConversationAPI` class.

packages/api-client/src/conversation/ConversationRole.ts
ffflorian: What's the resoning behind this?

bennyn: The conversation role is "wire_admin" or "wire_member". The conversation permission is the connection of conversation role and allowed actions.

ffflorian: > The conversation permission is the connection of conversation role and allowed actions.

But everywhere else the endpoints and interfaces which include this interface are calling it `ConversationRole` (see https://staging-nginz-https.zinfra.io/swagger-ui/#!/conversations/getConversationsRoles). Should we maybe rename `DefaultConversationRole` to `DefaultConversationRoleName` to avoid confusion?

In general, I try to apply the naming from wire-server, in this case I got it from [Roles.hs#L82](https://github.com/wireapp/wire-server/blob/e7e57d809253a3c55e7f0b3a01034c46647b0574/libs/galley-types/src/Galley/Types/Conversations/Roles.hs#L82)

bennyn: Ok, let's use `DefaultConversationRoleName` because I find it confusing that a `ConversationRole` interface has a member called `conversation_role`.

PR #2731, opened 2020-01-09
feat(api-client): Add headHandle & getUserByHandle endpoints

Yserz:


packages/api-client/src/user/UserAPI.ts
```diff
@@ -385,6 +385,35 @@ export class UserAPI {
...
+   * @param handle The handle to check
+   * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/users/checkUserHandle
+   */
+  public async headHandle(handle: string): Promise<void> {
+    const config: AxiosRequestConfig = {
+      method: 'head',
+      url: `${UserAPI.URL.USERS}/${UserAPI.URL.HANDLES}/${handle}`,
+    };
+
+    await this.client.sendJSON<string[]>(config);
```
ffflorian: `string[]` is not needed

```suggestion
    await this.client.sendJSON(config);
```

packages/api-client/src/user/UserAPI.ts
```diff
@@ -385,6 +385,35 @@ export class UserAPI {
...
+  }
+
+  /**
+   * Get a user by handle.
+   * @param handle The handle of a user to search for
+   * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/users/getUserByHandle
+   */
+  public async getUserByHandle(handle: string): Promise<{user: string}> {
```
ffflorian: Let's use [`HandleInfo`](https://github.com/wireapp/wire-web-packages/blob/master/packages/api-client/src/user/HandleInfo.ts).

PR #2754, opened 2020-01-15
chore(api-client): Remove long debug output

ffflorian:
This removes long outputs from the logs since it's not possible to hide `debug` outputs from logdown.

## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/tcp/ReconnectingWebsocket.ts
+++ b/packages/api-client/src/tcp/ReconnectingWebsocket.ts
@@ -103,7 +103,7 @@ export class ReconnectingWebsocket {
   };
 
   private readonly internalOnOpen = (event: Event) => {
-    this.logger.debug('WebSocket opened', event);
+    this.logger.debug('WebSocket opened');
     if (this.socket) {
       this.socket.binaryType = 'arraybuffer';
     }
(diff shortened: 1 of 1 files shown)


packages/api-client/src/tcp/ReconnectingWebsocket.ts
```diff
@@ -121,7 +121,7 @@ export class ReconnectingWebsocket {
   };
 
   private readonly internalOnClose = (event: CloseEvent) => {
-    this.logger.debug('WebSocket closed', event);
+    this.logger.debug('WebSocket closed');
```
Yserz: The data in the event is quite nice because it includes the close reason for example

ffflorian: Yes but it spams the terminal with a lot of unnecessary data.

Yserz: How about setting the debug level to info instead of debug?

ffflorian: Please point me to the part of logdown where I can set the debug level. :wink:

bennyn: I agree that we should remove the `event` from the log. It is a complex object which does not look good when being stringified. If someone is interested in that object the debugger in Chrome (or attached to Node.js) can be used.

Yserz: image

ffflorian: That's the Console level selector in Chrome, I was talking about terminals :)

PR #2796, opened 2020-01-24
feat(api-client): Handle & map domain not found error

Yserz:


packages/api-client/src/account/AccountError.ts
```diff
@@ -0,0 +1,36 @@
+/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
```
ffflorian:
```suggestion
 * Copyright (C) 2020 Wire Swiss GmbH
```

packages/api-client/src/http/BackendErrorLabel.ts
```diff
@@ -74,6 +74,8 @@ export enum BackendErrorLabel {
   NO_ADD_TO_MANAGED = 'no-add-to-managed',
   // Payment errors
   EXPIRED_CARD = 'expired_card',
+  // Domain errors
+  DOMAIN_NOT_FOUND = 'custom-backend-not-found',
```
ffflorian: Wouldn't it be better to name the enum after the content?

Yserz: I think it's okay this way because you enter a domain as input param. And in the error case the domain was not found.

ffflorian: I don't agree and I think we should stay as close as possible to the backend naming.

PR #2848, opened 2020-03-02
feat(api-client): Generate TypeDoc documentation for api-client

lipis:


bennyn:
Thanks for this great exploration. We will use it at a later point in time.

PR #2900, opened 2020-03-26
feat(api-client): Cancelable requests & progress handling

Yserz:


packages/api-client/src/http/RequestProgressHandler.ts
```diff
@@ -0,0 +1,10 @@
+export type ProgressCallback = (progress: number) => void;
+
+export const handleProgressEvent = (progressCallback?: ProgressCallback) => {
+  return (
+    progressCallback &&
+    ((progressEvent: any) => {
```
bennyn: Is there no definition of [ProgressEvent](https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent) in TypeScript?

Yserz: nope

Yserz: Axios defines any :/

bennyn: It seems to work, no? https://github.com/wireapp/wire-web-packages/pull/2900/commits/6ea5f80528f9128f61b24fe1aeb48bbeb565604f

Yserz: Ah that's what you mean. The thing is this event comes from Axios which specifies this event as any. So while this is typed now there is no guarantee that this isn't a different event.

Overall this seems to be good enough. Thx @bennyn :)

bennyn: Ok, I approved the PR. Let's see if we can also bring axios to the light side of the Force: https://github.com/axios/axios/pull/2851

PR #2909, opened 2020-03-30
refactor(react-ui-kit): Reduce derived colors

Yserz:
image


lipis:
Can we get all 12 colors on the same row?

PR #2912, opened 2020-03-31
feat(api-client): Get Members by ID list (BREAKING)

Yserz:
This endpoint is a draft and is currently not provided by the backend


Yserz:
BREAKING CHANGE: 
- Rename `getMembers` to `getAllMembers`

PR #2928, opened 2020-04-06
refactor(api-client): Improved search cancellation (BREAKING)

Yserz:


ffflorian:
When it's breaking things, this should be a feature, not a refactor.

packages/api-client/src/http/BackendErrorLabel.ts
```diff
@@ -95,6 +95,7 @@ export enum SyntheticErrorLabel {
   FORBIDDEN_PHONE_NUMBER = 'forbidden-phone-number',
   HANDLE_TOO_SHORT = 'handle-too-short',
   INVALID_PHONE_NUMBER = 'invalid-phone-number',
+  REQUEST_CANCELLATION = 'request-cancelled',
```
ffflorian, line 98: Shouldn't this rather be `REQUEST_CANCELLED`?

Yserz:
I disagree since it's an already available feature which got refactored. Putting a `BREAKING CHANGE` in the description will care about the correct versioning.

PR #2931, opened 2020-04-08
fix: Update husky commits

lipis:


lipis:
Yes, I just couldn't commit empty stuff and thought that was the error..

lipis:
Then I din't see any issues with it..

PR #2935, opened 2020-04-09
refactor(proteus): Don't generate empty keys in KeyPair constructor

bennyn:


packages/proteus/src/main/keys/KeyPair.ts
```diff
@@ -79,26 +79,24 @@ export class KeyPair {
...
+      const secretKey = SecretKey.decode(decoder);
+
+      decoder.u8();
+      const publicKey = PublicKey.decode(decoder);
+
+      return new KeyPair(publicKey, secretKey);
     }
 
-    return self;
+    throw new Error(`Unexpected number of properties: "${properties}"`);
```
ffflorian, line 100: Let's use a `DecodeError` here, shall we?

packages/proteus/src/main/session/SendChain.ts
```diff
@@ -18,7 +18,7 @@
  */
 
 import * as CBOR from '@wireapp/cbor';
-
+import {PublicKey, SecretKey} from '../keys';
```
ffflorian, line 21:
```suggestion

import {PublicKey, SecretKey} from '../keys';
```

bennyn: We can drop these extra lines.

packages/proteus/src/main/keys/KeyPair.ts
bennyn: Great idea.

PR #2938, opened 2020-04-09
refactor(proteus): Use constructor instead of ClassUtil (BREAKING) (WEBAPP-6789)

ffflorian:
This PR will
- remove the `ClassUtil` and use built-in constructors
- remove the switch case for decoding and just walk through the object instead
- convert snake_case to camelCase where possible without breaking things

**BREAKING CHANGES:**
- `Proteus.keys.PreKeyBundle.new()` ➡️ `new Proteus.keys.PreKeyBundle()`

## Pull Request Checklist


packages/proteus/src/main/derived/DerivedSecrets.ts
```diff
@@ -29,7 +27,7 @@ export class DerivedSecrets {
   mac_key: MacKey;
 
   constructor() {
-    this.cipher_key = new CipherKey();
+    this.cipher_key = new CipherKey(new Uint8Array([]));
```
bennyn, line 30: Can `Uint8Array` become a mandatory property of the `constructor`?

packages/proteus/spec/derived/CipherKeySpec.ts
```diff
@@ -96,8 +96,8 @@ describe('CipherKey sanity checks (IETF ChaCha20 test vectors)', () => {
       };
     };
 
-    ietf_vectors.forEach(vector => {
-      const result = decrypt_cipher_text(vector);
+    ietfVctors.forEach(vector => {
```
bennyn, line 99:
```suggestion
    ietfVectors.forEach(vector => {
```

packages/proteus/spec/derived/CipherKeySpec.ts
```diff
@@ -31,7 +31,7 @@ beforeAll(async () => {
 });
 
 describe('CipherKey sanity checks (IETF ChaCha20 test vectors)', () => {
-  const ietf_vectors: Vector[] = [
+  const ietfVctors: Vector[] = [
```
bennyn, line 34:
```suggestion
  const ietfVectors: Vector[] = [
```

packages/proteus/src/main/derived/DerivedSecrets.ts
ffflorian: Yes :)

PR #2952, opened 2020-04-15
Revert "feat(core): Make apiClient public"

lipis:
Reverts wireapp/wire-web-packages#2951


bennyn:
Replaced with https://github.com/wireapp/wire-web-packages/pull/2953

PR #2955, opened 2020-04-15
feat(api-client): Provide cancellation token for asset upload

bennyn:


packages/api-client/src/asset/AssetAPI.ts
```diff
@@ -17,11 +17,12 @@
  *
  */
 
-import {AxiosRequestConfig} from 'axios';
+import Axios, {AxiosRequestConfig} from 'axios';
```
ffflorian, line 20:
```suggestion
import axios, {AxiosRequestConfig} from 'axios';
```

packages/api-client/src/asset/AssetAPI.ts
```diff
@@ -83,14 +84,33 @@ export class AssetAPI {
 
     const footer = `\r\n--${BOUNDARY}--\r\n`;
 
-    const response = await this.client.sendRequest<AssetUploadData>({
+    const cancelSource = Axios.CancelToken.source();
```
ffflorian, line 87:
```suggestion
    const cancelSource = axios.CancelToken.source();
```

PR #2962, opened 2020-04-17
feat: remove old demo and deploy new demo to netlify

bardiawire:


lipis:
Can we make the circles smaller so they will fit 12 per line?!
Screenshot 2020-04-17 at 5 11 02 PM

PR #2985, opened 2020-04-28
fix(eslint-config): Add some comments back

lipis:


packages/api-client/src/http/BackendErrorLabel.ts
```diff
@@ -56,7 +56,9 @@ export enum BackendErrorLabel {
   NO_CONVERSATION = 'no-conversation',
   NO_CONVERSATION_CODE = 'no-conversation-code',
   NO_MANAGED_CONVERSATION = 'no-managed-team-conv',
+  // @deprecated
```
ffflorian, line 59:
```suggestion
  /** @deprecated */
```

packages/api-client/src/http/BackendErrorLabel.ts
```diff
@@ -56,7 +56,9 @@ export enum BackendErrorLabel {
...
   NO_OTHER_OWNER = 'no-other-owner',
+  // This error is thrown when an owner tries to delete hisself
```
ffflorian, line 61:
```suggestion
  // This error is thrown when an owner tries to delete themself
```

PR #3013, opened 2020-05-07
feat(api-client): Logout on access token refresh error

bennyn:


packages/api-client/src/http/HttpClient.ts
```diff
@@ -149,12 +151,22 @@ export class HttpClient extends EventEmitter {
...
-      ObfuscationUtil.obfuscateAccessToken(accessToken),
-    );
-    return this.accessTokenStore.updateToken(accessToken);
+    try {
+      const accessToken = await this.postAccess(expiredAccessToken);
+      this.logger.info(
+        `Received updated access token. It will expire in "${accessToken.expires_in}" seconds.`,
+        ObfuscationUtil.obfuscateAccessToken(accessToken),
+      );
+      return this.accessTokenStore.updateToken(accessToken);
```
ffflorian, line 160: Do we need `return await` here?

bennyn: no

PR #3028, opened 2020-05-11
feat(api-client): Get client mismatch info

bennyn:


packages/api-client/demo.ts
```diff
@@ -53,11 +53,11 @@ async function createContext(storeEngine: CRUDEngine, apiClient: APIClient, logi
     const {expiration, zuid} = await storeEngine.read(AUTH_TABLE_NAME, AUTH_COOKIE_KEY);
     const cookie = new Cookie(zuid, expiration);
     logger.log(`Found cookie "${zuid}".`);
-    logger.log('Logging in with EXISTING cookie.');
+    logger.log('Logging in with EXISTING cookie...');
     return apiClient.init(loginData.clientType, cookie);
   } catch (error) {
-    logger.log(`Failed to find existing cookie.`, error);
-    logger.log(`Logging in with NEW cookie: ${error.message}`);
+    logger.log(`Failed to find existing cookie: ${error.message}`, error);
```
ffflorian, line 59: Why log the error twice? This will generate a log like

`Failed to find existing cookie: <error message> Error: <error message>`.

packages/api-client/src/conversation/ConversationAPI.ts
```diff
@@ -473,11 +473,15 @@ export class ConversationAPI {
...
-        : await this.client.sendProtocolBuffer<ClientMismatch>(config, true);
-    return response.data;
+    try {
+      const response =
+        typeof messageData.recipients === 'object'
+          ? await this.client.sendJSON<ClientMismatch>(config, true)
+          : await this.client.sendProtocolBuffer<ClientMismatch>(config, true);
+      return response.data;
+    } catch (error) {
+      throw error;
```
ffflorian, line 483: What's the advantage of throwing the error after catching it?

bennyn: That the outside caller get's to see this error. When not doing it I was not able to try-catch it from the outside.

packages/api-client/src/conversation/ConversationAPI.ts
```diff
@@ -480,6 +480,15 @@ export class ConversationAPI {
     return response.data;
   }
 
+  public async postForClients(clientId: string, conversationId: string): Promise<ClientMismatch> {
+    try {
+      await this.postOTRMessage(clientId, conversationId);
+      throw Error(`Expected backend to throw error.`);
```
ffflorian, line 486:
```suggestion
```

bennyn: It should. The call should never pass.

ffflorian: Alright! But let's add the `new` keyword then :)
```suggestion
      throw new Error(`Expected backend to throw error.`);
```

PR #3038, opened 2020-05-18
feat!(api-client): Add asset v1 & v2 download, add request cancellation & progressCallback

Yserz:


packages/api-client/src/asset/AssetUtil.ts
```diff
@@ -19,4 +19,6 @@
 
 export const isValidAssetId = (key: string): boolean => /^[A-Za-z0-9-]+$/.test(key);
 
+export const isValidUUID = (id: string): boolean => /^[A-Za-z0-9-]+$/.test(id);
```
ffflorian, line 22: That's the same as `isValidAssetId`. Maybe we can remove `isValidAssetId` then?

PR #3050, opened 2020-05-26
feat: Add postVerifyPhoneCode to AccountAPI

bardiarastin98:


packages/api-client/src/account/AccountAPI.ts
```diff
@@ -130,6 +130,34 @@ export class AccountAPI {
     await this.client.sendJSON(config);
   }
 
+  /**
+   * Verify phone sms code
```
ffflorian, line 134:
```suggestion
   * Activate (i.e. confirm) a phone number
```

packages/api-client/src/account/AccountAPI.ts
```diff
@@ -130,6 +130,34 @@ export class AccountAPI {
...
+    };
+    try {
+      const response = await this.client.sendJSON(config);
+      if (response.status === 200) {
+        return true;
+      }
+      if (response.status === 404) {
+        return false;
+      }
+      return false;
```
ffflorian, line 155:
```suggestion
      return response.status === 200;
```

packages/api-client/src/account/AccountAPI.ts
```diff
@@ -130,6 +130,34 @@ export class AccountAPI {
...
+    } catch (error) {
+      throw error;
```
ffflorian, line 157: try/catch is not needed if you are not doing anything with it.

bardiarastin98: done in https://github.com/wireapp/wire-web-packages/pull/3050/commits/ea1aa0c2f970faf53850df0392e6a1ea566e6b42

PR #3061, opened 2020-06-05
feat(bot-api): Remember cookie

bardiarastin98:


packages/bot-api/src/Bot.ts
```diff
@@ -113,14 +114,35 @@ export class Bot {
...
 
     return apiClient;
   }
 
+  async getCookie(storeEngine: CRUDEngine): Promise<Cookie | undefined> {
+    try {
+      const {expiration, zuid} = await storeEngine.read(AUTH_TABLE_NAME, AUTH_COOKIE_KEY);
+      const cookie = new Cookie(zuid, expiration);
+      return cookie;
+    } catch (error) {
```
bennyn, line 140: Don't catch the error here. The surrounding code will catch it.

PR #3092, opened 2020-06-30
fix(api-client): Chunk `postMultiPreKeyBundles` requests to prevent `413 Request entity too large`

Yserz:


packages/api-client/src/user/UserAPI.ts
```diff
@@ -434,13 +435,7 @@ export class UserAPI {
...
   }
 
-  /**
-   * Given a map of user IDs to client IDs return a prekey for each one.
-   * Note: The maximum map size is 128 entries.
-   * @param userClientMap A map of the user's clients
-   * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/users/getMultiPrekeyBundles
-   */
-  public async postMultiPreKeyBundles(userClientMap: UserClients): Promise<UserPreKeyBundleMap> {
+  private async _postMultiPreKeyBundles(userClientMap: UserClients): Promise<UserPreKeyBundleMap> {
```
ffflorian, line 438: Can we rename this to something like `postMultiPreKeyBundleChunk`?

PR #3111, opened 2020-07-13
feat(react-ui-kit): Labeled Select component

Yserz:


packages/react-ui-kit/src/Form/Input.tsx
```diff
@@ -95,7 +95,7 @@ export const Input: React.FC<InputProps<HTMLInputElement>> = React.forwardRef<
     />
   );
   return props.label ? (
-    <Label bold={false} markInvalid={props.markInvalid}>
+    <Label bold={false} markInvalid={props.markInvalid} style={{width: '100%'}}>
```
thisisamir98, line 98: Are you sure that this will not break other places where this component is already being used?

Yserz: Should be fine in most cases™️

PR #3157, opened 2020-08-06
feat(commons): Add handle validation

bennyn:


packages/commons/src/main/util/ValidationUtil.ts
```diff
@@ -34,6 +34,10 @@ export function isValidEmail(email: string): boolean {
   return regExp.test(email);
 }
 
+export function isValidHandle(handle: string): boolean {
+  return /^@?[a-z_0-9.-]{2,256}$/.test(handle);
```
ffflorian, line 38: This will validate handles like `bennyn` and `@bennyn`. I think we should stick to one.

bennyn: I copied this one over from webapp.

ffflorian: Ah, we then cut the leading `@`. Okay.

bennyn:
Obsolete with https://github.com/wireapp/wire-web-packages/pull/3160

PR #3162, opened 2020-08-07
feat(bot-handler-wizard): Add a conversational wizard

bennyn:


packages/bot-handler-wizard/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "dependencies": {
+    "wizardy": "0.2.6"
+  },
+  "devDependencies": {
+    "@types/jasmine": "3.5.10",
+    "@types/node": "~12",
+    "@wireapp/bot-api": "7.10.27",
```
ffflorian, line 8: I think this should be in the dependencies.

packages/bot-handler-wizard/src/WizardHandler.ts
```diff
@@ -0,0 +1,89 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+import {MessageHandler} from '@wireapp/bot-api';
```
ffflorian, line 19:
```suggestion

import {MessageHandler} from '@wireapp/bot-api';
```

packages/bot-handler-wizard/package.json
bennyn: No. Bot API is an optional dependency. It's just listed in dev dependencies so that tests can work with it. It's the pattern we use for all other bot handlers.

PR #3171, opened 2020-08-11
feat(api-client): Remove last bits of contact import

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/user/data/UserPropertiesSetData.ts
+++ b/packages/api-client/src/user/data/UserPropertiesSetData.ts
@@ -32,10 +32,6 @@ export enum NotificationPreference {
 
 export interface WebappProperties {
   [property_key: string]: any;
-  contact_import: {
-    google?: number;
-    macos?: number;
-  };
   enable_debugging: boolean;
   settings: {
     call: {


packages/api-client/src/user/data/UserPropertiesSetData.ts
```diff
@@ -32,10 +32,6 @@ export enum NotificationPreference {
 
 export interface WebappProperties {
   [property_key: string]: any;
-  contact_import: {
-    google?: number;
-    macos?: number;
-  };
   enable_debugging: boolean;
```
ffflorian, line 35: I am wondering if these properties should be optional?

bennyn: Let's remove them.

ffflorian: I meant the other properties like `enable_debugging`.

bennyn: Does it cause any problems having them to be mandatory?

ffflorian: For now it doesn't :)

PR #3175, opened 2020-08-11
feat(api-client,bot-api,core): Send messages to specific clients only

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/auth/ClientPreKey.ts
+++ b/packages/api-client/src/auth/ClientPreKey.ts
@@ -20,6 +20,7 @@
 import {PreKey} from '../auth/';
 
 export interface ClientPreKey {
+  /** The client ID */
   client: string;
   prekey: PreKey;
 }
--- a/packages/api-client/src/auth/PreKey.ts
+++ b/packages/api-client/src/auth/PreKey.ts
@@ -18,6 +18,8 @@
  */
 
 export interface PreKey {
+  /** The PreKey ID */
   id: number;
+  /** The PreKey data */
   key: string;
 }
--- a/packages/api-client/src/auth/PreKeyBundle.ts
+++ b/packages/api-client/src/auth/PreKeyBundle.ts
@@ -20,6 +20,8 @@
 import {ClientPreKey} from '../auth/';
 
 export interface PreKeyBundle {
+  /** PreKeys of all clients */
   clients: ClientPreKey[];
+  /** The user ID */
   user: string;
 }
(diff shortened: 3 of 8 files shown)


packages/api-client/src/auth/ClientPreKey.ts
```diff
@@ -20,6 +20,7 @@
 import {PreKey} from '../auth/';
 
 export interface ClientPreKey {
+  /** The client ID */
```
lipis, line 23: Do we really need these comments? Wouldn't it be better to change the variable names to be more descriptive?

- `client` -> `clienId`

packages/api-client/src/auth/PreKey.ts
```diff
@@ -18,6 +18,8 @@
  */
 
 export interface PreKey {
+  /** The PreKey ID */
   id: number;
+  /** The PreKey data */
   key: string;
```
lipis, line 24: maybe `data` instead of key since it's data?

packages/api-client/src/auth/PreKeyBundle.ts
```diff
@@ -20,6 +20,8 @@
 import {ClientPreKey} from '../auth/';
 
 export interface PreKeyBundle {
+  /** PreKeys of all clients */
   clients: ClientPreKey[];
+  /** The user ID */
   user: string;
```
lipis, line 26: `userId`?

packages/api-client/src/auth/ClientPreKey.ts
ffflorian: These are not variables but return values from the server. We can't just rename them.

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -118,18 +118,36 @@ export class ConversationService {
...
+
+    let preKeys = await Promise.all(members.map(member => this.apiClient.user.api.getUserPreKeys(member)));
+
+    if (userIds && !Array.isArray(userIds)) {
+      preKeys = preKeys.map(preKey => {
+        if (preKey.user in userIds) {
+          preKey.clients = preKey.clients.filter(client => userIds[preKey.user].includes(client.client));
+        }
+        return preKey;
+      });
```
ffflorian, line 150: The actual implementation

packages/bot-api/src/MessageHandler.ts
```diff
@@ -96,31 +96,53 @@ export abstract class MessageHandler {
...
   }
 
-  async sendConfirmation(conversationId: string, firstMessageId: string): Promise<void> {
+  /**
+   * @param userIds Only send message to specified user IDs or to certain clients of specified user IDs
+   */
+  async sendConfirmation(
+    conversationId: string,
+    firstMessageId: string,
+    userIds?: string[] | Record<string, string[]>,
```
bennyn, line 137: Do we have a type for thus user / client map of `Record<string, string[]` to make clear which string is a user ID and which string is a client ID?

ffflorian: Not yet but let me create one!

PR #3182, opened 2020-08-13
fix(api-client): Skip logout REST request when credentials are invalid (WEBAPP-7041)

bennyn:


packages/api-client/src/APIClient.ts
```diff
@@ -263,20 +265,17 @@ export class APIClient extends EventEmitter {
     return this.init(clientType, CookieStore.getCookie());
   }
 
-  public async logout(options = {ignoreError: false}): Promise<void> {
+  public async logout(options = {skipLogoutRequest: false}): Promise<void> {
```
ffflorian, line 268:
```suggestion
  public async logout(options: {skipLogoutRequest: boolean} = {skipLogoutRequest: false}): Promise<void> {
```

PR #3187, opened 2020-08-18
feat(api-client): Add Services API

ffflorian:
## Pull Request Checklist


Yserz:
Hm, this looks pretty similar to other already existing endpoints with the prefix of the BotURL. Would it make sense to alter the other endpoints to have this prefix instead to duplicate the endpoints? WDYT?

ffflorian:
> Hm, this looks pretty similar to other already existing endpoints with the prefix of the BotURL. Would it make sense to alter the other endpoints to have this prefix instead to duplicate the endpoints? WDYT?

I think it's too confusing to use these in combination with other APIs. Let's keep them separated.

Yserz:
Mh, then consequently there should be a BotAssetAPI, no?

ffflorian:
I don't see the necessity for a BotAssetAPI since there is a lot of logic in the AssetAPI which we would just call from the Bot Asset API. But I see the necessity for a BotAPI because the descriptions are important IMO.

packages/api-client/src/asset/AssetAPI.ts
```diff
@@ -154,6 +154,7 @@ export class AssetAPI {
     token?: string | null,
     forceCaching: boolean = false,
     progressCallback?: ProgressCallback,
+    customUrl?: string,
```
Yserz, line 157: It feels `customUrl` might be a bit too broad especially since the function is running validation against the parameter `assetId` but might use a completely different one inside the `customUrl`. Also it's not guaranteed that the passed `customUrl` is a v3 URL. Maybe something like `urlPrefix` is more appropriate.

Yserz: Maybe even `isBotAsset`?

packages/api-client/src/bots/BotsAPI.ts
```diff
@@ -0,0 +1,215 @@
...
+   * Download an asset.
+   */
+  public getAsset(
+    assetId: string,
+    token?: string | null,
+    forceCaching: boolean = false,
+    progressCallback?: ProgressCallback,
+  ): Promise<RequestCancelable<AssetResponse>> {
+    const customUrl = `${BotsAPI.URL.BOT}/${BotsAPI.URL.ASSETS}/${assetId}`;
+    return new AssetAPI(this.client).getAssetV3(assetId, token, forceCaching, progressCallback, customUrl);
```
Yserz, line 188: I think it would be better to pass an AssetAPI instance in the constructor instead of creating new ones here

Yserz (approved):

PR #3210, opened 2020-09-01
refactor(commons): Improve OS family check

Yserz:


ffflorian (approved):

packages/commons/src/main/util/Runtime.ts
```diff
@@ -43,17 +43,28 @@ export class Runtime {
   }
 
   public static getOSFamily(): OperatingSystem {
+    const OS = {
+      DESKTOP: {
+        LINUX: ['linux', 'ubuntu', 'fedora', 'gentoo', 'debian', 'suse', 'centos', 'red hat', 'freebsd', 'openbsd'],
+        MAC: ['os x', 'mac os'],
+        WINDOWS: ['windows', 'windows server 2008 r2 / 7', 'windows server 2008 / vista', 'windows xp'],
```
AndyLnd, line 50: Isn't it a bit redundant to include all those names, as all of those contain the word `windows`?

packages/commons/src/main/util/Runtime.ts
```diff
@@ -43,17 +43,28 @@ export class Runtime {
   }
 
   public static getOSFamily(): OperatingSystem {
+    const OS = {
+      DESKTOP: {
+        LINUX: ['linux', 'ubuntu', 'fedora', 'gentoo', 'debian', 'suse', 'centos', 'red hat', 'freebsd', 'openbsd'],
```
AndyLnd, line 48: Where is `LINUX` used?

AndyLnd (changes requested):

PR #3279, opened 2020-10-19
fix(api-client): APIClient connect onConnect param type

Yserz:


packages/api-client/src/tcp/WebSocketClient.ts
```diff
@@ -50,6 +50,8 @@ export class AbortHandler {
   isAborted = () => this.aborted;
 }
 
+export type OnConnect = (abortHandler: AbortHandler) => Promise<void>;
```
AndyLnd, line 53: It might be better to have that as a n optional parameter.

AndyLnd (changes requested):

AndyLnd (approved):

PR #3349, opened 2020-11-27
refactor(react-ui-kit): Emotion v11 migration

Yserz:
Emotion provides a good amount of compatibility from v10 to v11 but there are some areas that need to be migrated.
https://emotion.sh/docs/emotion-11

The main change from our perspective is the renaming of the `@emotion/core` package to `@emotion/react`.


packages/react-ui-kit/src/Layout/Theme.tsx
```diff
@@ -18,29 +18,32 @@
...
-import {jsx} from '@emotion/core';
+import {jsx, Theme, ThemeProvider as EmotionThemeProvider} from '@emotion/react';
+
 import React from 'react';
-import {ThemeProvider as EmotionThemeProvider} from 'emotion-theming';
 import {COLOR} from '../Identity/colors';
 import {filterProps} from '../util';
 
+// Apply custom theme to Emotion so TypeScript can pick it up in the css prop
+declare module '@emotion/react' {
```
Yserz, line 28: Theme needs to be augmented in order to have correct typing for the css prop theme parameter.

packages/react-ui-kit/webpack.config.js
```diff
@@ -36,6 +36,9 @@ module.exports = {
   resolve: {
     alias: {
       '@wireapp/react-ui-kit': path.resolve(__dirname, 'src'),
+      // Needed until libs have support for emotion v11
+      // https://github.com/infinum/emotion-normalize/pull/12
+      '@emotion/core': '@emotion/react',
```
Yserz, line 41: We need to redirect calls from `@emotion/core` to `@emotion/react` until `emotion-normalize` supports it.

Yserz: https://github.com/infinum/emotion-normalize/pull/12

packages/react-ui-kit/src/Layout/Theme.tsx
Yserz: https://github.com/emotion-js/emotion/issues/1197

PR #3350, opened 2020-11-27
fix(proteus): Convert SessionTag to string only once

ffflorian:
Since `SessionTag.toString()` uses `sodium.to_hex()` and is therefore quite costly this will save a few CPU cycles.

## Pull Request Checklist

### DIFF
--- a/packages/proteus/src/main/session/Session.ts
+++ b/packages/proteus/src/main/session/Session.ts
@@ -194,7 +194,7 @@ export class Session {
 
   private _evict_oldest_session_state(): void {
     const oldest = Object.keys(this.session_states)
-      .filter(obj => obj.toString() !== this.session_tag_name)
+      .filter(sessionTagName => sessionTagName.toString() !== this.session_tag_name)
       .reduce((lowest, obj) => {
         return this.session_states[obj].idx < this.session_states[lowest].idx ? obj.toString() : lowest;
       });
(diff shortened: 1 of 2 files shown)


packages/proteus/src/main/session/Session.ts
```diff
@@ -329,9 +334,10 @@ export class Session {
     }
 
     encoder.u8(5);
-    encoder.object(Object.keys(this.session_states).length);
+    const sessionStatesIds = Object.keys(this.session_states);
+    encoder.object(sessionStatesIds.length);
 
-    for (const index in this.session_states) {
+    for (const index of sessionStatesIds) {
```
bennycode, line 340:
```suggestion
    for (const id of sessionStatesIds) {
```

That's not an index anymore but an ID.

packages/proteus/src/main/session/Session.ts
```diff
@@ -189,7 +194,7 @@ export class Session {
 
   private _evict_oldest_session_state(): void {
     const oldest = Object.keys(this.session_states)
-      .filter(obj => obj.toString() !== this.session_tag.toString())
+      .filter(obj => obj.toString() !== this.session_tag_name)
```
bennycode, line 197:
```suggestion
      .filter(sessionTagName => sessionTagName.toString() !== this.session_tag_name)
```

bennycode (changes requested):

packages/proteus/src/main/session/Session.ts
ffflorian: My naming was wrong, it should've been `sessionStatesIndices`.

bennycode (approved):

PR #3358, opened 2020-12-08
refactor(commons): Migrate from Jasmine/Karma to Jest

Yserz:


packages/commons/src/main/util/ArrayUtil.test.node.js
```diff
@@ -21,6 +21,9 @@
 
 const {ArrayUtil} = require('@wireapp/commons');
 
+/**
+ * @testEnvironment node
```
Yserz, line 25: Even if `jsdom` is specified as `testEnvironment` in `jest.config.js` it can be overridden with this comment.

Yserz: https://jestjs.io/docs/en/configuration#testenvironment-string

lipis (approved):

PR #3393, opened 2020-12-17
chore(proteus): Clean js files from src

ffflorian:
Proteus' build files are created in `src` but the `clean` script was removing files from the non-existent `dist` dir.


thisisamir98 (approved):

thisisamir98:
LGTM

PR #3512, opened 2021-02-05
fix(api-client): Adjust team search order param

Yserz:


packages/api-client/src/team/search/TeamSearchOptions.ts
```diff
@@ -29,7 +29,7 @@ export interface TeamSearchOptions {
   /** Filter results by member role */
   frole?: Role[];
   /** Sort order (asc | desc | undefined) */
-  order?: SearchOrder;
+  sortorder?: SearchOrder;
   /** Max number of search results. Defaults to 15 results. Min 1, max 500. */
```
lipis, line 33: I guess these kind of comments instead of `//` is for the IDEs to show descriptions?

lipis (approved):

packages/api-client/src/team/search/TeamSearchOptions.ts
Yserz: yes

PR #3842, opened 2021-05-12
feat(core): Add cipher options for encrypting assets and use named parameters (SQCORE-644)

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/core/src/main/conversation/message/MessageBuilder.ts
+++ b/packages/core/src/main/conversation/message/MessageBuilder.ts
@@ -106,8 +106,9 @@ export class MessageBuilder {
     originalMessageId: string,
     expectsReadConfirmation?: boolean,
     legalHoldStatus?: LegalHoldStatus,
+    __debugOptions?: {customHash?: Buffer; customCipher?: string},
   ): Promise<FileAssetMessage> {
-    const imageAsset = await this.assetService.uploadFileAsset(file);
+    const imageAsset = await this.assetService.uploadFileAsset(file, undefined, __debugOptions);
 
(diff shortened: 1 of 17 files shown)


packages/core/src/main/cryptography/AssetCryptography.node.ts
```diff
@@ -52,11 +53,14 @@ export const decryptAsset = async ({
   return Buffer.concat([decipherUpdated, decipherFinal]);
 };
 
-export const encryptAsset = async (plainText: Uint8Array): Promise<EncryptedAsset> => {
+export const encryptAsset = async (
+  plainText: Buffer | Uint8Array,
+  debugOptions?: DebugOptions,
```
bennycode, line 58: Let's also use a generic `options` param here which holds a `debug` property (the same like you did with `AssetOptions`).

packages/core/src/main/conversation/message/MessageBuilder.ts
```diff
@@ -184,8 +186,9 @@ export class MessageBuilder {
     messageId = MessageBuilder.createId(),
     expectsReadConfirmation?: boolean,
     legalHoldStatus?: LegalHoldStatus,
+    debugOptions?: DebugOptions,
```
bennycode, line 189: Now we have many optionals: `expectsReadConfirmation`, `legalHoldStatus` & `debugOptions`

Let's put them all together in one `options` object.

bennycode (changes requested):

bennycode (approved):

PR #3933, opened 2021-06-07
feat(api-client): Add qualified members to a conversation

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/core/src/main/conversation/ConversationService.ts
+++ b/packages/core/src/main/conversation/ConversationService.ts
@@ -60,6 +60,7 @@ import {
 import type {AssetContent, ClearedContent, DeletedContent, HiddenContent, RemoteData} from '../conversation/content/';
 import type {CryptographyService, EncryptedAsset} from '../cryptography/';
 import * as AssetCryptography from '../cryptography/AssetCryptography.node';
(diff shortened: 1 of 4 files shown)


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -962,11 +962,17 @@ export class ConversationService {
...
-  public async addUser(conversationId: string, userIds: string[]): Promise<string[]>;
-  public async addUser(conversationId: string, userIds: string | string[]): Promise<string | string[]> {
-    const ids = typeof userIds === 'string' ? [userIds] : userIds;
-    await this.apiClient.conversation.api.postMembers(conversationId, ids);
+  public async addUser<T extends string | QualifiedId | string[] | QualifiedId[]>(
+    conversationId: string,
+    userIds: T,
+  ): Promise<T> {
+    const ids = Array.isArray(userIds) ? userIds : [userIds];
+    if (typeof ids[0] === 'string') {
```
bennycode, line 970: Let's use a type predicate so we can avoid the `as` casting later on.

bennycode (changes requested):

PR #3979, opened 2021-06-23
feat(api-client,core): Send OTR messages to federated users (SQCORE-789)

ffflorian:
## Pull Request Checklist


packages/bot-api/src/MessageHandler.ts
```diff
@@ -124,7 +124,7 @@ export abstract class MessageHandler {
   async sendCall(conversationId: string, content: CallingContent, userIds?: string[] | UserClientsMap): Promise<void> {
     if (this.account?.service) {
       const callPayload = this.account.service.conversation.messageBuilder.createCall({conversationId, content});
-      await this.account.service.conversation.send(callPayload, userIds);
+      await this.account.service.conversation.send({payloadBundle: callPayload, userIds});
```
bennycode, line 127: What makes the payload a bundle? Is it bundled with other data?

packages/core/src/main/conversation/message/MessageService.ts
```diff
@@ -76,6 +83,130 @@ export class MessageService {
...
+              }
+            }
+          }
+        }
+      }
+    } else if (!!messageData.ignoreAll) {
+      // report nothing
+      return null;
+    } else if (!!messageData.reportAll) {
+      // report everything
```
bennycode, line 136: This case can be deleted if it has no business logic.

bennycode (changes requested):

packages/bot-api/src/MessageHandler.ts
ffflorian: Good question, it was already named like that.

PR #4050, opened 2021-07-13
feat(api-client): Get remote conversations

bennycode:
Can be tested with:

```typescript
const apiClient = await initClient();
const conversations = await apiClient.conversation.api.getRemoteConversations('staging.zinfra.io');
```

### DIFF
--- a/packages/api-client/src/conversation/Conversation.ts
+++ b/packages/api-client/src/conversation/Conversation.ts
@@ -18,6 +18,7 @@
  */
 
 import type {ConversationMembers} from './';
+import {QualifiedId} from '../user';
 
 export enum CONVERSATION_TYPE {
   REGULAR = 0,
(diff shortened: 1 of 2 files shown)


packages/api-client/src/conversation/Conversation.ts
```diff
@@ -48,6 +49,7 @@ export interface Conversation {
   members: ConversationMembers;
   message_timer: number | null;
   name: string;
+  qualified_id: QualifiedId;
```
ffflorian, line 52: Should be optional, shouldn't it?

packages/api-client/src/conversation/ConversationAPI.ts
```diff
@@ -235,6 +236,21 @@ export class ConversationAPI {
     return this._getConversations(startConversationId, undefined, limit);
   }
 
+  /**
+   * Get all remote conversations from a federated backend.
+   * @see https://staging-nginz-https.zinfra.io/api/swagger-ui/#/default/post_list_conversations
+   */
+  public async getRemoteConversations(ownDomain: string): Promise<Conversation[]> {
+    const config: AxiosRequestConfig = {
+      data: {},
```
ffflorian, line 245: Not needed if empty

PR #4111, opened 2021-08-06
feat(api-client): Restructured BillingAPI

Yserz:


packages/api-client/src/team/payment/PaymentAPI.ts
```diff
@@ -118,6 +139,9 @@ export class PaymentAPI {
     return response.data;
   }
 
+  /**
+   * @deprecated
```
ffflorian, line 143: Use BillingAPI? :slightly_smiling_face:

packages/api-client/src/team/billing/SupportedCurrency.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export enum SupportedCurrency {
+  CHF = 'CHF',
+  EUR = 'EUR',
+  USD = 'USD',
```
ffflorian, line 23: We could think about using `'CHF' | 'EUR' | 'USD'` as type instead of an enum so that we don't have to import `SupportedCurrency` every time we want to use `PlanData`

ffflorian (approved):

packages/api-client/src/team/payment/PaymentAPI.ts
Yserz: There's no replacement for that endpoint because it was never used.

packages/api-client/src/team/billing/SupportedCurrency.ts
Yserz: When using `getAvailablePlans` it can be filtered by currency. Thus we need it standalone :)

ffflorian: You can still use the filter. The proposed type would look like this:

```ts
export type SupportedCurrency = 'CHF' | 'EUR' | 'USD';
```

PR #4122, opened 2021-08-30
refactor: Upgrade TypeScript to 4.4

ffflorian:
## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/conversation/ConversationAPI.ts
+++ b/packages/api-client/src/conversation/ConversationAPI.ts
@@ -46,7 +46,7 @@ import type {
   ConversationReceiptModeUpdateEvent,
   ConversationRenameEvent,
 } from '../event/';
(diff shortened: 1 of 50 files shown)


packages/api-client/src/conversation/ConversationAPI.ts
```diff
@@ -522,9 +522,9 @@ export class ConversationAPI {
       const response = await this.client.sendJSON<ConversationJoinData>(config);
       return response.data;
     } catch (error) {
-      switch (error.label) {
+      switch ((error as BackendError).label) {
```
bennycode, line 525: Instead of casting it 3x I would do:

```ts
const backendError = error as BackendError;
switch (backendError) { ... }
```

bennycode (changes requested):

bennycode (approved):

PR #4128, opened 2021-09-15
feat(api-client): Remove obsolete otr_muted field

AndyLnd:


ffflorian (commented):
This should be a feature I think

PR #4129, opened 2021-09-15
fix: Wrap Uint8Array before sending protobuf data

atomrc:
:warning: This issue needs more investigation!!!

There is a weird issue with protobuf encoding: some data seem to be shared among `Uint8Array` instances. 
To be precise, the `buffer` property can sometimes have data that is not related to the current buffer being encoded. 

Let's say the Uint8Array you generate with protobuf looks something like this `[1, 10, 4]` 
This same array can have a buffer property that has the following `[3, 4, 2, 4, 5, 10, 8, 1, 10, 4]`. 

This PR addresses the issue, but I don't believe it is a satisfying solution... (And it probably doesn't belong in such low level packages). 

This is merely a place for discussing where the solution should live

[EDIT 16/09]
Ok, so this is a normal behavior of `protobuf`. When allocating `Uint8Array` they also allocate an 8k fixed buffer along with it. 
Then they set a `byteOffset` and `byteLength` that tell where the data is in the buffer. 
Now the problem is probably with Axios ignoring those offset and length...
(see https://github.com/protobufjs/protobuf.js/issues/852)

[EDIT2 16/09]
I confirmed the issue comes from Axios and posted an issue on their end https://github.com/axios/axios/issues/4068

## Pull Request Checklist


atomrc:
Superseeded by #4130

PR #4130, opened 2021-09-16
fix(api-client): Force protobuf arrays to be recreated before sent

atomrc:
To optimize performance when protobuf encodes an entity, it first generates a big ArrayBuffer which it fills with the content of the entity and then generates a DataView (namely `Uint8Array`) on top of it that has `byteOffset` and `byteLength` properties that tell how the buffer should be read.

Axios currently ignores those properties and send the entire 8k buffer through the wire. 

This fix slices the `Uint8Array` in order to create a new array that has exactly the right amount of bytes in it. Axios still uses the complete buffer but this time it only has the right bytes, no more, no less.

## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/conversation/ConversationAPI.ts
+++ b/packages/api-client/src/conversation/ConversationAPI.ts
@@ -654,7 +654,11 @@ export class ConversationAPI {
     messageData: ProtobufOTR.QualifiedNewOtrMessage,
   ): Promise<MessageSendingStatus> {
     const config: AxiosRequestConfig = {
-      data: ProtobufOTR.QualifiedNewOtrMessage.encode(messageData).finish(),
+      /*
+       * We need to slice the content of what protobuf has generated in order for Axios to send the correct data (see https://github.com/axios/axios/issues/4068)
+       * FIXME: The `slice` can be removed as soon as Axios publishes a version with the dataview issue fixed.
+       */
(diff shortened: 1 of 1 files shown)


packages/api-client/src/conversation/ConversationAPI.ts
```diff
@@ -654,7 +654,11 @@ export class ConversationAPI {
     messageData: ProtobufOTR.QualifiedNewOtrMessage,
   ): Promise<MessageSendingStatus> {
     const config: AxiosRequestConfig = {
-      data: ProtobufOTR.QualifiedNewOtrMessage.encode(messageData).finish(),
+      /*
+       * We need to slice the content of what protobuf has generated in order for Axios to send the correct data (see https://github.com/axios/axios/issues/4068)
+       * FIXME: The `slice` can be removed as soon as Axios publishes a version with the dataview issue fixed.
```
ffflorian, line 659: Let's add a link to the issue here

atomrc: It is already there

ffflorian: Whoops, true! :eyeglasses:

ffflorian (approved):

ffflorian:
Related: https://github.com/axios/axios/issues/4068

PR #4169, opened 2021-11-03
fix(core): Cleanly delete users from payload when mismatch happens

atomrc:
## Pull Request Checklist

### DIFF
--- a/packages/core/src/main/conversation/message/MessageService.ts
+++ b/packages/core/src/main/conversation/message/MessageService.ts
@@ -410,15 +410,16 @@ export class MessageService {
       // walk through deleted user ids
       for (const [deletedUserId] of Object.entries(deletedUserIdClients)) {
         // walk through message recipients
-        for (const recipientIndex in message.recipients) {
+        for (const recipients of message.recipients) {
           // check if message recipients' domain is the same as the deleted user's domain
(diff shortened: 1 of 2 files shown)


packages/core/src/main/conversation/message/MessageService.ts
```diff
@@ -410,15 +410,16 @@ export class MessageService {
...
+          if (recipients.domain === deletedUserDomain) {
             // check if message recipients' id is the same as the deleted user's id
-            for (const entriesIndex in message.recipients[recipientIndex].entries || []) {
-              const uuid = message.recipients[recipientIndex].entries![entriesIndex].user?.uuid;
+            for (const recipientEntry of recipients.entries || []) {
+              const uuid = recipientEntry.user?.uuid;
               if (!!uuid && bytesToUUID(uuid) === deletedUserId) {
                 // delete this user from the message recipients
-                delete message.recipients[recipientIndex].entries![entriesIndex];
+                const deleteIndex = recipients.entries!.indexOf(recipientEntry);
```
atomrc, line 421: `delete` operator on an array, is a bad idea!!! (will not change the size of the array, thus it will have undefined indexes)

AndyLnd (approved):

PR #4171, opened 2021-11-04
refactor(core): Simplify and streamline responsibilities of the MessageService

atomrc:
BREAKING CHANGE
- When calling `account.services.converation.send` with a list of `UserClients`, the core will not try to fetch the latest clients from the backend and will instead only send to the given clients. This might generate a mismatch error is there are missing clients given. 

This PR will:
- Simplify the `MessageService` API to only 2 public methods (`sendMessage`, `sendFederatedMessage`);
- Move all the encryption jobs to the `MessageService`, the `ConversationService` is not even aware encryption exists;
- Harmonize how mismatch are handled and retried in the `MessageService`;
- Cover the main sending scenarios (as external, with mismatch, as protobuf...) with tests;

## Pull Request Checklist

### DIFF
--- a/packages/core/src/main/conversation/ConversationService.ts
+++ b/packages/core/src/main/conversation/ConversationService.ts
@@ -285,18 +285,17 @@ export class ConversationService {
         reportMissing: isQualifiedUserClients(options.userIds), // we want to check mismatch in case the consumer gave an exact list of users/devices
         onClientMismatch: options.onClientMismatch,
       });
-    } else {
-      if (!isStringArray(userIds) && !isUserClients(userIds)) {
-        throw new Error('Invalid userIds option for sending');
-      }
-      const recipients = await this.getRecipientsForConversation(conversationId, userIds);
-      return this.messageService.sendMessage(sendingClientId, recipients, plainText, {
-        conversationId,
-        sendAsProtobuf: options.sendAsProtobuf,
-        reportMissing: isUserClients(options.userIds), // we want to check mismatch in case the consumer gave an exact list of users/devices
-        onClientMismatch: options.onClientMismatch,
-      });
     }
+    if (!isStringArray(userIds) && !isUserClients(userIds)) {
+      throw new Error('Invalid userIds option for sending');
+    }
+    const recipients = await this.getRecipientsForConversation(conversationId, userIds);
+    return this.messageService.sendMessage(sendingClientId, recipients, plainText, {
+      conversationId,
+      sendAsProtobuf: options.sendAsProtobuf,
+      reportMissing: isUserClients(options.userIds), // we want to check mismatch in case the consumer gave an exact list of users/devices
+      onClientMismatch: options.onClientMismatch,
+    });
   }
 
   private generateButtonActionGenericMessage(payloadBundle: ButtonActionMessage): GenericMessage {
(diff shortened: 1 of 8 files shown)


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -103,12 +103,12 @@ export class ConversationService {
...
   constructor(
     private readonly apiClient: APIClient,
-    private readonly cryptographyService: CryptographyService,
+    cryptographyService: CryptographyService,
     private readonly assetService: AssetService,
   ) {
     this.messageTimer = new MessageTimer();
     this.messageBuilder = new MessageBuilder(this.apiClient, this.assetService);
-    this.messageService = new MessageService(this.apiClient, this.cryptographyService);
+    this.messageService = new MessageService(this.apiClient, cryptographyService);
```
atomrc, line 111: Now, the `convesationService` doesn't need to use the `cryptoService` anymore, it just passes it on to the message service.

packages/core/src/main/conversation/message/MessageService.ts
```diff
@@ -32,139 +32,108 @@ import {
...
-  public async sendOTRMessage(
+  /**
+   * Sends a message to a non-federated backend.
+   * @param {string} sendingClientId
+   * @param {UserClients|UserPreKeyBundleMap} recipients
+   * @param {Uint8Array} plainText
+   * @param {any} options:{conversationId?:string;reportMissing?:boolean;sendAsProtobuf?:boolean;onClientMismatch?:(mismatch:ClientMismatch
+   * @returns {any}
+   */
+  public async sendMessage(
```
atomrc, line 58: now there are only 2 public methods in the MessageService:
- sendMessage
- sendFederatedMessage

All the routing to (`sendExternal`, `sendProtobuf`, `sendRegular`...) is done internally and simplify the work for the consumer (as you can see, the `ConversationService` has been trimmed from this kind of logic)

bennycode:
@atomrc, my application broke with error message "Error: Invalid userIds option for sending" when updating from @wireapp/bot-api v7.17.1 to v7.17.9. 

Strack trace is:

```
    at ConversationService.sendGenericMessage (@wireapp\core\src\main\conversation\ConversationService.ts:291:13)
    at ConversationService.send (@wireapp\core\src\main\conversation\ConversationService.ts:843:33)
    at WireCommandHandler.sendQuote (@wireapp\bot-api\src\MessageHandler.ts:338:47)
    at WireCommandHandler.sendReply (@wireapp\bot-api\src\MessageHandler.ts:351:17)
    at WireCommandHandler.handleEvent (src\messenger\wire\WireCommandHandler.ts:36:23)
```

My `WireCommandHandler` passes a conversation ID (UUIDv4), quoted message (string) & reply message (string) to [MessageHandler.sendReply](https://github.com/wireapp/wire-web-packages/blob/main/packages/bot-api/src/MessageHandler.ts#L345). What do I have to change to make it work again?

atomrc:
@bennycode yes there was a regression introduced with it. 
The version of the core from yesterday has the fix in it. There should also be a new version of the bot-api that goes with it `7.17.48`
Please let me know it fixes your problem

bennycode:
Ah, v7.17.48 fixes it. Thank you very much. 😊

atomrc:
Awesome. Thanks for the feedback :)

PR #4225, opened 2022-02-16
feat(api-client): adding 2-factor verification codes (SQSERVICES-1158)

tlebon:
Adds optional verification codes and a new endpoint for generating verification codes via email 
in accord with https://wearezeta.atlassian.net/wiki/spaces/SER/pages/557580675/Tech+spec+BSI+-+1.8+Improve+User+Authentication

## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/auth/AuthAPI.ts
+++ b/packages/api-client/src/auth/AuthAPI.ts
@@ -28,6 +28,7 @@ import type {CookieList} from './CookieList';
 import type {LoginCodeResponse} from './LoginCodeResponse';
 import type {RegisterData} from './RegisterData';
 import type {User} from '../user/';
+import {VerificationActionType} from './VerificationActionType';
 
 export class AuthAPI {
   constructor(private readonly client: HttpClient) {}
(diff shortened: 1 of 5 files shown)


atomrc (approved):

packages/api-client/src/auth/AuthAPI.ts
```diff
@@ -77,8 +77,10 @@ export class AuthAPI {
   }
 
   public async postLogin(loginData: LoginData): Promise<AccessTokenData> {
+    const {verificationCode, ...rest} = loginData;
     const login = {
-      ...loginData,
+      ...rest,
+      ...(verificationCode && {verification_code: verificationCode}),
```
thisisamir98, line 83: isn't it better if we do it like this?
what is added to the `login` object if `&&` condition is false? 
```
if (verificationCode) {
  login.verification_code = verificationCode;
}
```

tlebon: these are the same 
https://blog.agney.dev/optional-spreading/

PR #4229, opened 2022-02-22
feat(api-client): Prefix all api requests with version number

atomrc:
Will add the version prefix before any api call request


atomrc:
Not to merge now as backend still have some work to do on the `cookie` for the `/access` endpoint

tlebon (approved):

PR #4246, opened 2022-03-18
feat: add optional verification code to delete team endpoint

tlebon:
## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/team/team/TeamAPI.ts
+++ b/packages/api-client/src/team/team/TeamAPI.ts
@@ -78,10 +78,11 @@ export class TeamAPI {
     return response.data;
   }
 
-  public async deleteTeam(teamId: string, password: string): Promise<void> {
+  public async deleteTeam(teamId: string, password: string, verificationCode?: string): Promise<void> {
     const config: AxiosRequestConfig = {
       data: {
         password,
+        verificationCode,
       },
       method: 'delete',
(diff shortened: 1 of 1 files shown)


packages/api-client/src/team/team/TeamAPI.ts
```diff
@@ -78,10 +78,11 @@ export class TeamAPI {
     return response.data;
   }
 
-  public async deleteTeam(teamId: string, password: string): Promise<void> {
+  public async deleteTeam(teamId: string, password: string, verificationCode?: string): Promise<void> {
     const config: AxiosRequestConfig = {
       data: {
         password,
+        ...(verificationCode && {verification_code: verificationCode}),
```
ffflorian, line 85:
```suggestion
        verification_code: verificationCode,
```

Or is it a problem when sending an undefined verification code?

ffflorian (changes requested):

packages/api-client/src/team/team/TeamAPI.ts
tlebon: ive just double checked, it is not a problem (according to the BE), i thought in general it would be better to not pass one if one is not required

ffflorian: True, we just never did it this way. But either way is fine 🙂

ffflorian (approved):

PR #4250, opened 2022-03-29
feat(api-client): Add conversation guest link feature endpoint

atomrc:


packages/api-client/src/conversation/ConversationAPI.ts
```diff
@@ -201,6 +201,16 @@ export class ConversationAPI {
     return response.data;
   }
 
+  public async getConversationGuestLinkFeature(conversationId: string): Promise<{status: 'enabled' | 'disabled'}> {
```
ffflorian, line 204: What do you think about moving `{status: 'enabled' | 'disabled'}` to its own type?

ffflorian (changes requested):

packages/api-client/src/conversation/ConversationAPI.ts
atomrc: Sure, will do

PR #4257, opened 2022-05-02
refactor: Cleanup storage initialization

atomrc:
This PR harmonizes how an already initialized database can be given to the core.

Initially, there was 3 different methods where a db could be given. They would all have slightly different behavior and would mean exactly the same thing:
- either in the constructor of the Core, give a `storeEngineProvider` that will be called when a database needs to be accessed by the core
- or in the `login` or `init` methods give a direct instance of a database
- or overwrite the `storageEngine` of the instance directly (hack). 

Now there is only a single way to give an already existing database: add the `createStore` option in the constructor. 
This function will be called with the `storeName` and `context` so that the consumer can decide what kind of database should be given depending on the type of the device add (permanent, temporary or none). 

BREAKING CHANGE: 
- The constructor signature now changes.
If you were doing

```js
const account = new Account(apiClient, createStoreEngine);
```

Now you need to do 

```js
const account = new Account(apiClient, {createStore: createStoreEngine});
```

- The `login` and `init` function do not take a storage engine parameter anymore. You now need to give a `createStore` function to the constructor in order to give a custom storage engine to the core. 

BEFORE

```js
const account = new Accoun(apiClient);

account.login(data, initClient, clientInfo, database);
```

AFTER

```js
const account = new Accoun(apiClient, {createStore: () => database});

account.login(data, initClient, clientInfo);
```

### DIFF
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -95,12 +95,12 @@ export interface Account {
   on(event: TOPIC.ERROR, listener: (payload: CoreError) => void): this;
 }
 
-export type StoreEngineProvider = (storeName: string) => Promise<CRUDEngine>;
+export type CreateStoreFn = (storeName: string, context: Context) => undefined | Promise<CRUDEngine | undefined>;
 
 export class Account extends EventEmitter {
   private readonly apiClient: APIClient;
   private readonly logger: logdown.Logger;
-  private readonly storeEngineProvider: StoreEngineProvider;
+  private readonly createStore: CreateStoreFn;
   private storeEngine?: CRUDEngine;
 
   public static readonly TOPIC = TOPIC;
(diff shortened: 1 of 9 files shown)


packages/core/src/main/Account.ts
```diff
@@ -123,21 +123,16 @@ export class Account extends EventEmitter {
 
   /**
    * @param apiClient The apiClient instance to use in the core (will create a new new one if undefined)
-   * @param storeEngineProvider Used to store info in the database (will create a inMemory engine if undefined)
+   * @param storeEngineProvider Used to store info in the database (will create a inMemory engine if returns undefined)
    */
-  constructor(apiClient: APIClient = new APIClient(), storeEngineProvider?: StoreEngineProvider) {
+  constructor(
+    apiClient: APIClient = new APIClient(),
+    {createStore = () => undefined, enableMls}: {createStore?: CreateStoreFn; enableMls?: boolean} = {},
```
tlebon, line 130: I think i prefer capitalizing all of an abbreviation for readability `enableMLS` but this is just a suggestion

atomrc: Rigth, makes sense. Actually I should remove it altogether from this PR, since it's not the direct topic here. Good catch :)

tlebon (approved):

thisisamir98 (approved):

PR #4259, opened 2022-05-09
feat: Add MLS device creation types

atomrc:


packages/core/src/main/CoreCrypto.ts
```diff
@@ -0,0 +1,229 @@
+/*
```
atomrc, line 1: Please ignore this file, it's just a placeholder for the future coryCrypto bundle

packages/core/src/main/Account.ts
```diff
@@ -290,10 +294,22 @@ export class Account extends EventEmitter {
...
     this.apiClient.context!.clientId = loadedClient.id;
+    if (this.enableMLS) {
+      this.coreCryptoClient = await this.createCoreCryptoClient(loadedClient);
+    }
 
     return loadedClient;
   }
 
+  private async createCoreCryptoClient(client: RegisteredClient): Promise<CoreCrypto> {
+    const {CoreCrypto} = await import('./CoreCrypto');
```
atomrc, line 305: Since the coreCrypto bundle might be quite big, we want to lazy load it, only if the consumer creates an mls client. 

Other users will never download the bundle

PR #4263, opened 2022-05-16
feat: Allow consumer to set number of prekeys generated

atomrc:
Will allow the consumer to set a number of prekeys to generate. 

As of now it defaults to 1, which means that we only ever use the last resort prekey. 
This will allow generating more prekeys and let cryptobox generate new as they are consumed

### DIFF
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -100,7 +100,16 @@ export type CreateStoreFn = (storeName: string, context: Context) => undefined |
 interface AccountOptions {
   /** Used to store info in the database (will create a inMemory engine if returns undefined) */
   createStore?: CreateStoreFn;
-  /** Number of prekeys to generate when creating a new device (defaults to 100) */
+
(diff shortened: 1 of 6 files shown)


packages/core/src/main/Account.ts
```diff
@@ -97,11 +97,28 @@ export interface Account {
...
 
+interface AccountOptions {
+  /** Used to store info in the database (will create a inMemory engine if returns undefined) */
+  createStore?: CreateStoreFn;
+
+  /** Number of prekeys to generate when creating a new device (defaults to 100)
+   * Prekeys are a way to create unique secured sessions between two devices.
+   * Having a high value will:
+   *    - make creating a new device consuming more CPU resources
+   *    - make it possible that the user creates multiple session with multiple users without needing to create new prekeys (or be connected)
```
comawill, line 108:
```suggestion
   *    - make it less likely that all prekeys get consumed while the device is offline and the last resort prekey will not be used to create new session
```

comawill (changes requested):

comawill (approved):

PR #4264, opened 2022-05-17
fix: Avoid silently creating client when initializing core

atomrc:
BREAKING CHANGE: if you were relying on `account.init()` to also create a new client, this behavior now change and will throw an error instead. If you also want to create a device (if it doesn't exist already), you should use the `account.login()` or `account.initClient()` methods instead


packages/core/src/main/Account.ts
```diff
@@ -265,6 +254,10 @@ export class Account extends EventEmitter {
       const localClient = await this.loadAndValidateLocalClient();
       return {isNewClient: false, localClient};
     } catch (error) {
+      if (!clientInfo) {
+        // If no client info provided, the client should not be created
+        throw error;
```
atomrc, line 259: This is the main change (all the rest is just documentation and moving methods around)

tlebon (approved):

PR #4268, opened 2022-05-24
feat:: Improve websocket connection API

atomrc:
Will allow the consumer to get warned for all types of messages that arrive on the websocket.


packages/core/src/main/notification/NotificationService.ts
```diff
@@ -145,26 +124,30 @@ export class NotificationService extends EventEmitter {
...
-        data = await this.handleEvent(event, source);
+        const data = await this.handleEvent(event, source);
         if (!notification.transient) {
+          // keep track of the last handled notification for next time we fetch the notification stream
           await this.setLastNotificationId(notification);
         }
+        yield {
+          ...data,
+          mappedEvent: data.mappedEvent ? this.cleanupPayloadBundle(data.mappedEvent) : undefined,
+        };
```
atomrc, line 150: Instead of sending events, the method is now a generator function that warns its parent when an event has been processed. 
This way we have a clear parent>child relationship and the flow is easier to follow

thisisamir98 (approved):

PR #4274, opened 2022-05-30
feat: Prepare new select component

przemvs:
## Pull Request Checklist

### DIFF
--- a/packages/react-ui-kit/src/Form/__snapshots__/Select.test.tsx.snap
+++ b/packages/react-ui-kit/src/Form/__snapshots__/Select.test.tsx.snap
@@ -124,7 +124,6 @@ exports[`"Select" renders 1`] = `
       </button>
       <ul
         className="emotion-4"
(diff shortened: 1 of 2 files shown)


packages/react-ui-kit/src/Form/__snapshots__/Select.test.tsx.snap
```diff
@@ -35,77 +39,97 @@ exports[`"Select" renders 1`] = `
...
-        a
-      </option>
-      <option>
-        b
-      </option>
-    </select>
+      </button>
+      <ul
+        className="emotion-4"
+        data-uie-name="dropdown-undefined"
```
atomrc, line 127: Maybe we could remove `uie-name` if the value is `undefined`

atomrc, line 127: Maybe we could remove `uie-name` if the value is `undefined`

PR #4286, opened 2022-06-08
fix: Improve Checkbox and Select behaviors

przemvs:
## Pull Request Checklist

### DIFF
--- a/packages/react-ui-kit/src/Form/Checkbox.tsx
+++ b/packages/react-ui-kit/src/Form/Checkbox.tsx
@@ -25,7 +25,7 @@ import {COLOR, COLOR_V2} from '../Identity';
 import {Theme} from '../Layout';
 import {Text, TextProps, textStyle, textLinkStyle} from '../Text';
 import {filterProps} from '../util';
-import {INPUT_CLASSNAME, Input, InputProps} from './Input';
+import {INPUT_CLASSNAME, InputProps} from './Input';
 
 export interface StyledLabelProps<T = HTMLLabelElement> extends React.HTMLProps<T> {
   disabled?: boolean;


packages/react-ui-kit/src/Form/Checkbox.tsx
```diff
@@ -19,7 +19,7 @@
 
 /** @jsx jsx */
 import {jsx} from '@emotion/react';
-import React, {useId} from 'react';
+import React from 'react';
```
atomrc, line 22: Why removing the `useId`? It seems better than `Math.random().toString()`

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -27,23 +27,25 @@ import {inputStyle} from './Input';
...
   value: string | number;
   label: string;
+  description?: string;
 };
 
-export interface SelectProps {
+export interface SelectProps<T extends SelectOption> {
   id: string;
-  onChange: (selectedOption: string | number) => void;
+  onChange: (selectedOption: T['value']) => void;
```
atomrc, line 38: :+1:

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -135,7 +137,7 @@ const filterSelectProps = props => filterProps(props, ['markInvalid']);
 
 const placeholderText = '- Please select -';
 
-export const Select = ({
+export const Select = <T extends SelectOption>({
```
atomrc, line 140:
```suggestion
export const Select = <T extends SelectOption = SelectOptions>({
```

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -242,13 +247,22 @@ export const Select = ({
     };
   }, []);
 
+  useEffect(() => {
+    if (value) {
+      const valueIdx = options.findIndex(option => option.value === value.value);
+      setSelectedOption(valueIdx);
+    }
+  }, [options, value]);
```
atomrc, line 255: You don't need a `useEffect` here. You are just creating new state from existing state, there is no effect there

atomrc (changes requested):

packages/react-ui-kit/src/Form/Checkbox.tsx
atomrc: What is the error exactly?

atomrc (approved):

packages/react-ui-kit/src/Form/Checkbox.tsx
tlebon: did this get resolved?

packages/react-ui-kit/src/Form/Checkbox.tsx
```diff
@@ -93,10 +93,11 @@ interface CheckboxProps<T = HTMLInputElement> extends InputProps<T> {
 
 const filterCheckboxProps = (props: CheckboxProps) => filterProps(props, ['markInvalid']);
 
+// We use Math.random..., because some of apps doesn't migrated to newest version of React.
```
tlebon, line 96: resolves comment on L22

PR #4287, opened 2022-06-08
feat(core): Create MLS capable devices

atomrc:
This will push the public keys and keypackages to the backend when creating a new device with the `enableMLS` flag enabled


.gitattributes
```diff
@@ -12,3 +12,4 @@
 *.ttf binary
 *.woff binary
 *.woff2 binary
+*.tgz binary
```
atomrc, line 15: This can be useful in the future (although not needed for now). considering `tgz` as anything else than binary will most likely break the archive

packages/core/src/main/Account.ts
```diff
@@ -352,10 +357,22 @@ export class Account extends EventEmitter {
...
+    }
 
     return loadedClient;
   }
 
+  private async createMLSClient(client: RegisteredClient): Promise<CoreCrypto> {
+    const {CoreCrypto} = await import('@otak/core-crypto');
+    return CoreCrypto.init({
+      path: 'path/to/database',
+      key: 'a root identity key (i.e. enclaved encryption key for this device)',
```
tlebon, line 371: just curious about hese hardcoded values right now- are these just filler for the time being?

atomrc: Yes, they are placeholders. As of now, the core-crypto bundle doesn't have a database and doesn't need this part to work. We will fill them with the relevant data when it's time to do so :)

packages/core/src/main/Account.ts
```diff
@@ -364,12 +382,22 @@ export class Account extends EventEmitter {
...
     }
+    this.logger.info(`Creating new client {mls: ${!!this.enableMLS}}`);
     const registeredClient = await this.service.client.register(loginData, clientInfo, entropyData);
+    if (this.enableMLS) {
+      this.coreCryptoClient = await this.createMLSClient(registeredClient);
+      await this.service.client.uploadMLSPublicKeys(this.coreCryptoClient.clientPublicKey(), registeredClient.id);
+      await this.service.client.uploadMLSKeyPackages(
+        this.coreCryptoClient.clientKeypackages(this.nbPrekeys),
+        registeredClient.id,
+      );
```
thisisamir98, line 393: isn't it better to wrap this part in a try-catch and have a nice error message for it?

PR #4288, opened 2022-06-08
feat(api-client): Add support for creating mls based conversations (FS-554)

thisisamir98:
## Pull Request Checklist


packages/api-client/src/conversation/ConversationAPI.ts
```diff
@@ -451,7 +451,7 @@ export class ConversationAPI {
    * @param conversationData The new conversation
    * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/conversations/createOne2OneConversation
    */
-  public async post1to1(conversationData: NewConversation): Promise<void> {
+  public async post1to1(conversationData: Omit<NewConversation, 'protocol'>): Promise<void> {
```
atomrc, line 454: Why do you omit the protocol here? It's valid to create a 1:1 mls conversation, I believe

PR #4293, opened 2022-06-15
fix(core): Make sure access token is valid before connecting to websocket

atomrc:
We had a nasty race condition when reconnecting to the websocket after the access token was expired. In this race condition both the notification stream request and the websocket would both trigger an access token refresh ending up in both coliding with each other

## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/auth/AccessTokenStore.ts
+++ b/packages/api-client/src/auth/AccessTokenStore.ts
@@ -30,11 +30,19 @@ export interface AccessTokenStore {
   on(event: TOPIC.ACCESS_TOKEN_REFRESH, listener: (accessToken: AccessTokenData) => void): this;
 }
 
+type AccessTokenWithExpirationDate = AccessTokenData & {
+  /**
+   * The date at which the token will be invalid. This value should be use with a grain of salt as there might be some time shift between browser and server time
+   * It is suggested to add an error margin
+   */
+  expireAt: number;
+};
+
 export class AccessTokenStore extends EventEmitter {
   private readonly logger: logdown.Logger;
 
   public static readonly TOPIC = TOPIC;
-  public accessToken?: AccessTokenData;
+  public accessToken?: AccessTokenWithExpirationDate;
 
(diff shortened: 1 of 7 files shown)


packages/api-client/src/auth/AccessTokenStore.ts
```diff
@@ -30,11 +30,19 @@ export interface AccessTokenStore {
   on(event: TOPIC.ACCESS_TOKEN_REFRESH, listener: (accessToken: AccessTokenData) => void): this;
 }
 
+type AccessTokenWithExpirationDate = AccessTokenData & {
+  /**
+   * The date at which the token will be invalid. This value should be use with a grain of salt as there might be some time shift between browser and server time
+   * It is suggested to add an error margin
+   */
+  expireAt: number;
```
thisisamir98, line 38: shouldn't this be included in the access token interface itself?

atomrc: Well the `AccessTokenData` is the raw response from backend (which doesn't include the expireAt), we compute it only here, so I think it makes sense to have a new type on top of `AccessTokenData`. WDYT?

atomrc: But maybe you are right, I should not add it to the same object but rather another property of the store, let me try this

thisisamir98 (approved):

packages/api-client/src/auth/AccessTokenStore.ts
tlebon: whats the difference between expiresIn / expiresAt

tlebon (approved):

packages/api-client/src/auth/AccessTokenStore.ts
atomrc: expiresIn is the time that the token is valid (15min)
expiresAt is the Date at which the token will be invalid (June 15th 18:35 for example)

expiresAt is computed from expiresIn

PR #4295, opened 2022-06-22
feat(core, api-client): Add MLS group creation

atomrc:


thisisamir98 (approved):

packages/api-client/src/conversation/ConversationAPI.ts
```diff
@@ -546,6 +546,31 @@ export class ConversationAPI {
     }
   }
 
+  /**
+   * Create a new conversation
+   * @param conversationData The new conversation
+   * @see https://staging-nginz-https.zinfra.io/api/swagger-ui/#/default/post_conversations
+   */
+  public async postConversationV2(conversationData: NewConversation): Promise<Conversation> {
```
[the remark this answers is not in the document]
thisisamir98: https://github.com/wireapp/wire-web-packages/pull/4295/commits/4879a4f43bf178a1d3be23a1ba302caaf852a52a

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -916,9 +920,85 @@ export class ConversationService {
...
+        /**
+         * we should skip fetching key packages for current self client,
+         * it's already added by the backend on the group creation time
+         */
+        conversationData.creator_client,
+      ),
+      ...qualifiedUsers.map(qualifiedId =>
+        this.apiClient.api.client.claimMLSKeyPackages(qualifiedId.id, qualifiedId.domain),
+      ),
+    ]);
```
tlebon, line 966: *idea*: not sure if its cleaner to read but occurs to me we could just push the selfUser data into the qualifiedId list and then just spread one call inside the promise.all.

thisisamir98: yes but that way we're gonna need extra logic for conversationData.creator_client which we are passing as skip_own to not fetch key package for current self client

tlebon: ah yes i was thinking we could just pass it in once but it wouldn't work exactly that way. lets leave it as is

PR #4296, opened 2022-06-22
feat: modify StyledApp to accept either theme object or themeId as prop

V-Gira:
Modify StyledApp and ThemeProvider so you can pass either a themeId for the prebuilt light/dark theme or a complete theme object as StyledApp prop.

### DIFF
--- a/packages/react-ui-kit/src/Layout/Theme.tsx
+++ b/packages/react-ui-kit/src/Layout/Theme.tsx
@@ -34,6 +34,7 @@ export interface Theme {
   general: {
     backgroundColor: string;
     color: string;
+    themeColor?: string;
   };
   Input: {
     backgroundColor: string;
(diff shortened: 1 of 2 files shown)


packages/react-ui-kit/src/Layout/Theme.tsx
```diff
@@ -68,11 +69,12 @@ export const themes: {[themeId in THEME_ID]: Theme} = {
 };
 
 export interface ThemeProps<T = HTMLDivElement> extends React.HTMLProps<T> {
-  themeId: THEME_ID;
+  themeId?: THEME_ID;
```
atomrc, line 72: Is this still needed somewhere?

packages/react-ui-kit/src/Layout/Theme.tsx
```diff
@@ -34,6 +34,7 @@ export interface Theme {
   general: {
     backgroundColor: string;
     color: string;
+    themeColor?: string;
```
atomrc, line 37: Unrelated change?

atomrc (changes requested):

packages/react-ui-kit/src/Layout/Theme.tsx
atomrc: That said, I like the idea of having this generic `themeColor` that could be applied across all components :ok_hand:

PR #4299, opened 2022-06-24
feat: add custom theming to UI kit

V-Gira:
### Issue

- We need to be able to use Accent colors from the Web App with the UI-Kit components
- The addition of custom theming should not affect the Vanilla UI-Kit

### Solution

- Allow for custom theming by being able to pass an entire theme object to the StyledApp component
- Modify components to use accent-colors from the custom theme or default colors with default themes


packages/react-ui-kit/src/Form/Checkbox.tsx
```diff
@@ -34,8 +34,10 @@ export interface StyledLabelProps<T = HTMLLabelElement> extends React.HTMLProps<
 
 const filterStyledLabelProps = (props: StyledLabelProps) => filterProps(props, ['markInvalid']);
 
-const checkSvg =
-  '<svg width="15" height="13" viewBox="0 0 16 13" xmlns="http://www.w3.org/2000/svg"><path d="M5.65685 12.0711L15.9842 1.62738L14.57 0.213167L5.65685 9.24264L1.41421 5L0 6.41421L5.65685 12.0711Z" fill="white"/></svg>';
+const checkSvg = (theme: Theme) =>
+  `<svg width="15" height="13" viewBox="0 0 16 13" xmlns="http://www.w3.org/2000/svg"><path d="M5.65685 12.0711L15.9842 1.62738L14.57 0.213167L5.65685 9.24264L1.41421 5L0 6.41421L5.65685 12.0711Z" fill="${
+    theme.darkMode ? 'black' : 'white'
+  }"/></svg>`;
```
atomrc, line 40:
```suggestion
  `<svg width="15" height="13" viewBox="0 0 16 13" xmlns="http://www.w3.org/2000/svg"><path d="M5.65685 12.0711L15.9842 1.62738L14.57 0.213167L5.65685 9.24264L1.41421 5L0 6.41421L5.65685 12.0711Z" style={fille: ${theme.maincolor /* or equivalent */}}"/></svg>`;
```

Can we try this?

V-Gira: If I'm not missing something, that's precisely what's not working :upside_down_face:

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -48,11 +48,14 @@ export interface SelectProps<T extends SelectOption = SelectOption> {
...
 
-const ArrowDown = (theme: Theme) => `
-    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
-        <path fill="${theme.general.color}" fill-rule="evenodd" clip-rule="evenodd" d="M7.99963 12.5711L15.6565 4.91421L14.2423 3.5L7.99963 9.74264L1.75699 3.5L0.342773 4.91421L7.99963 12.5711Z"/>
-    </svg>
-`;
+const ArrowDown = (
+  <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
+    <path
+      fill="${black}"
```
atomrc, line 54: This looks a little suspicious, I don't see any `black` variable in the scope and a variable named `black` seems weird :thinking:

packages/react-ui-kit/src/Layout/StyledApp.tsx
```diff
@@ -23,11 +23,12 @@ import {CSSObject, jsx} from '@emotion/react';
...
 import {GlobalStyle} from '../GlobalStyle';
 import {filterProps} from '../util';
-import {THEME_ID, Theme, ThemeProvider} from './Theme';
+import {THEME_ID, Theme, ThemeProvider, themes} from './Theme';
 
 export interface StyledAppContainerProps<T = HTMLDivElement> extends React.HTMLProps<T> {
   backgroundColor?: string;
   themeId?: THEME_ID;
+  theme?: Theme;
 }
```
atomrc, line 32: You haven't solved my challenge yet ;) 
I'd like this type to only allow for a `themeId` or a `theme` but not both ;)

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -65,30 +68,36 @@ export const selectStyle: <T>(theme: Theme, props, error?: boolean) => CSSObject
...
   '&:disabled': {
-    color: COLOR_V2.GRAY,
+    color: theme.select.disabledColor,
   },
   appearance: 'none',
-  background: disabled
-    ? `${theme.Input.backgroundColorDisabled} center right 16px no-repeat url("${inlineSVG(ArrowDown(theme))}")`
-    : `${theme.Input.backgroundColor} center right 16px no-repeat url("${inlineSVG(ArrowDown(theme))}")`,
-  boxShadow: markInvalid ? `0 0 0 1px ${COLOR_V2.RED}` : `0 0 0 1px ${COLOR_V2.GRAY_40}`,
+  boxShadow: markInvalid ? `0 0 0 1px ${COLOR_V2.RED}` : `0 0 0 1px ${theme.select.borderColor}`,
```
atomrc, line 74: maybe the invalid color needs to be a Theme variable as well (as it might be different in dark/light mode). Doesn't it?

packages/react-ui-kit/src/Form/Select.tsx
V-Gira: Good catch, this is here just for testing, needed to be removed

packages/react-ui-kit/src/Form/Select.tsx
V-Gira: I'll use standard red for now (@red-500)

PR #4302, opened 2022-06-27
feat: Add encrypted storage for coreCrypto database key [FS-565]

atomrc:
## Pull Request Checklist

### DIFF
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -49,7 +49,7 @@ import {AccountService} from './account/';
 import {LinkPreviewService} from './linkPreview';
 import type {CoreCrypto} from '@otak/core-crypto';
 import {WEBSOCKET_STATE} from '@wireapp/api-client/src/tcp/ReconnectingWebsocket';
-import {createEncryptedStore, IEncryptedStore} from './util/encryptedStore';
+import {createCustomEncryptedStore, createEncryptedStore} from './util/encryptedStore';
 
 export type ProcessedEventPayload = HandledEventPayload;
 
@@ -96,14 +96,19 @@ export interface Account {
 }
 
 export type CreateStoreFn = (storeName: string, context: Context) => undefined | Promise<CRUDEngine | undefined>;
-type CreateSecretStoreFn = (dbName: string) => IEncryptedStore | Promise<IEncryptedStore>;
+type SecretCrypto<T> = {
+  encrypt: (value: Uint8Array) => Promise<T>;
+  decrypt: (payload: T) => Promise<Uint8Array>;
+};
 
-interface AccountOptions {
+interface AccountOptions<T> {
   /** Used to store info in the database (will create a inMemory engine if returns undefined) */
   createStore?: CreateStoreFn;
 
-  /** Storage used to store sensitive secret values. If not provided, will use an encrypted indexeddb database */
-  createSecretStore?: CreateSecretStoreFn;
+  /** encrypt/decrypt function pair that will be called before storing/fetching secrets in the secrets database.
+   * If not provided will use the built in encryption mechanism
+   */
+  secretsCrypto?: SecretCrypto<T>;
 
   /** Number of prekeys to generate when creating a new device (defaults to 2)
    * Prekeys are Diffie-Hellmann public keys which allow offline initiation of a secure Proteus session between two devices.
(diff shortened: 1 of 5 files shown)


packages/core/src/main/Account.ts
```diff
@@ -95,11 +96,14 @@ export interface Account {
 }
 
 export type CreateStoreFn = (storeName: string, context: Context) => undefined | Promise<CRUDEngine | undefined>;
+type CreateSecretStoreFn = (dbName: string) => IEncryptedStore | Promise<IEncryptedStore>;
 
 interface AccountOptions {
   /** Used to store info in the database (will create a inMemory engine if returns undefined) */
   createStore?: CreateStoreFn;
 
+  createSecretStore?: CreateSecretStoreFn;
```
tlebon, line 105: maybe a 1-liner JSDoc for this also?

packages/core/src/main/util/encryptedStore.test.node.ts
```diff
@@ -0,0 +1,40 @@
+/*
+ * Wire
+ * Copyright (C) 2021 Wire Swiss GmbH
```
tlebon, line 3: oops
```suggestion
 * Copyright (C) 2022 Wire Swiss GmbH
```

packages/core/src/main/util/encryptedStore.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+class EncryptedStore implements IEncryptedStore {
+  readonly #key: CryptoKey;
+  constructor(key: CryptoKey, private readonly db: IDBPDatabase<EncryptedDB>) {
+    this.#key = key;
+  }
+
+  async saveSecretValue(primaryKey: string, value: Uint8Array) {
+    const iv = await window.crypto.getRandomValues(new Uint8Array(12));
```
tlebon, line 49: should we think about our friends with a lack of trust for getRandomValues? 

also: should we consider a node case like we do in RandomUtil.ts?

packages/core/src/main/util/encryptedStore.test.node.ts
atomrc: Thanks :pray:

packages/core/src/main/util/encryptedStore.ts
atomrc: Talked with Sebastian, this is fine, they just need custom entropy value for identity generation :ok_hand:

packages/core/src/main/util/encryptedStore.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/**
+ * Will create a database that uses a custom encryption method. It needs the encrypt and decrypt function to be able to process the values stored.
+ * It's the responsability of the consumer to store the encryption key
+ *
+ * @param dbName the name of the database to create
+ * @param config contains the encrypt and decrypt methods
+ */
+export async function createCustomEncryptedStore<EncryptedPayload>(
```
arjita-mitra, line 124: can you explain the use case of custom encryption?

atomrc: If the consumer want to use their own encryption method, the lib allows them to do so. 

A concrete example of this is the electron wrapper. 
Electron has a native way of encrypting secrets (which is more secure that the browser's way). 
So on electron, we will have a custom encryption method that will be given to the core (see https://github.com/wireapp/wire-desktop/pull/5812/files for context) and the core can avoid using the less secure way of generating secrets and rely on electron's system :)

packages/core/src/main/Account.ts
```diff
@@ -119,11 +129,12 @@ const coreDefaultClient: ClientInfo = {
   model: '@wireapp/core',
 };
 
-export class Account extends EventEmitter {
+export class Account<T = unknown> extends EventEmitter {
   private readonly apiClient: APIClient;
   private readonly logger: logdown.Logger;
   private readonly createStore: CreateStoreFn;
   private storeEngine?: CRUDEngine;
+  private readonly secretsCrypto?: SecretCrypto<T>;
```
tlebon, line 137: i have a question about the usage of a generic here in general: 
from the code it seems like the type will either be of Uint8Array or EncryptedPayload (depending on the source). 

I am wondering if this will be extended (possibly) in the future or if there is someother reason to use a generic instead of a union?

atomrc: Since here we are from the point of view of a lib, we need to be as open as possible. 
One consumer could give a function that takes and returns a `string` or a `number` or any object that they want to use for encryption/decryption. We cannot really guess from the library perspective was will be used by the consumer.

tlebon (approved):
nothing seems particularly heinous to me 😅  LGTM

PR #4307, opened 2022-07-01
feat: Use mls config object

atomrc:
BREAKING CHANGE: the `enableMLS` flag has been removed in favor of a config object. If the config object is set, then MLS will be activated

### DIFF
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -102,14 +102,24 @@ type SecretCrypto<T> = {
   decrypt: (payload: T) => Promise<Uint8Array>;
 };
 
-interface AccountOptions<T> {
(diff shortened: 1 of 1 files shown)


packages/core/src/main/Account.ts
```diff
@@ -102,15 +102,25 @@ type SecretCrypto<T> = {
...
    * If not provided will use the built in encryption mechanism
    */
   secretsCrypto?: SecretCrypto<T>;
 
+  /**
+   * path on the public server to the core crypto wasm file.
+   * This file will be downloaded lazily when corecrypto is needed.
+   * It, thus, needs to know where, on the server, the file can be found
+   */
+  coreCrypoWasmFile: string;
```
thisisamir98, line 117: nice PR it would just more nice if we change this to coreCrypoWasmFilePath

PR #4308, opened 2022-07-04
feat(api-client): Adapt to api version 2

atomrc:
This is preparation work for supporting backend API v2. (cf. https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/603095166/API+changes+v1+v2)

### DIFF
--- a/packages/core/src/main/util/encryptedStore.test.node.ts
+++ b/packages/core/src/main/util/encryptedStore.test.node.ts
@@ -39,7 +39,7 @@ describe('encryptedStore', () => {
 
   describe('Store and restore secret values with custom encryption', () => {
     it('Stores secret values', async () => {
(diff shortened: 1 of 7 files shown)


packages/core/src/main/util/encryptedStore.test.node.ts
```diff
@@ -39,7 +39,7 @@ describe('encryptedStore', () => {
 
   describe('Store and restore secret values with custom encryption', () => {
     it('Stores secret values', async () => {
-      const store = await createCustomEncryptedStore('test', {
+      const store = await createCustomEncryptedStore('test-custom', {
```
atomrc, line 42: unrelated fix that fixes a flaky test (both tests were accessing the same DB with different schema)

thisisamir98 (approved):

PR #4309, opened 2022-07-04
runfix: Return correct users when creating mls conversations

thisisamir98:
Return correct users when creating mls conversation

### DIFF
--- a/packages/core/src/main/conversation/ConversationService.ts
+++ b/packages/core/src/main/conversation/ConversationService.ts
@@ -996,7 +996,8 @@ export class ConversationService {
       );
     }
     await Promise.all(sendingPromises);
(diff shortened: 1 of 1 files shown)


packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -996,7 +996,8 @@ export class ConversationService {
       );
     }
     await Promise.all(sendingPromises);
-    return newConversation;
+
+    return this.getConversations(newConversation.id);
```
atomrc, line 1000:
```suggestion
    // We the fresh version of the conversation created on backend with the newly added users
    return this.getConversations(newConversation.id);
```

atomrc (approved):

PR #4310, opened 2022-07-04
feat: update feature api

tlebon:
fix for deprecated endpoints, adding mls endpoint

## Pull Request Checklist

### DIFF
--- a/packages/api-client/src/team/feature/FeatureAPI.ts
+++ b/packages/api-client/src/team/feature/FeatureAPI.ts
@@ -66,9 +66,6 @@ export class FeatureAPI {
     return response.data;
   }
 
-  /**
-   * @deprecated Use `getAllFeatures()` instead.
-   */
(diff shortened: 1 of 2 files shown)


packages/api-client/src/team/feature/FeatureAPI.ts
```diff
@@ -66,9 +66,6 @@ export class FeatureAPI {
     return response.data;
   }
 
-  /**
-   * @deprecated Use `getAllFeatures()` instead.
```
atomrc, line 70: Actually, shouldn't it be the other one that is deprecated?

atomrc (approved):

packages/api-client/src/team/feature/FeatureAPI.ts
tlebon: this one is actually not deprecated as you say, however the regular GET feature-configs/ endpoint without a feature attached is also not deprecated

PR #4315, opened 2022-07-06
feat: Decrypt incoming MLS messages

atomrc:
This will allow the core to parse MLS `welcome` messages and decrypt incoming content messages

As of today (07/07/2022) this will only enable receiving messages from the person who **did not** create the conversation. 
The creator of the conversation cannot yet decrypt incoming messages. 

TODO:
- [x] Fix receiving messages for the conversation creator (will need to store the ConversationId => groupId mapping in the db)
- [x] Fix the DB storage of the conversationId => groupId mapping (use indexedDB instead of localStorage)


packages/core/src/main/notification/NotificationService.ts
```diff
@@ -231,8 +240,55 @@ export class NotificationService extends EventEmitter {
...
+        if (!coreCryptoClient) {
+          // TODO throw proper error
+          throw new Error('TODO');
+        }
+        const data = Decoder.fromBase64(event.data).asBytes;
+        // We extract the groupId from the welcome message and let coreCrypto store this group
+        const newGroupId = await coreCryptoClient.processWelcomeMessage(data);
+        const saveGroupId = (conversationId: QualifiedId, groupId: Uint8Array): void => {
+          // 🛑 TODO save the mapping between conversationId > groupID (not localStorage)
+          localStorage.setItem(`${conversationId.id}@${conversationId.domain}`, Encoder.toBase64(groupId).asString);
```
atomrc, line 255: This needs to change to a proper indexed DB storage

packages/core/src/main/notification/NotificationService.ts
```diff
@@ -231,8 +240,55 @@ export class NotificationService extends EventEmitter {
...
+
+      case Events.CONVERSATION_EVENT.MLS_MESSAGE_ADD:
+        if (!coreCryptoClient) {
+          // TODO throw proper error
+          throw new Error('TODO');
+        }
+        const encryptedData = Decoder.fromBase64(event.data).asBytes;
+        const findGroupId = (conversationId: QualifiedId) => {
+          // TODO fetch the groupId in DB from the conversatioId
+          const base64groupId = localStorage.getItem(`${conversationId.id}@${conversationId.domain}`);
```
atomrc, line 275: We need to get the groupId from the indexeddb store we used when the welcome message arrived

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -1,6 +1,6 @@
 /*
  * Wire
- * Copyright (C) 2018 Wire Swiss GmbH
+ * Copyright (C) 2022 Wire Swiss GmbH
```
tlebon, line 3: this shouldnt have been updated...

packages/api-client/src/event/ConversationEvent.ts
```diff
@@ -146,6 +154,16 @@ export interface ConversationOtrMessageAddEvent extends BaseConversationEvent {
...
+export interface ConversationMLSMessageAddEvent extends BaseConversationEvent {
+  data: ConversationMLSMessageAddData;
+  type: CONVERSATION_EVENT.MLS_MESSAGE_ADD;
+}
+
+export interface ConversationMLSWelcomeEvent extends BaseConversationEvent {
+  data: ConversationMLSWelcomeData;
+  type: CONVERSATION_EVENT.MLS_WELCOME_MESSAGE;
+}
+
```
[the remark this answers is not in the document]
tlebon: i like this personally but im not sure if i see why it shouldn't be like this.

packages/core/src/main/conversation/ConversationService.ts
```diff
@@ -113,6 +120,13 @@ function isMLS<T>(params: SendProteusMessageParams<T> | SendMlsMessageParams<T>)
 
 type SendProteusMessageParams<T> = SendCommonParams<T> &
   MessageSendingOptions & {
+    /**
+     * Can be either a QualifiedId[], string[], UserClients or QualfiedUserClients. The type has some effect on the behavior of the method. (Needed only for Proteus)
+     *    When given a QualifiedId[] or string[] the method will fetch the freshest list of devices for those users (since they are not given by the consumer). As a consequence no ClientMismatch error will trigger and we will ignore missing clients when sending
+     *    When given a QualifiedUserClients or UserClients the method will only send to the clients listed in the userIds. This could lead to ClientMismatch (since the given list of devices might not be the freshest one and new clients could have been created)
+     *    When given a QualifiedId[] or QualifiedUserClients the method will send the message through the federated API endpoint
+     *    When given a string[] or UserClients the method will send the message through the old API endpoint
+     */
```
tlebon, line 129: i like the documentation but this feels like a huge block of text

PR #4324, opened 2022-07-08
fix(react-ui-kit): fix select component current value empty (#SQSERVICES-1661)

PatrykBuniX:
`Select` component initial current value fix.

With this PR the "current value" state of the Select component depends only on the `value` prop.

## Pull Request Checklist


V-Gira (commented):
Looks good to me

tlebon (approved):

PR #4326, opened 2022-07-11
feat(react-ui-kit): add multiselect option to Select component

przemvs:
## Pull Request Checklist


V-Gira (commented):
Isn't this rolling back the fix @PatrykBuniX made on Friday?

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -47,6 +46,20 @@ export interface SelectProps<T extends SelectOption = SelectOption> {
...
+}
+
+interface MultipleSelectProps<T extends SelectOption = SelectOption> extends CommonSelectProps<T> {
+  onChange: (selectedOption: T[]) => void;
+  value?: T[] | null;
+}
+
+// export type SelectProps<T extends SelectOption = SelectOption, Multiple = boolean> = Multiple extends true ?
+//   MultipleSelectProps<T> :
+//   SingleSelectProps<T>
```
thisisamir98, line 61:
```suggestion
```

thisisamir98: this could be erased

tlebon:
needs to include checkboxes for selected options as discussed!

packages/react-ui-kit/src/Form/Checkbox.tsx
```diff
@@ -109,42 +114,49 @@ const filterCheckboxProps = (props: CheckboxProps) => filterProps(props, ['markI
...
+        style={{
+          height: '22px',
+          marginBottom: '0',
+          opacity: 0,
+          width: '22px',
+          cursor: disabled ? 'not-allowed' : 'pointer',
+        }}
+        disabled={disabled}
+        ref={ref}
+        className={clsx(INPUT_CLASSNAME, checkboxClassName)}
```
tlebon, line 150: do we need clsx for this? we cant combine the classes with emotion?

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -25,19 +25,20 @@ import {filterProps} from '../util';
 import {inputStyle} from './Input';
```
tlebon, line 25: this file is so big, can we figure out a way to split it up.

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -134,39 +165,41 @@ const dropdownOptionStyles = (theme: Theme, isSelected: boolean): CSSObject => (
...
-  dataUieName,
-  wrapperCSS = {},
-  ...props
-}: SelectProps<T>) => {
-  const currentOptionIdx = options.findIndex(option => option.value === value?.value);
-  const selectedOption = currentOptionIdx === -1 ? null : currentOptionIdx;
+const isSingle = (props: SingleSelectProps | MultipleSelectProps): props is SingleSelectProps =>
+  !(props as SingleSelectProps).isMultiSelect;
+
+const isMultiple = (props: SingleSelectProps | MultipleSelectProps): props is MultipleSelectProps =>
```
tlebon, line 177: isnt this just the opposite value of isSingle

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -134,39 +165,41 @@ const dropdownOptionStyles = (theme: Theme, isSelected: boolean): CSSObject => (
...
+    dataUieName,
+    wrapperCSS = {},
+    ...rest
+  } = props;
 
   const selectContainerRef = useRef<HTMLDivElement>(null);
   const listRef = useRef<HTMLUListElement>(null);
+
   const [isDropdownOpen, setIsDropdownOpen] = useState(false);
+  const [selectedOption, setSelectedOption] = useState(null);
```
tlebon, line 201: seems like we may run into similar issues with internal state

packages/react-ui-kit/src/Form/Select.tsx
tlebon: im sort of thinking about an abstracted general select and then two wrappers, one for single input select and one for multi-select. it could make the whole family more reusable and more readable

packages/react-ui-kit/src/Form/Select.tsx
PatrykBuniX: Yes, do we really need, to store selected option inside the component? This is basically what I've removed on last Friday. We should rather render select's UI based on its props (`value` props). 🤔

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -112,16 +124,35 @@ const dropdownStyles = (theme: Theme, isDropdownOpen: boolean): CSSObject => ({
...
+      }
+    : {
+        '&:hover, &:active, &:focus': {
+          background: theme.general.primaryColor,
+          borderColor: theme.general.primaryColor,
+          color: theme.Select.contrastTextColor,
+        },
+      }),
   '&:first-of-type': {
     borderRadius: '10px 10px 0 0',
```
V-Gira, line 157:
```suggestion
    borderRadius: '12px 12px 0 0',
```
Could you adress this feedback from Anette? (also in lines 116 and 160)
image

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -1,6 +1,6 @@
 /*
  * Wire
- * Copyright (C) 2018 Wire Swiss GmbH
+ * Copyright (C) 2022 Wire Swiss GmbH
```
aweiss-dev, line 3: I got told that we dont chage the copyright date, so it should be 2018 I guess

packages/react-ui-kit/src/Misc/useOutsideClick.ts
```diff
@@ -0,0 +1,40 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {RefObject, useEffect} from 'react';
+
```
aweiss-dev, line 21: is there a reason we define this ourselves instead of using the go-to library -> https://github.com/streamich/react-use

It already has a solution for that (useClickAway) and will help us in different other cases as well (probably). Through tree shaking it is also not bigger than defining the hooks ourselves.

packages/react-ui-kit/src/Form/SelectContainer.tsx
```diff
@@ -0,0 +1,290 @@
...
+  label,
+  error,
+  helperText,
+  options = [],
+  onOptionChange,
+  required,
+  markInvalid,
+  dataUieName,
+  wrapperCSS = {},
+  isMultiSelect = false,
```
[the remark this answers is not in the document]
tlebon: not really a full idea but maybe this can give you some idea of what i had in mind. 

options: SelectOption[] | multiSelectOption[]

const singleselectoptions = options.map((x)=>{
<SelectOption options={option} onClick={selectOptionclick}/>}

const multiselectoptions = options.map((x)=>{
<MultiSelectOption options={options} onclick={selectMultiOptionClick}/>}

//SingleSelect
SelectContainer options={singleselectoptions} ...
...
//MultiSelect
SelectContainer options=[multiselectoptions} ...

packages/react-ui-kit/src/Misc/useOutsideClick.ts
aweiss-dev: Shouldn't our tree-shaking just use the singular function we use from this package and remove the rest of the code?

packages/react-ui-kit/src/Form/SelectContainer.tsx
```diff
@@ -0,0 +1,290 @@
...
+  wrapperCSS?: CSSObject;
+  isMultiSelect?: boolean;
+  selectedLabel?: string;
+  selectedValue?: string;
+  checkIsSelected?: (optionValue: string | number) => boolean;
+}
+
+const filterSelectProps = props => filterProps(props, ['markInvalid']);
+
+const placeholderText = '- Please select -';
```
tlebon, line 59: this should be translated

packages/react-ui-kit/src/Form/Checkbox.tsx
```diff
@@ -57,6 +58,9 @@ const StyledLabel = (props: StyledLabelProps) => {
           [`.${INPUT_CLASSNAME}:hover + &::before`]: {
             borderColor: theme.general.primaryColor,
           },
+          [`.${INPUT_CLASSNAME}.hover + &::before`]: {
```
arjita-mitra, line 61: line number 58 and 61 are very similar. why do we need line 61? @przemvs

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -19,240 +19,238 @@
...
-  label?: string;
-  disabled?: boolean;
-  required?: boolean;
-  markInvalid?: boolean;
-  error?: ReactElement;
-  wrapperCSS?: CSSObject;
-}
+// eslint-disable-next-line react/display-name
+const CustomOption = (dataUieName: string) => (props: OptionProps<Option>) => {
+  const {children, data, isFocused, isMulti, isSelected} = props;
```
tlebon, line 144: isMulti param survives again

tlebon: now i finally understand why we are using react-select 😅

packages/react-ui-kit/src/Form/Checkbox.tsx
```diff
@@ -123,8 +123,10 @@ export const Checkbox: React.FC<CheckboxProps<HTMLInputElement>> = React.forward
         outline: `1px solid ${theme.general.primaryColor}`,
         outlineOffset: '0.4rem',
       },
+      ...wrapperCSS,
     })}
     style={style}
+    onClick={event => event.stopPropagation()}
```
tlebon, line 129: could this impact performance in other locations the checkbox is used?

packages/react-ui-kit/src/Form/Select.tsx
tlebon: is this still 2022?

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -19,239 +19,267 @@
...
+  valueContainer: provided => ({
+    ...provided,
+    padding: 0,
+    width: '100%',
+    display: 'grid',
+  }),
 });
 
-const filterSelectProps = props => filterProps(props, ['markInvalid']);
+const DropdownIndicator = props => {
```
tlebon, line 132: feels like this should be a separate file/component?

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -264,82 +292,30 @@ export const Select = <T extends SelectOption = SelectOption>({
...
-      </div>
+      <ReactSelect
+        id={id}
+        styles={customStyles(theme as Theme, markInvalid) as StylesConfig}
+        components={{
+          DropdownIndicator,
+          Option: CustomOption(dataUieName),
+          Menu: Menu(dataUieName),
+          ValueContainer,
+          IndicatorsContainer,
```
tlebon, line 310: in general i think all the components should live in separate files.

packages/react-ui-kit/src/Layout/Theme.tsx
```diff
@@ -142,6 +144,7 @@ export const themes: {[themeId in THEME_ID]: Theme} = {
       disabledColor: COLOR_V2.GRAY_60,
       contrastTextColor: COLOR.BLACK,
       borderColor: COLOR_V2.GRAY_90,
+      descriptionColor: COLOR_V2.WHITE,
```
arjita-mitra, line 147: just wanted to confirm that descriptionColor is white(descriptionColor: COLOR_V2.WHITE) for both light and dark theme?
@przemvs

arjita-mitra (changes requested):

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -19,240 +19,50 @@
...
   description?: string;
+  isDisabled?: boolean;
 };
 
-export interface SelectProps<T extends SelectOption = SelectOption> {
+interface CustomSelectProps extends StateManagerProps {
   id: string;
-  onChange: (selectedOption: T['value']) => void;
   dataUieName: string;
-  options: T[];
```
tlebon, line 39: i know StateManager adds a lot of these props but my concern is that we lose some readability of our API typing if we don't also define options, onChange, value etc. 
Especially since our options now has a specific type expected options: Option[]

WDYT?

aweiss-dev: Are these optional values in the state manager? If yes they shouldn't be in there and instead be typed at the component where they are needed.
Not a big fan of state managers that include everything eventually. I prefer strongly typed components where I can see what they need directly.

tlebon: that would be my thought as well, it makes our components unreadable to anyone who isn't fully versed on the state manager. (in this case the state manager is coming from react-select). 
As this in a component library we should tell people what params our components can use.

tlebon (changes requested):

PatrykBuniX (approved):

V-Gira (approved):

tlebon (approved):

PR #4328, opened 2022-07-14
add unit test for checkbox

arjita-mitra:
Introduce react testing library for unit testing.

- write test case for checkbox check/uncheck


packages/react-ui-kit/src/Form/Checkbox.test.tsx
```diff
@@ -42,3 +45,15 @@ describe('"Checkbox"', () => {
 describe('"CheckboxLabel"', () => {
   it('renders', () => matchComponent(<CheckboxLabel>Label</CheckboxLabel>));
 });
+
+test('account creation terms and condition checkbox is checked/unchecked', () => {
+  const {getByTestId} = render(
+    <StyledApp themeId={THEME_ID.LIGHT}>
+      <Checkbox id="1" data-testid="do-terms"></Checkbox>
+    </StyledApp>,
```
tlebon, line 53: do you want to move this to a custom render function here or in a follow-up PR?

tlebon: if you prefer merging this, its fine for me, lets just make a ticket or something to track this custom render function for using testing library for ui-kit. Perhaps in some places it requires some other provider also (translation? maybe not/something else)

arjita-mitra: @tlebon 

I would like to merge this and add a TO-DO with the ticket number in our code. Do you know under which project should I create the ticket?

arjita-mitra: Created a ticket https://wearezeta.atlassian.net/browse/SQSERVICES-1672

PR #4339, opened 2022-07-21
feat(core): prepare conversationService addUsers() for MLS and Proteus Versions, clean up code, fix scripts (FS-815)

aweiss-dev:
BREAKING CHANGE: 

- The addUser() function in ConversationService.ts has been renamed to addUsers() and its API changed in preparation of the MLS and Proteus split.


tlebon (approved):
LGTM

PR #4349, opened 2022-07-27
feat(api-client): add protocol to conversation

aweiss-dev:
feat(api-client): add protocol to conversation

### DIFF
--- a/packages/api-client/src/conversation/Conversation.ts
+++ b/packages/api-client/src/conversation/Conversation.ts
@@ -17,7 +17,7 @@
  *
  */
 
-import type {ConversationMembers} from './';
+import type {ConversationMembers, ConversationProtocol} from './';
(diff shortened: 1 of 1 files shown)


packages/api-client/src/conversation/Conversation.ts
```diff
@@ -91,4 +91,6 @@ export interface Conversation {
 
   /** MLS conversations only */
   group_id?: string;
+
+  protocol: ConversationProtocol;
```
tlebon, line 95: should this be an optional param? wondering about backwards compatability

aweiss-dev: Good catch. But the backend is sending this value already on the production server. So it is not optional anymore.

tlebon: ok then

PR #4353, opened 2022-07-29
fix(react-ui-kit): select types

przemvs:
## Pull Request Checklist

### DIFF
--- a/packages/react-ui-kit/src/Form/Select.tsx
+++ b/packages/react-ui-kit/src/Form/Select.tsx
@@ -31,16 +31,17 @@ import type {Theme} from '../Layout';
 import InputLabel from './InputLabel';
 
 export type Option = {
-  value: string | number;
+  value: string;
   label: string;
   description?: string;
   isDisabled?: boolean;
 };
 
-interface SelectProps<IsMulti extends boolean> extends StateManagerProps<Option, IsMulti> {
+interface SelectProps<IsMulti extends boolean> extends Omit<StateManagerProps<Option, IsMulti>, 'onChange'> {
   id: string;
   dataUieName: string;
   options: Option[];
+  onChange: (value: Option) => void;
   wrapperCSS?: CSSObject;
   label?: string;
   helperText?: string;
@@ -58,6 +59,7 @@ export const Select = <IsMulti extends boolean = false>({
   dataUieName,
   options,
   isMulti,
+  onChange,
   wrapperCSS = {},
   markInvalid = false,
   required = false,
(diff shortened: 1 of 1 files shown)


aweiss-dev (approved):

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -31,16 +31,17 @@ import type {Theme} from '../Layout';
...
   description?: string;
   isDisabled?: boolean;
 };
 
-interface SelectProps<IsMulti extends boolean> extends StateManagerProps<Option, IsMulti> {
+interface SelectProps<IsMulti extends boolean> extends Omit<StateManagerProps<Option, IsMulti>, 'onChange'> {
   id: string;
   dataUieName: string;
   options: Option[];
+  onChange: (value: Option) => void;
```
PatrykBuniX, line 44: Isn't the value `Option[]` when select is `IsMulti`?

packages/react-ui-kit/src/Form/Select.tsx
```diff
@@ -31,16 +31,17 @@ import type {Theme} from '../Layout';
 import InputLabel from './InputLabel';
 
 export type Option = {
-  value: string | number;
+  value: string;
```
tlebon, line 34: the value *will* sometimes be a number

packages/react-ui-kit/src/Form/Select.tsx
tlebon: that is incorrect from my testing.. for multi select it passes all

packages/react-ui-kit/src/Form/Select.tsx
aweiss-dev: in what case do we return a number? when we talk about inputs the input value will always be a string

packages/react-ui-kit/src/Form/Select.tsx
aweiss-dev: that's what I assumed as well when talking to @przemvs

packages/react-ui-kit/src/Form/Select.tsx
tlebon: cipher suites need to be a number for the BE to work
Screenshot 2022-07-29 at 13 49 15
Screenshot 2022-07-29 at 13 49 38

tlebon (changes requested):

packages/react-ui-kit/src/Form/Select.tsx
aweiss-dev: But that gets done from the onChange handler right? The option will and the onChange on the select are supposed to work with strings

tlebon: sorry i missed this mesage but yea thats fine also

PR #4354, opened 2022-08-02
feat: add a link variant for use in the webapp

V-Gira:
#### Link

- Add a Link variant to use the UI-kit Link in the web-app
- Keep the previous option as the default not to require any change to team-management

#### Checkbox

- Remove hardcoded style for color and links from Checkbox component

#### Select

- Fix styles that were not being applied because of React-Select prebaked theme
- Restore the disabled prop to Select and apply the corresponding styles
- Apply a border radius to Input corresponding to webapp design
Peek 2022-08-09 15-25


aweiss-dev (approved):
Good Job. Love the theme approach

PR #4358, opened 2022-08-10
feat: move the theme in ui-kit and expose css variables

V-Gira:
#### What?

- Move the theme object from the webapp to the ui-kit
- Expose css variables in GlobalStyles so they can be accessible from team setting and removed from the webapp .less files

#### Why?

- It ables us to change components in the ui-kit wthout having to fiddle with .less files on the webapp
- It will allow team-setting to use the same theme as webapp


packages/react-ui-kit/src/GlobalCssVariables.tsx
```diff
@@ -0,0 +1,98 @@
+/*
+ * Wire
+ * Copyright (C) 2019 Wire Swiss GmbH
```
tlebon, line 3: 2022
```suggestion
 * Copyright (C) 2022 Wire Swiss GmbH
```

PR #4362, opened 2022-08-17
feat: periodically commit proposals (fs-690)

aweiss-dev:
Implemented the following logic:

1. Always commit proposals for an MLS Conversation before sending a new message
2. When receiving MLS-message-events, also retrieve pending proposals for the conversation
2a. No Proposals -> proceed as usual
2b. Proposals without delay -> continue with "case I."
2c. Proposals with delay -> continue with "case II."
3. At App start, check the DB for open proposal commits -> continue with "case III."

"case I.":
1. immediately commit pending proposals

"case II.":
1. store delayed proposal in the database
2. schedule task for the delayed proposal
3. execute the task after the delay, commit the proposal and delete the DB entry
Note: This works only if the application is not closed/doesn't go to sleep since we rely on using timeouts. Fallback is "case III."

"case III.":
1. check the database for uncommitted proposals
2. schedule tasks for each found entry
3. execute the task after the delay, commit the proposal and delete the DB entry
Note: Some tasks will be executed immediately because the timeframe lies in the past


packages/core/src/main/notification/NotificationService.ts
```diff
@@ -356,4 +370,82 @@ export class NotificationService extends EventEmitter {
...
+  /**
+   * ## MLS only ##
+   * Commit all pending proposals for a given groupId
+   *
+   * @param groupId groupId of the conversation
+   * @param skipDelete if true, do not delete the pending proposals from the database
+   */
+  public async commitPendingProposals({groupId, skipDelete = false}: CommitPendingProposalsParams) {
+    const coreCryptoClient = this.coreCryptoClientProvider();
+    if (coreCryptoClient) {
```
thisisamir98, line 415: better check for !coreCryptoClient and throw an error

packages/core/src/main/util/TaskScheduler/TaskScheduler.ts
```diff
@@ -0,0 +1,74 @@
...
+      await task();
+      delete activeTimeouts[key];
+    },
+    delay > 0 ? delay : 0,
+  );
+
+  // add the task to the list of active tasks
+  activeTimeouts[key] = timeout;
+
+  console.info('New scheduled task', key, 'to be executed at', execute);
```
thisisamir98, line 55: let's use a logger here, example https://github.com/wireapp/wire-web-packages/blob/main/packages/core/src/main/notification/NotificationService.ts#L67

packages/core/src/main/conversation/ConversationService/ConversationService.ts
```diff
@@ -1169,7 +1169,7 @@ export class ConversationService {
       qualified_users: undefined,
     });
     const {group_id: groupId, qualified_id: qualifiedId} = newConversation;
-    const groupIdDecodedFromBase64 = Decoder.fromBase64(groupId!).asBytes;
+    const groupIdDecodedFromBase64 = Decoder.fromBase64(`${groupId}`).asBytes;
```
tlebon, line 1172: why this? converting it to a string from a Uint8Array? seems like it should happen a different way.

packages/core/src/main/notification/NotificationService.ts
tlebon: this is a good question, should be a singleton no?

aweiss-dev: I agree. It should be done in a separate PR tho.

tlebon (approved):

PR #4374, opened 2022-09-01
feat(core): update keys periodically (FS-562)

PatrykBuniX:
[link to confluence use case](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/611910060/Use+case+periodic+update+of+key+material+MLS)

Each user periodically updates their key material for each MLS group they're in.

Goals:

(1)
Store the date (timestamp) of group creation / welcome message as last-kay-material:
 - [x] when creating a group - after conversation is successfully created in `createMLSConversation` (`ConversationService.ts`)
 - [x] when being added to a group (welcome message) - after user is successfully added in `sendExternalJoinProposal` (ConversationService.ts)

Dates are stored in indexedDB's `last_key_material_update_dates` table as objects ({groupId:string, previousUpdateDate: number})

(2)
 - [x] After 30 days (likely to change) since creating/being added to a group - renew keying material for each mls group - achieved with `TaskScheduler` - each time app is launched it gets the dates of last key update and schedules tasks to renew key material (3) for `> 30 days` old dates.

(3)
 - [x] For each group calculate the key material update (core crypto `updateKeyingMaterial(groupId)`) - returns a commit
 - [x] Send commit message and accept it.
 - [x] Update last-key-material timestamp for each group and schedule new renewal task

(4)
- [x] Every time you (as a group admin) remove somebody else from the conversation, your keying material gets updated, so we should update current group's the timer (indexDB entry) with user-removal date.

(5)
- [x] threshold (`<timer>`) value should be configurable per environment to make it easy for QA to change the time after which keys should get updated (not to make them wait 30 days 😆) - we're passing threshold value (`number` - milliseconds) value via `mlsConfig` object using optional `FEATURE_MLS_CONFIG_KEYING_MATERIAL_UPDATE_THRESHOLD` env variable.

This pull request also adds new `MLSService` class which contains mls-related methods shared between conversation and notification services. This is the only service that has access to coreCryptoClient (which is lazy-initialised and possible `undefined` - proteus case). All other services should call coreCrypto methods indirectly through `MLSService` - this way we're sure coreCryptoProvider exists (an error would be thrown otherwise).

## Pull Request Checklist

### DIFF
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -51,6 +51,7 @@ import type {CoreCrypto} from '@otak/core-crypto';
 import {WEBSOCKET_STATE} from '@wireapp/api-client/src/tcp/ReconnectingWebsocket';
 import {createCustomEncryptedStore, createEncryptedStore} from './util/encryptedStore';
 import {Encoder} from 'bazinga64';
+import {MLSService} from './mls/MLSService/MLSService';
 
 export type ProcessedEventPayload = HandledEventPayload;
 
@@ -243,6 +244,9 @@ export class Account<T = any> extends EventEmitter {
       await this.initClient({clientType});
       // initialize schedulers for pending mls proposals once client is initialized
       await this.service?.notification.checkExistingPendingProposals();
+
+      // initialize schedulers for renewing key materials
+      await this.service?.notification.checkForKeyMaterialsUpdate();
     }
     return context;
   }
@@ -349,12 +353,14 @@ export class Account<T = any> extends EventEmitter {
     });
 
     const clientService = new ClientService(this.apiClient, this.storeEngine, cryptographyService);
+    const mlsService = new MLSService(this.apiClient, () => this.coreCryptoClient!);
     const connectionService = new ConnectionService(this.apiClient);
     const giphyService = new GiphyService(this.apiClient);
     const linkPreviewService = new LinkPreviewService(assetService);
     const notificationService = new NotificationService(
       this.apiClient,
       cryptographyService,
+      mlsService,
       this.storeEngine,
       () => this.coreCryptoClient,
     );
@@ -365,8 +371,9 @@ export class Account<T = any> extends EventEmitter {
         // We can use qualified ids to send messages as long as the backend supports federated endpoints
         useQualifiedIds: this.backendFeatures.federationEndpoints,
       },
-      () => this.coreCryptoClient!,
       notificationService,
+      mlsService,
+      () => this.coreCryptoClient!,
     );
 
     const selfService = new SelfService(this.apiClient);
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -51,7 +51,8 @@ import type {CoreCrypto} from '@otak/core-crypto';
 import {WEBSOCKET_STATE} from '@wireapp/api-client/src/tcp/ReconnectingWebsocket';
 import {createCustomEncryptedStore, createEncryptedStore} from './util/encryptedStore';
 import {Encoder} from 'bazinga64';
-import {MLSService} from './mls/MLSService/MLSService';
+import type {MLSService} from './mls';
+import type {MLSConfig} from './mls/types';
 
 export type ProcessedEventPayload = HandledEventPayload;
 
@@ -98,30 +99,6 @@ export interface Account {
 }
 
 export type CreateStoreFn = (storeName: string, context: Context) => undefined | Promise<CRUDEngine | undefined>;
-type SecretCrypto<T> = {
-  encrypt: (value: Uint8Array) => Promise<T>;
-  decrypt: (payload: T) => Promise<Uint8Array>;
-};
-
-export interface MLSConfig<T = any> {
-  /**
-   * encrypt/decrypt function pair that will be called before storing/fetching secrets in the secrets database.
-   * If not provided will use the built in encryption mechanism
-   */
-  secretsCrypto?: SecretCrypto<T>;
-
-  /**
-   * path on the public server to the core crypto wasm file.
-   * This file will be downloaded lazily when corecrypto is needed.
-   * It, thus, needs to know where, on the server, the file can be found
-   */
-  coreCrypoWasmFilePath: string;
-
-  /**
-   * (milliseconds) period of time between automatic updates of the keying material (30 days by default)
-   */
-  keyingMaterialUpdateThreshold?: number;
-}
 
 interface AccountOptions<T> {
   /** Used to store info in the database (will create a inMemory engine if returns undefined) */
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -103,7 +103,7 @@ type SecretCrypto<T> = {
   decrypt: (payload: T) => Promise<Uint8Array>;
 };
 
-interface MLSConfig<T = any> {
+export interface MLSConfig<T = any> {
   /**
    * encrypt/decrypt function pair that will be called before storing/fetching secrets in the secrets database.
    * If not provided will use the built in encryption mechanism
@@ -116,6 +116,11 @@ interface MLSConfig<T = any> {
    * It, thus, needs to know where, on the server, the file can be found
    */
   coreCrypoWasmFilePath: string;
+
+  /**
+   * (milliseconds) period of time between automatic updates of the keying material (30 days by default)
+   */
+  keyingMaterialUpdateThreshold: number;
 }
 
 interface AccountOptions<T> {
@@ -353,7 +358,7 @@ export class Account<T = any> extends EventEmitter {
     });
 
     const clientService = new ClientService(this.apiClient, this.storeEngine, cryptographyService);
-    const mlsService = new MLSService(this.apiClient, () => this.coreCryptoClient);
+    const mlsService = new MLSService(this.mlsConfig, this.apiClient, () => this.coreCryptoClient);
     const connectionService = new ConnectionService(this.apiClient);
     const giphyService = new GiphyService(this.apiClient);
     const linkPreviewService = new LinkPreviewService(assetService);
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -51,7 +51,7 @@ import type {CoreCrypto} from '@otak/core-crypto';
 import {WEBSOCKET_STATE} from '@wireapp/api-client/src/tcp/ReconnectingWebsocket';
 import {createCustomEncryptedStore, createEncryptedStore} from './util/encryptedStore';
 import {Encoder} from 'bazinga64';
-import type {MLSService} from './mls';
+import {MLSService} from './mls';
 import type {MLSConfig} from './mls/types';
 
 export type ProcessedEventPayload = HandledEventPayload;
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -362,7 +362,6 @@ export class Account<T = any> extends EventEmitter {
       cryptographyService,
       mlsService,
       this.storeEngine,
-      () => this.coreCryptoClient,
     );
     const conversationService = new ConversationService(
       this.apiClient,
@@ -373,7 +372,6 @@ export class Account<T = any> extends EventEmitter {
       },
       notificationService,
       mlsService,
-      () => this.coreCryptoClient!,
     );
 
     const selfService = new SelfService(this.apiClient);
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -353,7 +353,7 @@ export class Account<T = any> extends EventEmitter {
     });
 
     const clientService = new ClientService(this.apiClient, this.storeEngine, cryptographyService);
-    const mlsService = new MLSService(this.apiClient, () => this.coreCryptoClient!);
+    const mlsService = new MLSService(this.apiClient, () => this.coreCryptoClient);
     const connectionService = new ConnectionService(this.apiClient);
     const giphyService = new GiphyService(this.apiClient);
     const linkPreviewService = new LinkPreviewService(assetService);
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -120,7 +120,7 @@ export interface MLSConfig<T = any> {
   /**
    * (milliseconds) period of time between automatic updates of the keying material (30 days by default)
    */
-  keyingMaterialUpdateThreshold: number;
+  keyingMaterialUpdateThreshold?: number;
 }
 
 interface AccountOptions<T> {
--- a/packages/core/src/main/conversation/ConversationService/ConversationService.ts
+++ b/packages/core/src/main/conversation/ConversationService/ConversationService.ts
@@ -1262,6 +1262,10 @@ export class ConversationService {
 
     const messageResponse = await this.mlsService.uploadCoreCryptoCommitBundle(groupIdDecodedFromBase64, commitBundle);
 
+    //key material gets updated after removing a user from the group, so we can reset last key update time value in the store
+    const userRemovalTime = new Date().getTime();
+    await this.notificationService.storeLastKeyMaterialUpdateDate({groupId, previousUpdateDate: userRemovalTime});
+
     const conversation = await this.getConversations(conversationId.id);
 
     return {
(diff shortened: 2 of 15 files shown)


packages/core/src/main/conversation/ConversationService/ConversationService.ts
```diff
@@ -100,19 +100,12 @@ import {
...
+import {Decoder} from 'bazinga64';
 import {mapQualifiedUserClientIdsToFullyQualifiedClientIds} from '../../util/mapQualifiedUserClientIdsToFullyQualifiedClientIds';
-import {CommitBundle} from '@otak/core-crypto/platforms/web/corecrypto';
-
-//@todo: this function is temporary, we wait for the update from core-crypto side
-//they are returning regular array instead of Uint8Array for commit and welcome messages
-const optionalToUint8Array = (array: Uint8Array | []): Uint8Array => {
-  return Array.isArray(array) ? Uint8Array.from(array) : array;
-};
+import {MLSService} from '../../mls/MLSService/MLSService';
```
thisisamir98, line 108: should be `import type`

packages/core/src/main/cryptography/CryptographyDatabaseRepository.ts
```diff
@@ -27,6 +27,7 @@ export enum DatabaseStores {
   SESSIONS = 'sessions',
   GROUP_IDS = 'group_ids',
   PENDING_PROPOSALS = 'pending_proposals',
+  LAST_KEY_MATERIAL_UPDATE_DATES = 'last_key_material_update_dates',
```
thisisamir98, line 30: I think we should update webapp as well 🤔

PatrykBuniX: yeah, got it already updated locally

packages/core/src/main/util/TaskScheduler/TaskScheduler.ts
```diff
@@ -48,13 +51,26 @@ const addTask = ({task, firingDate, key}: ScheduleTaskParams) => {
...
-  const timeout = setTimeout(
-    async () => {
-      await task();
-      delete activeTimeouts[key];
-    },
-    delay > 0 ? delay : 0,
-  );
+  let timeout: ReturnType<typeof setTimeout>;
+
+  if (delay <= MAX_TIMEOUT_DELAY_VALUE) {
```
PatrykBuniX, line 56: When we try to pass delay higher than [max delay value (~28.4 days)](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value) the callback passed to `setTimeout` will execute immediately. To avoid it (but still support long delay times) in cases when delay time is too long for `setTimeout` we use `setInterval` instead and check every minute whether a task should be executed.

packages/core/src/main/notification/NotificationService.ts
```diff
@@ -453,4 +443,78 @@ export class NotificationService extends EventEmitter {
...
+  }
+
+  private createKeyMaterialUpdateTaskSchedulerId(groupId: string) {
+    return `renew_key_material-${groupId}`;
+  }
+
+  private async scheduleTaskToRenewKeyMaterial({groupId, previousUpdateDate}: LastKeyMaterialUpdateParams) {
+    const groupConversationExists = await this.mlsService.conversationExists(Decoder.fromBase64(groupId).asBytes);
+
+    if (!groupConversationExists) {
```
PatrykBuniX, line 486: If for some reason user is no longer a member of the group we store `last-key-material-update` date for, we need to remove its entry and don't schedule a renewal task.

packages/core/src/main/conversation/ConversationService/ConversationService.ts
```diff
@@ -1347,21 +1283,26 @@ export class ConversationService {
...
    */
   public async sendExternalJoinProposal(conversationGroupId: string, epoch: number) {
-    const coreCryptoClient = this.coreCryptoClientProvider();
     const groupIdDecodedFromBase64 = Decoder.fromBase64(conversationGroupId!).asBytes;
-    const externalProposal = await coreCryptoClient.newExternalProposal(ExternalProposalType.Add, {
+    const externalProposal = await this.mlsService.newExternalProposal(ExternalProposalType.Add, {
       epoch,
       conversationId: groupIdDecodedFromBase64,
     });
     await this.apiClient.api.conversation.postMlsMessage(
```
PatrykBuniX, line 1291: Two questions:
1. Does it mean that user was added to a conversation after message is successfully posted?
2. Don't we have to accept a commit on our side? 🤔

packages/core/src/main/mls/MLSService/MLSService.ts
```diff
@@ -0,0 +1,177 @@
...
+import {Decoder, Encoder} from 'bazinga64';
+import type {MLSConfig} from '../types';
+
+//@todo: this function is temporary, we wait for the update from core-crypto side
+//they are returning regular array instead of Uint8Array for commit and welcome messages
+export const optionalToUint8Array = (array: Uint8Array | []): Uint8Array => {
+  return Array.isArray(array) ? Uint8Array.from(array) : array;
+};
+
+export class MLSService {
```
PatrykBuniX, line 46: The reason I've added MLSService is I wanted to have only one place that has access to coreCryptoClient which is possibly undefined (proteus protocol), so using client through this service would be safe (otherwise error is thrown). This would be good place to store some mls logic shared between different services (`ConversationService` and `NotificationService` for now). Let me know what you think. More context in conversation with @thisisamir98:

Screenshot 2022-09-07 at 12 44 04

packages/core/src/main/conversation/ConversationService/ConversationService.ts
```diff
@@ -1347,21 +1283,26 @@ export class ConversationService {
...
       //@todo: it's temporary - we wait for core-crypto fix to return the actual Uint8Array instead of regular array
       optionalToUint8Array(externalProposal),
     );
+
+    //We store the info when user was added (and key material was created), so we will know when to renew it
+    const userAddedTimeStamp = new Date().getTime();
+    await this.notificationService.storeLastKeyMaterialUpdateDate({
```
arjita-mitra, line 1298: @PatrykBuniX  Can we make it a function and pass the timestamp as its repeated above at line number 1179?


await this.notificationService.storeLastKeyMaterialUpdateDate({
      previousUpdateDate: groupCreationTimeStamp,
      groupId,
    });

arjita-mitra (changes requested):

packages/core/src/main/mls/MLSService/MLSService.ts
atomrc: Thanks for the explanation, and I think it's a solid reason to do so. Very good call 👌

packages/core/src/main/mls/MLSService/MLSService.ts
```diff
@@ -0,0 +1,177 @@
...
+        //@todo: it's temporary - we wait for core-crypto fix to return the actual Uint8Array instead of regular array
+        optionalToUint8Array(commitBundle.commit),
+      );
+      await coreCryptoClient.commitAccepted(groupIdDecodedFromBase64);
+      return messageResponse;
+    }
+    return null;
+  }
+
+  public async addUsersToExistingMLSConversation(groupIdDecodedFromBase64: Uint8Array, invitee: Invitee[]) {
```
atomrc, line 79: I think, now that it's in `mls` scope, we can remove `MLS` from method names
```suggestion
  public async addUsersToExistingConversation(groupIdDecodedFromBase64: Uint8Array, invitee: Invitee[]) {
```

packages/core/src/main/mls/MLSService/MLSService.ts
```diff
@@ -0,0 +1,177 @@
...
+        //@todo: it's temporary - we wait for core-crypto fix to return the actual Uint8Array instead of regular array
+        optionalToUint8Array(memberAddedMessages.commit),
+      );
+      await coreCryptoClient.commitAccepted(groupIdDecodedFromBase64);
+      return messageResponse;
+    }
+    return null;
+  }
+
+  public async getCoreCryptoKeyPackagesPayload(qualifiedUsers: QualifiedUsers[]) {
```
atomrc, line 98:
```suggestion
  public async getKeyPackagesPayload(qualifiedUsers: QualifiedUsers[]) {
```

No need to mention CoreCrypto here

packages/core/src/main/mls/MLSService/MLSService.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  private getCoreCryptoClient() {
+    const client = this.coreCryptoClientProvider();
+    if (!client) {
+      throw new Error('Could not get coreCryptoClient');
+    }
+    return client;
+  }
+
+  public async uploadCoreCryptoCommitBundle(groupIdDecodedFromBase64: Uint8Array, commitBundle: CommitBundle) {
```
atomrc, line 61:
```suggestion
  public async uploadCommitBundle(groupIdDecodedFromBase64: Uint8Array, commitBundle: CommitBundle) {
```

packages/core/src/main/Account.ts
```diff
@@ -349,14 +335,15 @@ export class Account<T = any> extends EventEmitter {
     });
 
     const clientService = new ClientService(this.apiClient, this.storeEngine, cryptographyService);
+    const mlsService = new MLSService(this.mlsConfig, this.apiClient, () => this.coreCryptoClient);
```
atomrc, line 338: Maybe (in a follow up PR, not this one) we could move **all** code that uses coreCrypto to the MLS scope :)

packages/core/src/main/conversation/ConversationService/ConversationService.ts
```diff
@@ -1325,12 +1257,16 @@ export class ConversationService {
...
     );
 
-    const commitBundle = await coreCryptoClient.removeClientsFromConversation(
+    const commitBundle = await this.mlsService.removeClientsFromConversation(
       groupIdDecodedFromBase64,
       fullyQualifiedClientIds,
     );
 
-    const messageResponse = await this.sendCommitBundleRemovalMessages(groupIdDecodedFromBase64, commitBundle);
+    const messageResponse = await this.mlsService.uploadCoreCryptoCommitBundle(groupIdDecodedFromBase64, commitBundle);
```
atomrc, line 1265: Let's merge those 2 methods in a single one and make  the other private so that it doesn't leak the coreCrypto bundle and this stays internal implementation details :)

packages/core/src/main/util/TaskScheduler/TaskScheduler.ts
atomrc: How about we always use `setInterval` and try to avoid 2 different handling (that could create unexpected bugs when we reach a certain delay). 

I'd suggest you write a unit test that goes with, that would be a nice exercise

atomrc (changes requested):
That's a solid PR, there are a few optimization that we could do though :)

packages/core/src/main/util/TaskScheduler/TaskScheduler.ts
PatrykBuniX: Setting just one interval sounds good to me, but then I can't use TaskScheduler Adrian has created for committing pending proposals, because it has to use `setTimeout` for accuracy - what's also fine for me 👍 .

packages/core/src/main/Account.ts
PatrykBuniX: Yeah, this is what we aim for! 😄

packages/core/src/main/conversation/ConversationService/ConversationService.ts
PatrykBuniX: I've done it for both: `updateKeyingMaterial` and `removeClientsFromConversation` methods 👍

packages/core/src/main/conversation/ConversationService/ConversationService.ts
PatrykBuniX: Sure, why not! 😄

packages/core/src/main/util/TaskScheduler/TaskScheduler.ts
atomrc: Oh right, that makes sense. Maybe we could have 2 classes then:
- HighPrecisionTaskScheduler => more precise but with the performance issue that it create 1 timeout per task
- TaskScheduler => less precise but one has one single interval running

WDYT?

PatrykBuniX: Yeah, makes sense to me, both would be very useful, we'll decide which one to choose based how precisely we want the task to be executed 👍

PatrykBuniX: After digging into the code: Creating two task schedulers makes sense, but since we want to create single `setInterval` we won't be able to use it in our case. TaskScheduler is responsible for running a callback at specified `firingDate`. The thing is in this case we don't have single firingDate, we have many of them (one for each group), so there's no way to specify date at witch it should fire. The simplest solution would be just running single interval on client's init. What do you think?

packages/core/src/main/mls/index.ts
```diff
@@ -0,0 +1,20 @@
+/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
```
thisisamir98, line 3:
```suggestion
 * Copyright (C) 2022 Wire Swiss GmbH
```

packages/core/src/main/util/LowPrecisionTaskScheduler/LowPrecisionTaskScheduler.test.node.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    LowPrecisionTaskScheduler.addTask({
+      key: 'test-key',
+      firingDate: 0,
+      intervalDelay: 1000,
+      task: mockedTask,
+    });
+
+    jasmine.clock().tick(1001);
+    await Promise.resolve();
```
PatrykBuniX, line 61: Can't find other way to make scheduled async task execute before `expect` check... Any ideas? @atomrc 😄

packages/core/src/main/util/LowPrecisionTaskScheduler/LowPrecisionTaskScheduler.ts
```diff
@@ -0,0 +1,87 @@
...
+const addTask = ({key, firingDate, task, intervalDelay}: ScheduleLowPrecisionTaskParams) => {
+  const existingIntervalId = intervals[intervalDelay]?.timeoutId;
+  if (existingIntervalId) {
+    clearInterval(existingIntervalId);
+  }
+
+  const tasks = intervals[intervalDelay]?.tasks || [];
+  tasks.push({key, firingDate, task});
+
+  const timeoutId = setInterval(async () => {
```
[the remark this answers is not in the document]
atomrc: side note: then you cannot set an interval value for each task, it's going to be a general value of the module.

one (to do later maybe) optimization could be that the module adapts the interval depending on the tasks firing date. 
- if a task is added that need to fire in a few seconds we can set an interval with a min order of magnitude
- if a task is added that need to be fired in days, then we can adapt the timer frequency to fire only every hours
- we can do some kind of mean value of the firing dates of the tasks and kill/restart an interval if the value changes

But probably not for now.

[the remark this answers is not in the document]
atomrc: > ooohhh I missed that. Then is sounds perfect actually 💪

packages/core/src/main/util/LowPrecisionTaskScheduler/LowPrecisionTaskScheduler.test.node.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    LowPrecisionTaskScheduler.cancelTask({
+      intervalDelay: 4000,
+      key: 'test3-key',
+    });
+
+    jasmine.clock().tick(4001);
+
+    await Promise.resolve();
+    await Promise.resolve();
```
atomrc, line 123: genuine question: do you really need 2 of those here? I would think that 1 is enough (since the other micro task have priority over this one)? Can you give it a try?

atomrc (approved):
That's quality PR 💪 
Great job Patryk!!!

packages/core/src/main/util/LowPrecisionTaskScheduler/LowPrecisionTaskScheduler.test.node.ts
PatrykBuniX: Yeah, for some reason it doesn't work with only one of it for case when there are at least 2 tasks scheduled 🤔

arjita-mitra (approved):

atomrc (approved):

PR #4378, opened 2022-09-06
feat: Also create MLS device when logging in with old proteus device

atomrc:
Will allow a user to also create an mls device when logging in with a device that was previously proteus-only


PatrykBuniX (approved):
Straightforward 👍

thisisamir98 (approved):

V-Gira (approved):

PR #4379, opened 2022-09-07
feat: Update getAllParticipantsClients to use better endpoints without a hack

thisisamir98:
This PR will change getAllParticipantsClients, basically previously we used to send an empty message to backend to get a list of missing clients as a hacky way to get a list of user clients in a conversation (**BROKEN FOR MLS CONVERSATIONS**), in this PR the logic is slightly changed without a breaking change, what is done is instead of calling an endpoint with empty message we first fetch the conversation from the backend and extract a fresh list of users in it then make a request to another endpoint we have which if we give it a list of user ids it'll return a list of clients.

## Pull Request Checklist

- [X] My code is covered by tests
- [X] I will [merge the PR as breaking change](https://github.com/wireapp/wire-web-packages/wiki/Releases#create-major-release), if the API contract changes

### DIFF
--- a/packages/core/src/main/conversation/ConversationService/ConversationService.ts
+++ b/packages/core/src/main/conversation/ConversationService/ConversationService.ts
@@ -144,20 +144,17 @@ export class ConversationService {
     return genericMessage;
   }
 
-  private async getConversationQualifiedMembers(conversationId: QualifiedId): Promise<QualifiedId[]> {
+  private async getConversationQualifiedMembers(conversationId: string | QualifiedId): Promise<QualifiedId[]> {
     const conversation = await this.apiClient.api.conversation.getConversation(conversationId);
     /*
      * If you are sending a message to a conversation, you have to include
      * yourself in the list of users if you want to sync a message also to your
      * other clients.
      */
-    return (
-      conversation.members.others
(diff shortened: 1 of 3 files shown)


packages/core/src/main/conversation/ConversationService/ConversationService.ts
```diff
@@ -144,20 +144,17 @@ export class ConversationService {
...
-      conversation.members.others
-        .filter(member => !!member.qualified_id)
-        .map(member => member.qualified_id!)
-        // TODO(Federation): Use 'domain' from 'conversation.members.self' when backend has it implemented
-        .concat({domain: this.apiClient.context!.domain!, id: conversation.members.self.id})
-    );
+    return conversation.members.others
+      .filter(member => !!member.qualified_id)
+      .map(member => member.qualified_id!)
+      .concat(conversation.members.self.qualified_id!);
```
atomrc, line 157: Are we absolutely sure backend now implements this everywhere?

thisisamir98: I tested on staging.zinfra.io and it works

atomrc (approved):
Solid PR 👍

PR #4381, opened 2022-09-08
feat: add promise queue package

tlebon:
This adds a PromiseQueue util that can be used in the core to queue sending of messages


packages/promise-queue/jest.config.js
```diff
@@ -0,0 +1,11 @@
+/*
```
tlebon, line 1: the babel config has a copyright and this doesn't. not sure if either of these type of files need one.

tlebon:
fine for me but i cant approve really.

packages/promise-queue/src/PromiseQueue.ts
```diff
@@ -0,0 +1,152 @@
...
+  paused: false,
+};
+
+export class PromiseQueue {
+  private blocked: boolean;
+  private runningTasks: number;
+  private interval?: number;
+  private paused: boolean;
+  private readonly concurrent: number;
+  private readonly logger?: {warn: (...args: any[]) => void};
```
tlebon, line 35: just curious if we should include a default logger option as well as allowing the consumer to set one.

packages/promise-queue/webpack.config.js
```diff
@@ -0,0 +1,37 @@
...
+    [`${projectName}.test`]: `${__dirname}/src/test/browser/index.js`,
+  },
+  mode: 'production',
+  output: {
+    filename: '[name].bundle.js',
+    library: 'PromiseQueue',
+    path: `${__dirname}/dist`,
+    publicPath: '/',
+  },
+};
```
atomrc, line 37: Oh, we don't need this :D

packages/promise-queue/src/PromiseQueue.ts
atomrc: I'd like our libraries to be rather agnostic from any particular logger. 
Initially it pulled the entire `logdown` lib for logging, but I don't think it makes sense. If at some point we migrate the webapp's logger, I would like the change to reflect on all the libraries we use. 
So I think injecting the logger makes a lot of sense and let the webapp completely control how things are logged and with what namespace

PR #4382, opened 2022-09-08
feat: Always send OTR messages using protobuf

atomrc:
using JSON to send proteus message payload is deprecated and protobuf is the new way to do thing. 

Protobuf has been battle tested with the federation env, we can now get rid of the old JSON API and remove some dead code

BREAKING CHANGE: the `sendAsProtobuf` option for sending message has disappeared. All messages will be sent using protobuf from now one


atomrc:
Superseded by https://github.com/wireapp/wire-web-packages/pull/4927

PR #4383, opened 2022-09-12
runfix(core): address dexie error on deleting all stores on login 

PatrykBuniX:
Client was not able to delete all stores on login from a fresh (deleted) device.

`Object.keys` (eg. `"AMPLIFY"`) was used instead of `Object.values` (eg. `"amplify"`) for table names, so the client was not able to find the tables and showed `unexpected error` message on client's side.

### DIFF
--- a/packages/core/src/main/cryptography/CryptographyDatabaseRepository.ts
+++ b/packages/core/src/main/cryptography/CryptographyDatabaseRepository.ts
@@ -37,9 +37,7 @@ export class CryptographyDatabaseRepository {
 
   public deleteStores(): Promise<boolean[]> {
     return Promise.all(
-      (
-        Object.keys(CryptographyDatabaseRepository.STORES) as (keyof typeof CryptographyDatabaseRepository.STORES)[]
-      ).map(store => this.storeEngine.deleteAll(store)),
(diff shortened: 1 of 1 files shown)


packages/core/src/main/cryptography/CryptographyDatabaseRepository.ts
```diff
@@ -37,9 +37,7 @@ export class CryptographyDatabaseRepository {
 
   public deleteStores(): Promise<boolean[]> {
     return Promise.all(
-      (
-        Object.keys(CryptographyDatabaseRepository.STORES) as (keyof typeof CryptographyDatabaseRepository.STORES)[]
-      ).map(store => this.storeEngine.deleteAll(store)),
+      Object.values(CryptographyDatabaseRepository.STORES).map(store => this.storeEngine.deleteAll(store)),
```
arjita-mitra, line 40: may be we can add a comment here about the table name mismatch to prevent such error in future?

atomrc (approved):
Nice catch 😎

packages/core/src/main/cryptography/CryptographyDatabaseRepository.ts
PatrykBuniX: added a comment 👍

PR #4384, opened 2022-09-12
feat(core): sync keys with backend (FS-655)

PatrykBuniX:
## Periodically sync key packages with backend.

[link to use case doc on confluence](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/556499553/Use+case+refill+key+packages+MLS)

We want to make sure that client always has a pack (constant `"n"` value - 100 by default) of valid key packages on backend. 

Goals:
- [x] `[1]` - Check if backend can be queried for key-packages count
   We allow the client to query the backend (step `[2]`) if following conditions are truthy:
   1. the client did not query the backend for the last 24h, and
   2. `clientValidKeypackagesCount` method call from core-crypto returns value lower or equal `"n"` / 2

- [x] `[2]` - Query backend if new key packages needs to be uploaded
    Query the backend with client's `clientId`:
    `this.apiClient.api.client.getMLSKeyPackageCount(clientId)`: [`GET: /mls/key-packages/self/{clientId}/count`](https://staging-nginz-https.zinfra.io/api/swagger-ui/#/default/get_mls_key_packages_self__client__count)
    If value returned from the api call is <= `"n"` / 2 -> upload new key packages to backend (step `[3]`)

- [x] `[3]` - Create and upload new key packages 
    1. create a new set of key packages - call core-crypto's `clientKeypackages("n")` method with `"n"` value (to reach requested amount)
    2. upload new key packages with client's `clientId`:
        `this.mlsService.uploadMLSKeyPackages(newKeyPackages, clientId)` [`POST: /mls/key-packages/self/{clientId}`](https://staging-nginz-https.zinfra.io/api/swagger-ui/#/default/post_mls_key_packages_self__client_)


packages/core/src/main/notification/NotificationService.ts
```diff
@@ -519,4 +521,62 @@ export class NotificationService extends EventEmitter {
...
+      const clientId = this.apiClient.validatedClientId;
+
+      //check numbers of keys on backend
+      const backendKeyPackagesCount = await this.apiClient.api.client.getMLSKeyPackageCount(clientId);
+
+      if (backendKeyPackagesCount <= minAllowedNumberOfKeyPackages) {
+        //upload new keys
+        const newKeyPackages = await this.mlsService.clientKeypackages(INITIAL_NUMBER_OF_KEY_PACKAGES);
+
+        await this.mlsService.uploadMLSKeyPackages(newKeyPackages, clientId);
```
PatrykBuniX, line 552: ❓ Is uploading the keys all we have to do? Don't we want to update the keys locally also, so we can use new keys right away in current client session? 🤔 Unless it's somehow done automatically under the hood 🤔

tlebon: do you need the keys locally? i think keys always just live on the backend and we call the backend to get them as needed

packages/core/src/main/notification/NotificationService.ts
```diff
@@ -55,6 +55,8 @@ enum TOPIC {
 
 const DEFAULT_KEYING_MATERIAL_UPDATE_THRESHOLD = 1000 * 60 * 60 * 24 * 30; //30 days
 
+const INITIAL_NUMBER_OF_KEY_PACKAGES = 100;
```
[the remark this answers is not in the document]
tlebon: i believe it is 100 for now.

packages/core/src/main/notification/NotificationService.ts
PatrykBuniX: hmmm, the task is about syncing, so my thought is it should sync the packages on both (client and backend) sides 🤔

atomrc: Humm yeah I think there is something wrong here 🤔 
Normally it should be coreCrypto that generates (and store) the new key material. There must be a method for this. 
And you should be requesting `INITIAL_NUMBER.. - validKeyPackageCount` I believe (we just want to generate the difference, not add another 100 to it, I believe)

packages/core/src/main/notification/NotificationService.ts
```diff
@@ -519,4 +521,62 @@ export class NotificationService extends EventEmitter {
       this.logger.error('Could not get last key material update dates', error);
     }
   }
+
+  private scheduleTaskToQueryKeyPackagesCountAndSyncWithBackend(firingDate: number) {
```
atomrc, line 525:
```suggestion
  private scheduleKeyPackagesSync(firingDate: number) {
```

packages/core/src/main/cryptography/CryptographyDatabaseRepository.ts
```diff
@@ -28,6 +28,7 @@ export enum DatabaseStores {
   GROUP_IDS = 'group_ids',
   PENDING_PROPOSALS = 'pending_proposals',
   LAST_KEY_MATERIAL_UPDATE_DATES = 'last_key_material_update_dates',
+  LAST_KEY_PACKAGE_QUERY_DATE = 'last_key_package_query_date',
```
atomrc, line 31: Let's maybe avoid creating countless indexed db tables, it's going to be hard to maintain/migrate in the future. 

If we are going to create a lot of entries like this, maybe we could use `localStorage` for this. It's more light weight and it won't hold application critical informations. 
WDYT?

packages/core/src/main/notification/NotificationService.ts
PatrykBuniX: My understanding was that they handle it on `core-cypto` side.

Screenshot 2022-09-12 at 16 57 40

atomrc: ohhh thanks for linking the doc, it then makes perfect sense 👌

packages/core/src/main/cryptography/CryptographyDatabaseRepository.ts
PatrykBuniX: Yeah, good idea!

atomrc (approved):
LGTM 😎

PR #4385, opened 2022-09-13
feat: preshared key (FS-564)

tlebon:
For the backend to remove users from a group, a public removal key must be included with the conversation upon its creation. This removal key is cipher specific(?). If it is added, users will automatically be removed by the BE when the conditions arise for them to be. (To my understanding)

## Pull Request Checklist


packages/core/src/main/conversation/ConversationService/ConversationService.ts
```diff
@@ -1148,7 +1148,15 @@ export class ConversationService {
       throw new Error('You need to pass self user qualified id in order to create an MLS conversation');
     }
 
-    await this.mlsService.createConversation(groupIdDecodedFromBase64);
+    const mlsKeys = (await this.apiClient.api.client.getPublicKeys()).removal;
+    const mlsKeyBytes = Object.values(mlsKeys).map((key: string) => Decoder.fromBase64(key).asBytes);
+    const config: ConversationConfiguration = {
+      externalSenders: mlsKeyBytes,
+      ciphersuite: 1,
```
atomrc, line 1155: I think we have an enum in core-crypto for ciphersuite identifiers, don't we?

atomrc (approved):
LGTM, please just make sure if there is not enum we could use to define the ciphersuite

packages/core/src/main/conversation/ConversationService/ConversationService.ts
tlebon: when i used the enum before it was sending back the name of the cipher and not the number, but i will check again now.

tlebon: Screenshot 2022-09-22 at 17 12 50
unfortunately the enum from CoreCrypto is currently undefined in value, while the config needs a numerical value.

atomrc: Can you try with the latest core-crypto (if you rebase to the latest main branch) maybe. I think there was a change around this in the 0.5.0 (in case you are not using this one of course)

tlebon: this has just been merged with the main branch

atomrc (approved):

packages/core/src/main/conversation/ConversationService/ConversationService.ts
tlebon: will follow up with core-crypto folks but we are blocked here for now i believe

tlebon: this is broken still, merging for now with a TODO not to update when its fixed

PR #4392, opened 2022-09-13
feat: Queue message sending [FS-882]

atomrc:
This will allow not sending messages before we have handled the complete notification stream and be sure that we are up to date with the current `epoch` of each and every conversation

### DIFF
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -53,6 +53,7 @@ import {createCustomEncryptedStore, createEncryptedStore, deleteEncryptedStore} 
 import {Encoder} from 'bazinga64';
 import {MLSService} from './mls';
 import type {MLSConfig} from './mls/types';
+import {resumeMessageSending} from './conversation/message/messageSender';
 
 export type ProcessedEventPayload = HandledEventPayload;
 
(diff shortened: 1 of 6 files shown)


packages/core/src/main/Account.ts
```diff
@@ -591,6 +592,9 @@ export class Account<T = any> extends EventEmitter {
       );
       // We can now unlock the websocket and let the new messages being handled and decrypted
       this.apiClient.transport.ws.unlock();
+      // We need to way for the notification stream to be fully handled before releasing the message sending queue.
```
tlebon, line 595:
```suggestion
      // We need to wait for the notification stream to be fully handled before releasing the message sending queue.
```

packages/core/src/main/conversation/message/messageSender.ts
```diff
@@ -0,0 +1,34 @@
+/*
+ * Wire
+ * Copyright (C) 2019 Wire Swiss GmbH
```
tlebon, line 3:
```suggestion
 * Copyright (C) 2022 Wire Swiss GmbH
```

PatrykBuniX (approved):
Looks good 👍

V-Gira (approved):

PR #4397, opened 2022-09-15
feat: Handle external proposals

atomrc:
Handles incoming external proposal messages

needs https://github.com/wireapp/wire-web-packages/pull/4398

### DIFF
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -52,7 +52,7 @@ import {WEBSOCKET_STATE} from '@wireapp/api-client/src/tcp/ReconnectingWebsocket
 import {createCustomEncryptedStore, createEncryptedStore, deleteEncryptedStore} from './util/encryptedStore';
 import {Encoder} from 'bazinga64';
 import {MLSService} from './mls';
-import type {MLSConfig} from './mls/types';
+import type {MLSCallbacks, MLSConfig} from './mls/types';
 import {resumeMessageSending} from './conversation/message/messageSender';
 
 export type ProcessedEventPayload = HandledEventPayload;
@@ -332,6 +332,10 @@ export class Account<T = any> extends EventEmitter {
     }
   }
 
+  configureMLSCallbacks(mlsCallbacks: MLSCallbacks) {
+    this.service?.mls.configureMLSCallbacks(mlsCallbacks);
+  }
+
   public async initServices(context: Context): Promise<void> {
     this.storeEngine = await this.initEngine(context);
     const accountService = new AccountService(this.apiClient);
@@ -401,7 +405,6 @@ export class Account<T = any> extends EventEmitter {
         this.mlsConfig,
         entropyData,
       );
-      this.coreCryptoClient.registerCallbacks({authorize: () => true, clientIdBelongsToOneOf: () => true});
     }
 
     return loadedClient;
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -401,6 +401,7 @@ export class Account<T = any> extends EventEmitter {
         this.mlsConfig,
         entropyData,
       );
+      this.coreCryptoClient.registerCallbacks({authorize: () => true, clientIdBelongsToOneOf: () => true});
     }
 
     return loadedClient;
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -332,6 +332,13 @@ export class Account<T = any> extends EventEmitter {
     }
   }
 
+  /**
+   * In order to be able to send MLS messages, the core needs a few information from the consumer.
+   * Namely:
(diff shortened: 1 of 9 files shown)


packages/core/src/main/mls/MLSService/MLSService.ts
```diff
@@ -59,23 +59,24 @@ export class MLSService {
...
-    if (commitBundle.commit) {
+    if (commit) {
       try {
         const messageResponse = await this.apiClient.api.conversation.postMlsMessage(
           //@todo: it's temporary - we wait for core-crypto fix to return the actual Uint8Array instead of regular array
-          optionalToUint8Array(commitBundle.commit),
+          optionalToUint8Array(commit),
         );
         await coreCryptoClient.commitAccepted(groupId);
+        if (welcome) {
```
atomrc, line 72: Commit messages need to be sent first only when it's successful can we send the welcome message

packages/core/src/main/notification/NotificationService.ts
```diff
@@ -257,25 +257,31 @@ export class NotificationService extends EventEmitter {
...
+        const groupId = await this.getGroupIdFromConversationId(
           event.qualified_conversation ?? {id: event.conversation, domain: ''},
         );
+        const groupIdBytes = Decoder.fromBase64(groupId).asBytes;
 
         // Check if the message includes proposals
-        const {proposals, commitDelay, message} = await this.mlsService.decryptMessage(groupId, encryptedData);
-        if (proposals.length > 0) {
+        const res = await this.mlsService.decryptMessage(groupIdBytes, encryptedData);
+        const {proposals, commitDelay, message} = res;
```
PatrykBuniX, line 274: Not a big thing, but why we can't destructure it right away? Is it for readability? I'd suggest some better naming for `res` then, like `decryptedMessage` 😄

PatrykBuniX (approved):

packages/core/src/main/notification/NotificationService.ts
atomrc: Oh totally right, this was for debugging only 😱 !!!! Nice catch 👌

packages/core/src/main/Account.ts
```diff
@@ -400,6 +400,7 @@ export class Account<T = any> extends EventEmitter {
         this.mlsConfig,
         entropyData,
       );
+      this.coreCryptoClient.registerCallbacks({authorize: () => true, clientIdBelongsToOneOf: () => true});
```
atomrc, line 403: TODO: implement correct checks (if the user is admin of the conversation).
The info needs to come from the consumer side, so we will need to add callbacks to the constructor of `Account`

packages/core/src/main/notification/NotificationService.ts
```diff
@@ -337,42 +340,21 @@ export class NotificationService extends EventEmitter {
...
 
-  /**
-   * ## MLS only ##
-   * If there is a groupId in the conversation, we need to store the conversationId => groupId pair
-   * in order to find the groupId when decrypting messages
-   * This is a bit hacky but since mls messages do not embed the groupId we need to keep a mapping of those
-   *
-   * @param conversation conversation with group_id
-   */
-  public async saveConversationGroupId(conversation: Conversation) {
```
atomrc, line 348: see https://github.com/wireapp/wire-web-packages/pull/4397/files#r974442782

PatrykBuniX (approved):

packages/core/src/main/notification/NotificationService.ts
```diff
@@ -258,25 +257,30 @@ export class NotificationService extends EventEmitter {
...
+        if (typeof commitDelay === 'number' || proposals.length > 0) {
+          // we are dealing with a proposal, add a task to process this proposal later on
+          // Those proposals are stored inside of coreCrypto and will be handled after a timeout
           await this.handlePendingProposals({
-            groupId: groupId.toString(),
+            groupId,
             delayInMs: commitDelay ?? 0,
             eventTime: event.time,
           });
+          // This is not a text message, there is nothing more to do
```
tlebon, line 282: what do you mean by this comment?

atomrc: since it's a proposal message, there is no content to the actual event so we can just return the raw event. 
Would you suggest another wording?

tlebon (approved):
failing some tests but i approve

packages/core/src/main/notification/NotificationService.ts
tlebon: maybe to me its more clear if message is renamed to proposalMessage on line 273? and then this comment is removed. its fine either way

atomrc: well this `message` _can_ be a text message. But if there are `delay` or `proposal` content, then we know it's not a text message and we can early return avoiding the text-specific code to execute

PR #4401, opened 2022-09-20
feat: leave mls convo - refactor (FS-683)

PatrykBuniX:
Remove `leaveConversation` method - duplicated functionality of `removeUserFromProteusConversation` (renamed to `removeUserFromConversation`)


packages/core/src/main/notification/NotificationService.test.browser.js
```diff
@@ -50,10 +50,10 @@ describe('NotificationService', () => {
       });
 
       const notificationService = new NotificationService(apiClient, {}, {}, engine);
-      jest.spyOn(notificationService.database, 'getLastEventDate').
```
PatrykBuniX, line 53: any reason for this here? 😆

packages/core/src/main/conversation/ConversationService/ConversationService.ts
```diff
@@ -856,21 +856,18 @@ export class ConversationService {
...
-      domain: this.apiClient.context.domain,
-      id: this.apiClient.context.userId,
-    });
-  }
-
   /**
-   * @depricated seems not to be used and is outdated. use leaveConversation instead
+   * @depricated seems not to be used and is outdated. use removeUserFromConversation instead
    */
   public async leaveConversations(conversationIds?: string[]): Promise<ConversationMemberLeaveEvent[]> {
```
PatrykBuniX, line 862: Can't see a place where it's being used, can we remove it?

packages/core/src/main/notification/NotificationService.test.browser.js
atomrc: wow 😱

atomrc (approved):

PR #4402, opened 2022-09-20
feat: update yarn to v2, update typescript

aweiss-dev:
This PR concentrates on workspace/tooling improvements and dependency updates.

The following has been done:

- update yarn to v2 (https://yarnpkg.com/)
- use yarn 2 typescript-plugin (https://yarnpkg.com/features/plugins/#official-plugins)
- update typescript globally to the latest version
- update other dependencies to match typescript (ts-node, karma etc.)
- move some dev-dependencies from the dependencies array in package.json to devDepenendencies 
- fix a lot of missing dependency problems in the packages (yarn2 is more strict)
- fix some ts related errors
- fix most of yarns resolution warnings
- link workspace dependencies accordingly
- remove lerna bootsrap from the pipeline (https://github.com/lerna/lerna/issues/2910)
- remove keyword 'type' from imports, since it was problematic together with ts-node


README.md
```diff
@@ -49,7 +49,7 @@ Collection of npm modules created by Wire's web team:
 
 ```bash
 ## Download dependencies and link packages
-yarn && yarn boot
+yarn
```
atomrc, line 52: does yarn2 provides the exact same feature as lerna bootstrap? Why did we need the bootstrap before?

package.json
```diff
@@ -49,19 +49,13 @@
...
+    "test:all": "yarn dist && lerna run --no-sort --concurrency 8 test"
   },
   "workspaces": {
-    "nohoist": [
-      "**/@jest/**/*",
-      "**/@types/!(long)**/*",
-      "**/emotion-normalize",
-      "**/jest-emotion",
-      "**/react"
-    ],
```
atomrc, line 62: What was this?

packages/copy-config/package.json
```diff
@@ -1,7 +1,5 @@
 {
-  "bin": {
-    "copy-config": "src/main/cli.js"
```
atomrc, line 3: are you sure about this change? Sounds like a breaking change to me,  doesn't it?

README.md
aweiss-dev: as far as i understand, lerna bootstrap is from ancient times when npm had no workspaces and yarn didn't exist.
Yarn2 has a very good workspace integration that takes care of the things lerna bootstrap was supposed to do.

package.json
aweiss-dev: This is yarn hoisting (decides where to place npm modules). Its handled different in yarn2 (take a look into the yarn.yml), so this code is not needed anymore (yarn actually warns you that is deprecated)

packages/copy-config/package.json
aweiss-dev: good catch, this one was not on purpose

atomrc (approved):
Super nice, this feels quite clean. 
Thanks man good job

PR #4417, opened 2022-10-03
fix: Improve websocket lifecycle hooks reliability

atomrc:
This will give a clean event API for handling the websocket connection states. We now explicitely call the `onConnectionStateChange` with the appropriate values

BREAKING CHANGE: the `onConnected` callback has been removed in favor of a clean lifecycle event `ConnectionState.LIVE`

### DIFF
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -61,6 +61,17 @@ enum TOPIC {
   ERROR = 'Account.TOPIC.ERROR',
 }
 
+export enum ConnectionState {
+  /** The websocket is closed and notifications stream is not being processed */
+  CLOSED = 'closed',
+  /** The websocket is being opened */
+  CONNECTING = 'connecting',
+  /** The websocket is open but locked and notifications stream is being processed */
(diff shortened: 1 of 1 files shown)


packages/api-client/src/tcp/WebSocketClient.ts
```diff
@@ -111,12 +110,10 @@ export class WebSocketClient extends EventEmitter {
       // before we try any connection, we first refresh the access token to make sure we will avoid concurrent accessToken refreshes
       await this.refreshAccessToken();
     }
-    // Note: Do NOT await `onConnect` otherwise the websocket will not connect during notification stream processing
-    void this.onConnect();
```
atomrc, line 115: The problem lies here in the fact that we would call the `onConnect` **before** actually being connected to the websocket. 
Which means that we start processing the notification stream but there could be messages that come in between the moment we start processing the notification stream and the moment the websocket is actually connected (and thus we have potential missing messages)

thisisamir98 (approved):

PatrykBuniX (approved):

PR #4418, opened 2022-10-03
fix: Avoid message gaps during websocket connection

atomrc:
This changes the behavior of the initial connection to the websocket. 

Currently the flow goes like this:
- we call `onConnect` function (which will download the notification stream);
- we connect to the websocket (not waiting for the notification stream to have been fully parsed)

So in between the time we start processing the notification stream and the time the websocket is connected, we could potentially miss messages. 

Let's take a example to illustrate this (x axis is time):

```text
ws:                  --1----2-------
message:             -----1---------
notif stream:        -1-------2-----
```

- we start the connection flow
- `ws 1`: websocket connection starts
- `notif stream 1`: we start downloading the notification stream
- `message 1`: one message has been sent to this device
- `ws 2`: the websocket connection is up but `message1` has been sent before
- `notif stream 2`: the notification stream has been processed but `message1` was sent after the download

This change makes sure the websocket connection is fully live before downloading the notification stream

```text
ws:                  --1----2---------
message:             -----1-----------
notif stream:        --------1-------2
```

Since we download the notification stream **only** after the websocket is fully connected the `message1` will appear in the notification stream (since we download the notification stream slightly later). We then fill the gap where message could be lost.

### DIFF
--- a/packages/api-client/src/tcp/WebSocketClient.ts
+++ b/packages/api-client/src/tcp/WebSocketClient.ts
@@ -19,7 +19,7 @@
 
 import {EventEmitter} from 'events';
 import logdown from 'logdown';
-import {CloseEvent, ErrorEvent, Event} from 'reconnecting-websocket';
+import {CloseEvent, ErrorEvent} from 'reconnecting-websocket';
 
 import {HttpClient, NetworkError} from '../http/';
 import {InvalidTokenError, MissingCookieError} from '../auth/';
@@ -62,7 +62,6 @@ export class WebSocketClient extends EventEmitter {
   public client: HttpClient;
   private isSocketLocked: boolean;
   private bufferedMessages: string[];
-  private onConnect: () => Promise<void> = () => Promise.resolve();
   private abortHandler?: AbortHandler;
 
   public static readonly TOPIC = TOPIC;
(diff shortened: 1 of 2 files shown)


packages/api-client/src/tcp/WebSocketClient.ts
```diff
@@ -138,25 +135,19 @@ export class WebSocketClient extends EventEmitter {
...
-        this.abortHandler = new AbortHandler();
-        try {
-          this.logger.info('Calling "onConnect"');
-          await onConnect(this.abortHandler);
-        } catch (error) {
-          this.logger.warn(`Error during execution of "onConnect"`, error);
-          this.emit(WebSocketClient.TOPIC.ON_ERROR, error);
-        }
-        this.onStateChange(this.socket.getState());
-      };
```
atomrc, line 153: The problem lies here in the fact that we would call the onConnect before actually being connected to the websocket.
Which means that we start processing the notification stream but there could be messages that come in between the moment we start processing the notification stream and the moment the websocket is actually connected (and thus we have potential missing messages)

thisisamir98 (approved):

PatrykBuniX (approved):

PR #4419, opened 2022-10-04
feat: update emotion

aweiss-dev:
## Pull Request Checklist


packages/react-ui-kit/src/Layout/Theme.tsx
```diff
@@ -285,10 +285,13 @@ export const themes: {[themeId in THEME_ID]: Theme} = {
...
   theme?: Theme;
+  children?: React.ReactNode;
 }
 
 const filterThemeProps = (props: ThemeProps) => filterProps(props, ['theme']);
 
 export const ThemeProvider = (props: ThemeProps) => (
-  <EmotionThemeProvider theme={props.theme} {...filterThemeProps(props)} />
+  <EmotionThemeProvider theme={props.theme} {...filterThemeProps(props)}>
+    {props.children}
```
atomrc, line 295: I believe you are passing the children twice (both in the `props` and as direct children). Am I wrong?

atomrc: Confirm, in this case React will use the direct children rather than the property. 

```js
function App(props) {
  return (
    <div className='App' children={<div>ok</div>}>
      <div>ok2</div> {/* < this one will be rendered/*} 
    </div>
  );
}
```

Let's filter the `children` from the props and use the direct children approach

atomrc (changes requested):
Let's improve the `children` situation

packages/react-ui-kit/src/Layout/Theme.tsx
aweiss-dev: We don't need to filter at all. The current filter only filtered the theme, which gets passed along anyways (needs to be explicitly passed because it can be undefined).

Since a Themeprovider is useless without children (and it also forces us to have defined children), I just added the children as a requirement now to satisfy the types.

atomrc (approved):

PR #4420, opened 2022-10-04
refactor: cleanup test for api-client

atomrc:
This removes the dependencies to webpack/karma related process that are slowing down quite a lot. 
We need generate a pristine tsc compiled library that doesn't require building for a browser.


packages/api-client/src/APIClient.test.ts
```diff
@@ -281,10 +281,10 @@ describe('APIClient', () => {
...
       const testError = new Error('Test rejection');
 
-      spyOn(client.api.auth, 'postLogout').and.returnValue(Promise.reject(testError));
-      spyOn(client, 'disconnect').and.returnValue();
-      spyOn(client['accessTokenStore'], 'delete').and.returnValue(Promise.resolve(undefined));
-      spyOn(client['logger'], 'warn').and.returnValue();
+      jest.spyOn(client.api.auth, 'postLogout').mockReturnValue(Promise.reject(testError));
+      jest.spyOn(client, 'disconnect').mockReturnValue();
+      jest.spyOn(client['accessTokenStore'], 'delete').mockReturnValue(Promise.resolve(undefined));
+      jest.spyOn(client['logger'], 'warn').mockReturnValue();
```
tlebon, line 287: maybe makes sense to destructure spyOn from jest just to have a few less characters?

packages/api-client/src/APIClient.test.ts
```diff
@@ -293,10 +293,10 @@ describe('APIClient', () => {
...
       const client = new APIClient();
 
-      spyOn(client.api.auth, 'postLogout');
-      spyOn(client, 'disconnect').and.returnValue();
-      spyOn(client['accessTokenStore'], 'delete').and.returnValue(Promise.resolve(undefined));
-      spyOn(client['logger'], 'warn').and.returnValue();
+      jest.spyOn(client.api.auth, 'postLogout');
+      jest.spyOn(client, 'disconnect').mockReturnValue();
+      jest.spyOn(client['accessTokenStore'], 'delete').mockReturnValue(Promise.resolve(undefined));
+      jest.spyOn(client['logger'], 'warn').mockReturnValue();
```
tlebon, line 299: same comment 
https://github.com/wireapp/wire-web-packages/pull/4420/files#r987001260

thisisamir98 (approved):

packages/api-client/src/tcp/WebSocketClient.test.ts
```diff
@@ -58,83 +58,85 @@ const fakeSocket = {
 describe('WebSocketClient', () => {
   describe('handler', () => {
     it('calls "onOpen" when WebSocket opens', async () => {
-      const websocketClient = new WebSocketClient('url', fakeHttpClient);
-      const onOpenSpy = spyOn<any>(websocketClient, 'onOpen').and.callThrough();
+      const websocketClient = new WebSocketClient('ws://url', fakeHttpClient);
+      const onOpenSpy = jest.spyOn(websocketClient as any, 'onOpen');
```
tlebon, line 62: maybe destructuring makes sense in this file also since its called many times. also wondering about all these `as any` blocks if they can be improved

packages/api-client/src/conversation/ConversationAPI/ConversationAPI.test.ts
```diff
@@ -20,9 +20,11 @@
 import {HttpClient} from '../../http';
 import {ConversationAPI} from './ConversationAPI';
 
-const httpClientMock = jasmine.createSpyObj('httpClient', {sendJSON: () => ({data: ''})});
-
 describe('ConversationAPI', () => {
+  const httpClientMock = new HttpClient({urls: {name: 'test', rest: 'https://test', ws: 'sw://test'}}, {} as any);
```
tlebon, line 24:
```suggestion
  const httpClientMock = new HttpClient({urls: {name: 'test', rest: 'https://test', ws: 'ws://test'}}, {} as any);
```

packages/api-client/src/tcp/WebSocketClient.test.ts
atomrc: they sadly cannot be improved easily since the test is relying on mocking private methods :/ 
This would be too much of a hustle trying to fix this, so I'm restoring to the original behavior

packages/api-client/src/APIClient.test.ts
atomrc: If you don't have strong opinion on this, I'd like to keep it this way. I don't think character count is a very useful metrics

tlebon (approved):

PR #4425, opened 2022-10-10
refactor: Streamine the message sending API

atomrc:
BREAKING CHANGE: The MessageBuilder API has changed quite a bit and doesn't require a `conversation` or `sender` property anymore. Those have been moved to the `conversationService.send` method

### The situation

We previously had a 2 step transformation of messages before being able to send them. 
The consuming side would have to call `MessageBuilder.create.*()` and then internally we would convert it to a `GenericMessage`. 
**But** this intermediate representation, doesn't hold any meaning and has duplicated informations in the payload (like the conversation or the sender which are also passed to the call to `ConversationService.send`. 
It also required, when adding a new type, to add:
- a new json representation
- a new json to GenericMessage converter

Now adding a new type of message is streamline and the webapp could actually send custom message without needing any update from the core

### What's new

Now the intermediate representation is gone. For the consumer it changes few things as the `MessageBuilder` stays relatively similar.


packages/bot-api/src/MessageHandler.test.node.ts
```diff
@@ -24,7 +24,7 @@ import UUID from 'uuidjs';
 import {ClientType} from '@wireapp/api-client/src/client';
 import {TextMessage} from '@wireapp/core/src/main/conversation/message/OtrMessage';
 import {Connection} from '@wireapp/api-client/src/connection';
-import {MessageBuilder} from '@wireapp/core/src/main/conversation/message/MessageBuilder';
+import {MessageBuilder} from '@wireapp/core';
```
atomrc, line 27: Since we now expose the `index.ts` of the core, we can access the message builder directly :)

packages/core/package.json
```diff
@@ -65,7 +65,7 @@
     "!src/main/**/*test*.ts"
   ],
   "license": "GPL-3.0",
-  "main": "src/main/Account",
+  "main": "src/main/index",
```
atomrc, line 68: This will help us move from outputting the built file in `src` folder. We can here export whatever is needed to the consumer

aweiss-dev (approved):

PR #4428, opened 2022-10-12
fix: Avoid unlocking websocket when connection is aborted

atomrc:
We have identified a potential missing message gap in the handling of the notification stream. 

Here is the full explanation for this:

We we connect to the websocket, the very first thing we do is locking the websocket (so that no message coming in will be parsed before the notification stream). 
Once locked we start processing the notifications given by the `/notifications` endpoint one by one. 

**Here is the catch** if, somehow, the websocket loses the connection **while** we are processing the notification stream, we are going to cancel the notification handling process **BUT** we are not going to avoid `unlocking` the websocket. 

As a result, if ever, there were some messages buffered in the websocket, they are going to be released at this point but we will still have missed the notifications thats were canceled before. 

Here is a concrete example:

- I regain internet connection 
- the websocket connects successfully
- we have 10 notifications that we fetch from backend
- while processing the notifications, we receive 1 notification from the websocket
- at notification 5, the websocket connection is lost (bad internet, backend issue...)
- we cancel the notification handling process (there are still 5 other notifications to process)
- we release the websocket (as well as the 1 message stuck in the websocket buffer)
- the webapp processes this buffered message
-> we now store the last notification ID as the last message from the websocket. This will discard any previous notification that we haven't processed (the 5 notifications that we canceled).

This is coherent with report logs we got
Screenshot 2022-10-12 at 16 28 24

### DIFF
--- a/packages/core/src/main/Account.ts
+++ b/packages/core/src/main/Account.ts
@@ -616,7 +616,7 @@ export class Account<T = any> extends EventEmitter {
       // Lock websocket in order to buffer any message that arrives while we handle the notification stream
       onConnectionStateChanged(ConnectionState.PROCESSING_NOTIFICATIONS);
       this.apiClient.transport.ws.lock();
-      await this.service!.notification.handleNotificationStream(
+      const results = await this.service!.notification.processNotificationStream(
         async (notification, source, progress) => {
           await handleNotification(notification, source);
           onNotificationStreamProgress(progress);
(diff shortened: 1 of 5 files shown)


packages/core/src/main/Account.ts
```diff
@@ -616,19 +616,24 @@ export class Account<T = any> extends EventEmitter {
...
           onNotificationStreamProgress(progress);
         },
         onMissedNotifications,
         abortHandler,
       );
+      this.logger.log(`Finished processing notifications ${JSON.stringify(results)}`, results);
+      if (abortHandler.isAborted()) {
+        this.logger.warn('Ending connection process as websocket was closed');
+        return;
+      }
```
atomrc, line 631: The actual fix is only this, stopping the entire process when the aborthandler is actually aborted

packages/core/src/main/notification/NotificationService.ts
```diff
@@ -141,28 +140,44 @@ export class NotificationService extends EventEmitter {
...
+    const lastNotificationId = await this.database.getLastNotificationId();
+    const {notifications, missedNotification} = await this.getAllNotifications(lastNotificationId);
     if (missedNotification) {
       onMissedNotifications(missedNotification);
     }
 
+    const results = {total: notifications.length, error: 0, success: 0};
+    const logMessage =
+      notifications.length > 0
+        ? `Start processing ${notifications.length} notification since notification id ${lastNotificationId}`
```
V-Gira, line 157:
```suggestion
        ? `Start processing ${notifications.length} notifications since notification id ${lastNotificationId}`
```

V-Gira (approved):

PR #4429, opened 2022-10-13
fix: Fine tune websocket connection and add tests

atomrc:


packages/core/src/main/Account.test.node.ts
```diff
@@ -50,6 +53,26 @@ async function createAccount(storageName = `test-${Date.now()}`): Promise<{accou
   return {account, apiClient};
 }
 
+const waitFor = (assetion: () => void) => {
```
PatrykBuniX, line 56: typo `assertion`?

atomrc: 😅

PatrykBuniX (approved):

PR #4432, opened 2022-10-18
chore: Setup dependabot

atomrc:


tlebon (approved):

tlebon:
i suggest updating auto-merge to make sure we don't wake up to 100 PRs on monday

.github/auto-merge.yml
```diff
@@ -1,3 +1,3 @@
 - match:
     dependency_type: all
-    update_type: 'semver:major'
+    update_type: 'semver:minor'
```
tlebon, line 3: i guess patch also

tlebon: some nice ideas here, similar to our other auto-merge.yml files https://github.com/marketplace/actions/enable-github-automerge

atomrc: Oh I'd expect `minor` to also include what is smaller that minor (so to include `patch` in it). Isn't it?

atomrc: Oh it does include `patch`
image
(cf. https://github.com/ahmadnassri/action-dependabot-auto-merge#configuration-file-syntax)

PR #4507, opened 2022-10-18
feat(core, api-client): Add sender client id to mls message event (FS-1045)

thisisamir98:
## Pull Request Checklist


packages/core/src/main/notification/NotificationService.ts
```diff
@@ -571,3 +586,5 @@ export class NotificationService extends EventEmitter {
     this.scheduleKeyPackagesSync(nextKeyPackagesQueryDate);
   }
 }
+
+console.info('BARDIA YALC');
```
PatrykBuniX, line 590: 😄

PatrykBuniX (approved):

atomrc (approved):

packages/core/src/main/notification/NotificationService.ts
thisisamir98: oops! 😄

PR #4544, opened 2022-10-20
refactor: Factorize reencryption logic

atomrc:
This PR will help in the effort of bringing the fingerprint computation to the core. 
The idea is to gather all manipulation of prekeys only inside of the `CryptoService`. It should be the only entity aware of prekeys. 

Prekeys will be needed in order to create missing sessions with devices we will want to compute fingerprint of

### DIFF
--- a/packages/core/src/main/conversation/message/MessageService.ts
+++ b/packages/core/src/main/conversation/message/MessageService.ts
@@ -130,10 +130,10 @@ export class MessageService {
     const send = (payload: QualifiedOTRRecipients) => {
       return this.sendFederatedOtrMessage(sendingClientId, payload, options);
     };
-    const encrypted = await this.cryptographyService.encryptQualified(plainText, recipients);
+    const encryptedPayload = await this.cryptographyService.encryptQualified(plainText, recipients);
 
     try {
-      return await send(encrypted);
+      return await send(encryptedPayload);
     } catch (error) {
       if (!this.isClientMismatchError(error)) {
         throw error;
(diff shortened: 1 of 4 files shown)


packages/core/src/main/conversation/message/MessageService.ts
```diff
@@ -80,10 +80,7 @@ export class MessageService {
       cipherText = externalPayload.cipherText;
     }
 
-    const {encrypted, missing} = await this.cryptographyService.encrypt(plainTextPayload, recipients);
-    const encryptedPayload = Object.keys(missing).length
-      ? await this.reencryptAfterMismatch({missing, deleted: {}}, encrypted, plainText)
-      : encrypted;
+    const encryptedPayload = await this.cryptographyService.encrypt(plainTextPayload, recipients);
```
atomrc, line 83: Now the `encrypt` method automatically encrypts for clients that do not have a session yet (no more missing recipients)

packages/core/src/main/cryptography/CryptographyService.ts
```diff
@@ -207,18 +204,59 @@ export class CryptographyService {
           missing[userId].push(clientId);
         }
       }
+      const measuredTime = Date.now() - startedAt;
+      this.logger.info(
+        `Encrypted payload for ${Object.keys(encrypted).length} recipients in ${measuredTime}ms (domain: ${domain})`,
```
atomrc, line 209: some performance logging, that's always nice 😎

atomrc:
Close in favor of https://github.com/wireapp/wire-web-packages/pull/4643

PR #4545, opened 2022-10-20
feat: Add fingerprint generation for local and remove clients

atomrc:
This adds 2 new methods to the `CryptographyService`:
- `getLocalFingerprint` that will give back the fingerprint of the client the user is currently using
- `getRemoteFingerprint` that will give the fingerprint of another device


PatrykBuniX (approved):
😎

PR #4546, opened 2022-10-21
chore: remove karma + jasmine and replace with jest globally

aweiss-dev:
### Until now we used a couple of solutions like:

- jasmine
- kasmine + karma
- jest + babel
- jest + ts-jest

### The goal of this PR is:

1. Use Jest for every package of the Monorepo
2. Use Typescript for all of our test files
3. Speed up test execution.
4. Have a common config in the workspace root and just additions in the packages themselves
5. Make life easier for future updates (Compiler changes etc.)

### What has been done to archive the Goal:

- [x] Remove jasmine + karma globally
- [x] Transfer .js tests to .ts
- [x] Create a common jest config that is shared between all packages
- [x] Set up jest for every package
- [x] Make jest executable from root and the packed themselves

#### 1. Running jest from root
When running jest from root by executing "yarn jest", some package tests will fail because dependencies like TextEncoder can't be found or some global variables are not defined. I don't have a fix for that (yet), but its also not on the priority list since we execute scripts directly from the package folder (that don't have this problem). 
As soon as we have a fix for this behaviour we can just execute jest from root to collect combined coverage of all packages. Also enables tools like Wallaby to be fully usable from root without having some tests error.

#### 2. Running build before test execution
Some test (not many) need the built binaries from other packages to be able to execute their tests. Currently we build every package before running the tests, which costs a lot of time. We have a high potential to speed up the process if we cut down the build step before test execution.


packages/webapp-events/package.json
```diff
@@ -1,6 +1,8 @@
 {
   "description": "Events shared between Wire's web app and desktop app.",
   "devDependencies": {
+    "@swc/core": "^1.3.10",
```
atomrc, line 4: I don't think it's needed here

.prettierrc
```diff
@@ -0,0 +1 @@
+{}
```
atomrc, line 1: This is a mistake, right?

bin/testUpdated.js
```diff
@@ -17,7 +17,7 @@
  *
  */
 
-const {execSync} = require('child_process');
+const { execSync } = require('child_process');
```
atomrc, line 20: Those changes are probably related to the empty prettier config, please revert

packages/archive/store-engine-sqleet/src/SQLeetEngine.test.ts
```diff
@@ -148,7 +148,6 @@ describe('SQLeetEngine', () => {
       try {
         const entity = {'name\'"`': 'Otto'};
         await engine.create('users', '1', entity);
-        fail();
```
atomrc, line 151: Why?!

packages/commons/src/main/util/RandomUtil.test.ts
```diff
@@ -29,10 +31,10 @@ describe('RandomUtil', () => {
       const randomLimit = 10;
       for (let index = 0; index < iterations; index++) {
         if (RandomUtil.randomInt(randomLimit) > randomLimit) {
-          fail(`randomInt() exeeded limit "${randomLimit}"`);
```
atomrc, line 32: Why using throw here? Fail is the way to make a test fail and is supposed to have a better report than a throw, isn't it?

atomrc (changes requested):
Maybe we could try to hoist the import of jest and swc to the root package.json and remove them from every single package, WDYT?

packages/archive/store-engine-sqleet/src/SQLeetEngine.test.ts
aweiss-dev: Good catch. Fail doesn't exist in jest, but I need to throw an error here

jest.config.js
```diff
@@ -1,6 +1,6 @@
 /*
```
atomrc, line 1: Does this work as a global config? Could we remove all the `jest.config.js` files?

packages/commons/src/main/util/RandomUtil.test.ts
atomrc: Oh my mistake, throwing seems the way to go now with Jest 👍 
https://github.com/facebook/jest/issues/11698

aweiss-dev:
> Maybe we could try to hoist the import of jest and swc to the root package.json and remove them from every single package, WDYT?

I tried this. I could only run tests from root in this case and trying to run them from packages error Ed because the dependencies haven't been found. I think this is the case because the yarn.yaml hoists to packages per setting.

atomrc:
> I tried this. I could only run tests from root in this case and trying to run them from packages error Ed because the dependencies haven't been found. I think this is the case because the yarn.yaml hoists to packages per setting.

I think it's because it's a binary that is executed and in this case it should be defined at the package level. 
Let's try with swc dependencies, they might very well work (but for jest probably not indeed, was worth trying)

aweiss-dev:
> > I tried this. I could only run tests from root in this case and trying to run them from packages error Ed because the dependencies haven't been found. I think this is the case because the yarn.yaml hoists to packages per setting.
> 
> I think it's because it's a binary that is executed and in this case it should be defined at the package level. Let's try with swc dependencies, they might very well work (but for jest probably not indeed, was worth trying)

I actually found a solution. Enjoy :)

jest.config.js
aweiss-dev: No, every package needs a own config, but we can let jest life in the root and have a common config we extend in the children

packages/api-client/src/http/HttpClient.test.ts
```diff
@@ -70,7 +70,6 @@ describe('HttpClient', () => {
 
       try {
         await client._sendRequest({method: 'GET', baseURL: testConfig.urls.rest, url: AuthAPI.URL.ACCESS});
-        fail();
```
aweiss-dev, line 73: Throw

packages/api-client/src/http/HttpClient.test.ts
```diff
@@ -92,7 +91,6 @@ describe('HttpClient', () => {
 
     try {
       await client._sendRequest({method: 'GET', baseURL: testConfig.urls.rest, url: AuthAPI.URL.ACCESS});
-      fail();
```
aweiss-dev, line 95: Throw

packages/certificate-check/package.json
```diff
@@ -4,11 +4,12 @@
   },
   "description": "Utilities to check that Wire's domains use the expected certificate.",
   "devDependencies": {
-    "@types/jasmine": "4.3.0",
+    "@swc/core": "^1.3.10",
+    "@swc/jest": "^0.2.23",
+    "@types/jest": "^29.2.0",
```
aweiss-dev, line 9: Remove modules

packages/cli-client/package.json
```diff
@@ -12,6 +12,8 @@
     "fs-extra": "10.1.0"
   },
   "devDependencies": {
+    "@swc/core": "^1.3.10",
+    "@swc/jest": "^0.2.23",
```
aweiss-dev, line 16: Remove

packages/commons/package.json
```diff
@@ -1,22 +1,25 @@
 {
   "dependencies": {
+    "@types/jest": "^29.2.0",
     "ansi-regex": "5.0.1",
     "fs-extra": "10.1.0",
+    "jest": "^29.2.1",
+    "jest-environment-jsdom": "^29.2.1",
```
aweiss-dev, line 7: Remove

packages/copy-config/package.json
```diff
@@ -10,9 +10,15 @@
   },
   "description": "Wire's internal configuration copy tool",
   "devDependencies": {
+    "@swc/core": "^1.3.10",
+    "@swc/jest": "^0.2.23",
```
aweiss-dev, line 14: Remove

packages/core/package.json
```diff
@@ -24,19 +24,19 @@
     "@babel/preset-env": "^7.19.1",
     "@babel/preset-typescript": "^7.18.6",
     "@faker-js/faker": "^7.6.0",
+    "@swc/core": "^1.3.10",
+    "@swc/jest": "^0.2.23",
```
aweiss-dev, line 28: Remove

packages/priority-queue/package.json
```diff
@@ -1,5 +1,7 @@
 {
   "devDependencies": {
+    "@swc/core": "^1.3.10",
+    "@swc/jest": "^0.2.23",
```
aweiss-dev, line 4: Remove

packages/priority-queue/package.json
```diff
@@ -30,10 +32,12 @@
...
-    "test:project": "yarn dist && yarn test",
-    "test:node": "cross-env JASMINE_CONFIG_PATH=src/test/node/support/jasmine.json jasmine"
+    "test": "jest"
   },
-  "version": "1.8.10"
+  "version": "1.8.10",
+  "dependencies": {
+    "@types/jest": "^29.2.0",
+    "jest": "^29.2.1",
+    "jest-environment-jsdom": "^29.2.1"
```
aweiss-dev, line 41: Remove

packages/promise-queue/package.json
```diff
@@ -3,6 +3,8 @@
     "@babel/core": "^7.19.1",
     "@babel/preset-env": "^7.19.1",
     "@babel/preset-typescript": "7.18.6",
+    "@swc/core": "^1.3.10",
+    "@swc/jest": "^0.2.23",
```
aweiss-dev, line 7: Remove

packages/react-ui-kit/package.json
```diff
@@ -16,9 +16,11 @@
     "@emotion/jest": "11.10.0",
     "@emotion/react": "^11.10.4",
     "@hot-loader/react-dom": "17.0.2",
+    "@swc/core": "^1.3.10",
+    "@swc/jest": "^0.2.23",
```
aweiss-dev, line 20: Remove

packages/store-engine/package.json
```diff
@@ -1,15 +1,13 @@
 {
   "devDependencies": {
-    "@types/jasmine": "4.3.0",
-    "@types/karma": "6.3.3",
+    "@swc/core": "^1.3.10",
+    "@swc/jest": "^0.2.23",
```
aweiss-dev, line 4: Remove

atomrc (approved):
Solid 💪

PR #4559, opened 2022-10-24
fix: More resilient id comparison for external proposals

atomrc:
It can happen that core-crypto is giving us uppercase user ID when requesting is a user is part of a group or not. 
We need to be more resilient about it and always lower case the ids given by core-crypto

### DIFF
--- a/packages/core/src/main/mls/MLSService/MLSService.ts
+++ b/packages/core/src/main/mls/MLSService/MLSService.ts
@@ -95,7 +95,7 @@ export class MLSService {
         const {user} = parseFullQualifiedClientId(decoder.decode(client));
         return otherClients.some(client => {
           const {user: otherUser} = parseFullQualifiedClientId(decoder.decode(client));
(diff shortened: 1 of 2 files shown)


packages/core/src/main/mls/MLSService/MLSService.ts
```diff
@@ -70,6 +72,9 @@ export class MLSService {
           optionalToUint8Array(commit),
         );
         await this.coreCryptoClient.commitAccepted(groupId);
+        const newEpoch = await this.getEpoch(groupId);
+        const groupIdStr = Encoder.toBase64(groupId).asString;
+        this.logger.log(`Commit have been accepted for group "${groupIdStr}". New epoch is "${newEpoch}"`);
```
atomrc, line 77: Some more logging, because we like logs

V-Gira (approved):

thisisamir98 (approved):

PR #4561, opened 2022-10-25
refactor: produce output of packages in lib folders

aweiss-dev:
## The goal of this PR is to parse trough all packages and improve the workflow / setup

This includes:

- [x] Produce build output in a "lib" folder, instead of the src folder
- [x] Streamline package.json's
- [x] Streamline tsconig.js - files
- [x] Remove src/main folders (if they still exist). Put files in src/ directly
- [x] Clean lib-folder before a new build

[BREAKING CHANGE](https://conventionalcommits.org/#summary): Imports that point to `@wireapp/packageName/src/**` need to be changed to `@wireapp/packageName/lib/**`


packages/webapp-events/package.json
```diff
@@ -1,22 +1,22 @@
...
+        "rimraf": "^3.0.2",
+        "typescript": "4.8.4"
+    },
+    "scripts": {
+        "build": "yarn clean && tsc",
+        "clean": "rimraf lib",
+        "dist": "yarn build",
+        "test": "exit 0"
+    },
+    "version": "0.14.9"
```
atomrc, line 21: I think you have indentation problem here

atomrc (approved):
Amazing 🤩 💪

packages/webapp-events/package.json
aweiss-dev: Fixed it with my latest commit. package.json has been on the prettier ignore list, not anymore

PR #4567, opened 2022-10-27
feat: add initial db migration trigger logic [FS-1095]

PatrykBuniX:
We want to migrate from cryptobox to corecrypto. We will trigger a migration script after dexie db version has upgraded (`upgrade` hook on webapp side). 

Blockers:
We need to have core-crypto proteus client initialised to call `coreCryptoClient.proteusCryptoboxMigrate` method. [FS-1096](https://wearezeta.atlassian.net/browse/FS-1096)

### DIFF
--- a/packages/core/src/mls/MLSService/MLSService.ts
+++ b/packages/core/src/mls/MLSService/MLSService.ts
@@ -237,4 +237,8 @@ export class MLSService {
   public async wipeConversation(conversationId: ConversationId): Promise<void> {
     return this.coreCryptoClient.wipeConversation(conversationId);
   }
+
+  public async proteusCryptoboxMigrate(storeName: string) {
+    return this.coreCryptoClient.proteusCryptoboxMigrate(storeName);
+  }
 }
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -145,9 +145,16 @@ type InitOptions = {
    * An unknown client is a client we don't yet have a session with
    */
   onNewClient?: (sessionId: SessionId) => void;
-};
 
-type DBMigrationConfig = {storeName: string};
+  /**
+   * Database migration config object storing a name of the db and onSuccess callback function.
+   * When provided, the client will trigger migration script after initialisation.
+   */
+  dbMigrationConfig?: {
+    storeName: string;
+    onSuccess: () => void;
+  };
+};
 
 const coreDefaultClient: ClientInfo = {
   classification: ClientClassification.DESKTOP,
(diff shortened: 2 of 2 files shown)


packages/core/src/mls/MLSService/MLSService.ts
```diff
@@ -95,7 +95,7 @@ export class MLSService {
   public configureMLSCallbacks({groupIdFromConversationId, ...coreCryptoCallbacks}: MLSCallbacks): void {
     this.coreCryptoClient.registerCallbacks({
       ...coreCryptoCallbacks,
-      clientIdBelongsToOneOf: (client, otherClients) => {
+      clientIsExistingGroupUser: (client, otherClients) => {
```
PatrykBuniX, line 98: it was renamed in new version of core-crypto

packages/core/src/mls/MLSService/MLSService.ts
```diff
@@ -237,4 +237,8 @@ export class MLSService {
   public async wipeConversation(conversationId: ConversationId): Promise<void> {
     return this.coreCryptoClient.wipeConversation(conversationId);
   }
+
+  public async proteusCryptoboxMigrate(storeName: string) {
+    return this.coreCryptoClient.proteusCryptoboxMigrate(storeName);
```
PatrykBuniX, line 242: this will probably live in some other place (waiting for proteus client to be initialised)

packages/core/src/mls/types.ts
```diff
@@ -25,7 +25,7 @@ type SecretCrypto<T> = {
   decrypt: (payload: T) => Promise<Uint8Array>;
 };
 
-export interface MLSCallbacks extends Pick<CoreCryptoCallbacks, 'authorize'> {
+export interface MLSCallbacks extends Pick<CoreCryptoCallbacks, 'authorize' | 'userAuthorize'> {
```
PatrykBuniX, line 28: seems to be required now

packages/core/src/mls/MLSService/MLSService.ts
atomrc: you might need to defined the `authorize` callback also :/

atomrc (commented):
I wonder if the API is the best. The idea is definitely good, but this creates a temporal coupling between `scheduleDBMigration` and `init`. (if you call `init` before `schedule` then the migration will never happen, and these are hard to debug situation). 

How about, instead, the consumer (the webapp), is responsible for storing the fact that a migration is needed and passes this as a param to the `init` function?

PatrykBuniX:
Closing in favour of https://github.com/wireapp/wire-web-packages/pull/4647.

PR #4574, opened 2022-10-31
feat:Core Crypto proteus initiation

tlebon:
## Pull Request Checklist


packages/core/src/Account.ts
```diff
@@ -245,43 +250,67 @@ export class Account<T = any> extends EventEmitter {
...
-    // Assumption: client gets only initialized once
-    if (initClient) {
-      await this.initClient({clientType});
+      // client gets initialized twice: once for proteus/ once for mls
+      this.proteusClient = await CoreCrypto.init({
+        databaseName: `corecrypto-${this.generateDbName(context)}-proteus`,
+        key: Encoder.toBase64(key).asString,
+        clientId: `PROTEUS_${context.clientId}`,
+        wasmFilePath: this.cryptoProtocolConfig!.coreCrypoWasmFilePath,
+        ...(entropyData && {entropySeed: entropyData}),
```
tlebon, line 278: not really sure what to do about this right now.

atomrc: I think it should only be in the `register` case, doesn't it? 
Once the identification of the device has been created, we don't need extra entropy anymore. 

When we call `init` we know that the device has been created, thus we do not need extra entropy

packages/core/src/Account.ts
```diff
@@ -245,43 +250,67 @@ export class Account<T = any> extends EventEmitter {
...
+        : await createEncryptedStore(dbName);
 
-        await this.apiClient.api.client.putClient(context.clientId!, {prekeys});
-        this.logger.debug(`Successfully uploaded '${prekeys.length}' PreKeys.`);
-      },
+      let key = await secretStore.getsecretValue(coreCryptoKeyId);
 
-      onNewSession: onNewClient,
-    });
+      if (!key) {
```
aweiss-dev, line 267: Why do we call the same function twice if it fails to return a key to us? This code is hard to understand for me.

packages/core/src/Account.ts
```diff
@@ -368,7 +397,11 @@ export class Account<T = any> extends EventEmitter {
         }
 
         this.logger.log('Last client was permanent - Deleting cryptography stores');
-        await this.service!.cryptography.deleteCryptographyStores();
+        if (this.cryptoProtocolConfig?.proteus) {
+          await this.logout(true);
+        } else {
+          await this.service!.cryptography.deleteCryptographyStores();
+        }
```
atomrc, line 404: can't we always use `logout(true)` here and remove the condition? I guess `logout(true)` will take care of deleting the DB

packages/core/src/mls/MLSService/MLSService.ts
```diff
@@ -92,17 +92,21 @@ export class MLSService {
...
-      clientIsExistingGroupUser: (client, otherClients) => {
-        const decoder = new TextDecoder();
-        const {user} = parseFullQualifiedClientId(decoder.decode(client));
-        return otherClients.some(client => {
-          const {user: otherUser} = parseFullQualifiedClientId(decoder.decode(client));
-          return otherUser.toLowerCase() === user.toLowerCase();
-        });
-      },
+      userAuthorize: (groupId, client, otherClients) =>
+        coreCryptoCallbacks.authorize(groupId, client) && this.clientIsExistingGroupUser(client, otherClients),
```
atomrc, line 108: Humm I think this will need to be implemented on the webapp side, please revert this change

packages/core/src/Account.ts
```diff
@@ -451,9 +480,11 @@ export class Account<T = any> extends EventEmitter {
     const loadedClient = await this.service!.client.getLocalClient();
     await this.apiClient.api.client.getClient(loadedClient.id);
     this.apiClient.context!.clientId = loadedClient.id;
-    await this.service!.cryptography.initCryptobox();
+    if (!this.cryptoProtocolConfig?.proteus && !this.cryptoProtocolConfig?.mls) {
+      await this.service!.cryptography.initCryptobox();
+    }
     if (this.cryptoProtocolConfig?.mls && this.backendFeatures.supportsMLS) {
-      this.coreCryptoClient = await this.createMLSClient(
+      this.mlsClient = await this.createMLSClient(
```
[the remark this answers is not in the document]
aweiss-dev: I agree, I would go a step further and rename createMLSClient() function to getCoreCryptoClientInstance()

atomrc:
closed in favor of #4574

PR #4580, opened 2022-11-03
feat: fetch all notifications when 404 without body returned

PatrykBuniX:
Due to API v3 changes ([see changelog](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/672006169/API+changes+v2+v3)) `/notifications` endpoint has been updated: It does not return **error body** - list of notifications since encoded timestamp (timestamp is not contained in UUIDv4) or from the beginning. 

From now when api responds with 404 error, body does not contain notifications, we have to handle this error and fetch all the notifications from the beginning by manually calling `/notifications` endpoint without `since` param.

### DIFF
--- a/packages/api-client/src/notification/NotificationAPI/NotificationAPI.ts
+++ b/packages/api-client/src/notification/NotificationAPI/NotificationAPI.ts
@@ -19,7 +19,7 @@
 
 import axios, {AxiosRequestConfig} from 'axios';
 
-import {HttpClient} from '../../http';
+import {BackendError, BackendErrorLabel, HttpClient} from '../../http';
 import {Notification, NotificationList} from '..';
 
 export const NOTIFICATION_SIZE_MAXIMUM = 10000;
@@ -108,6 +108,10 @@ export class NotificationAPI {
             hasMissedNotifications = true;
             payload = {...defaultPayload, ...error.response?.data};
           }
+        } else if (error instanceof BackendError && error.label === BackendErrorLabel.NOT_FOUND) {
+          //notification was not found in the database,
+          //we need to load all the notifications from the beginning (without 'since' param)
+          return getNotificationChunks(currentClientId);
         } else {
           throw error;
         }


packages/api-client/src/notification/NotificationAPI/NotificationAPI.ts
```diff
@@ -108,6 +108,10 @@ export class NotificationAPI {
             hasMissedNotifications = true;
             payload = {...defaultPayload, ...error.response?.data};
           }
+        } else if (error instanceof BackendError && error.label === BackendErrorLabel.NOT_FOUND) {
+          //notification was not found in the database,
+          //we need to load all the notifications from the beginning (without 'since' param)
+          return getNotificationChunks(currentClientId);
```
PatrykBuniX, line 114: Since we're fetching all the notifications, should we also reset `notificationList` array to empty array?

packages/api-client/src/notification/NotificationAPI/NotificationAPI.ts
```diff
@@ -108,6 +108,10 @@ export class NotificationAPI {
             hasMissedNotifications = true;
             payload = {...defaultPayload, ...error.response?.data};
           }
+        } else if (error instanceof BackendError && error.label === BackendErrorLabel.NOT_FOUND) {
```
PatrykBuniX, line 111: older API's (e.g. `v2`) should not be affected since they are always `AxiosError` type and contain body with `notifications`

packages/api-client/src/notification/NotificationAPI/NotificationAPI.ts
atomrc: hummm does this returns a payload that says `{missedNotifications: true, ...}`?

Can you maybe write a test case for both scenarios?

PatrykBuniX: Sure!

packages/api-client/src/notification/NotificationAPI/NotificationsAPI.test.ts
```diff
@@ -75,5 +78,28 @@ describe('NotificationAPI', () => {
...
+      //second call returns all the notifications with id of missed notification
+      jest
+        .spyOn(client, 'sendJSON')
+        .mockImplementationOnce(() => Promise.reject(ErrorResponse))
+        .mockImplementationOnce(() =>
+          Promise.resolve<AxiosResponse>({
+            status: 200,
+            data: {...mockedResultData},
+          } as AxiosResponse),
+        );
```
PatrykBuniX, line 96: `getNotificationChunks` recursive function that makes the actual api call is called under the hood inside `getAllNotifications`. When backend not found error is caught, we manually send another call to the same endpoint without `since` param, what should give use full list of notifications in the response.

PatrykBuniX:
Not sure what do we want to do with `400` error returned when uuid parser has failed. Backend team says that 400 errors should not be handled but rather considered bugs (what I agree with, since uuid won't be unparseable unless somebody mess up with it manually) wdyt @atomrc? The current behaviour for `v3` api is that the method will respond with `{notifications: [], missedNotification: undefined}`. Do we want to handle 400 the same way we handle 404? 🤔

atomrc (approved):

atomrc (approved):

PR #4609, opened 2022-11-10
feat: Ability to register an already existing conversation into coreCrypto

atomrc:


aweiss-dev (approved):

packages/core/src/mls/MLSService/MLSService.test.ts
```diff
@@ -0,0 +1,82 @@
+/*
```
PatrykBuniX, line 1: great to have tests 💪

PatrykBuniX (approved):
👊

V-Gira (approved):

PR #4610, opened 2022-11-11
feat: extract mthods from ConversationService into ProteusService

aweiss-dev:
[BREAKING CHANGE](https://conventionalcommits.org/#summary): Removed deprecated overwrite of createProteusConversation


tlebon (approved):
nothing really sticks out to me, seems good.

PR #4612, opened 2022-11-14
feat: add a user setting for font sizes [ACC-302]

V-Gira:
### Feature
- Add a property and webapp event for setting the font size in preferences
- see [this ticket](https://wearezeta.atlassian.net/browse/ACC-302) and the linked figma for more information

#### WIP of the feature:

Peek 2022-11-14 17-38

### DIFF
--- a/packages/api-client/src/user/data/UserPropertiesSetData.ts
+++ b/packages/api-client/src/user/data/UserPropertiesSetData.ts
@@ -42,7 +42,7 @@ export interface WebappProperties {
       replace_inline: boolean;
     };
     interface: {
-      fontSize: string;
+      font_size: string;
       theme: 'dark' | 'default';
       view_folders: boolean;
     };
--- a/packages/api-client/src/user/data/UserPropertiesSetData.ts
+++ b/packages/api-client/src/user/data/UserPropertiesSetData.ts
@@ -23,6 +23,18 @@ export enum AudioPreference {
   SOME = 'some',
 }
 
+export enum FontSizePreference {
+  XXS = '10px',
+  XS = '12px',
+  S = '14px',
+  M = '16px',
+  L = '18px',
+  XL = '20px',
+  XXL = '24px',
+  XXXL = '30px',
+  XXXXL = '36px',
+}
+
 export enum NotificationPreference {
   NONE = 'none',
   OBFUSCATE = 'obfuscate',
@@ -42,6 +54,7 @@ export interface WebappProperties {
       replace_inline: boolean;
     };
     interface: {
+      fontSize: FontSizePreference;
       theme: 'dark' | 'default';
       view_folders: boolean;
     };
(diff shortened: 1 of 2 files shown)


packages/api-client/src/user/data/UserPropertiesSetData.ts
```diff
@@ -42,6 +54,7 @@ export interface WebappProperties {
       replace_inline: boolean;
     };
     interface: {
+      fontSize: FontSizePreference;
```
aweiss-dev, line 57: Can be imported from the theme

aweiss-dev (changes requested):

packages/api-client/src/user/data/UserPropertiesSetData.ts
```diff
@@ -23,6 +23,18 @@ export enum AudioPreference {
...
+  XXS = '10px',
+  XS = '12px',
+  S = '14px',
+  M = '16px',
+  L = '18px',
+  XL = '20px',
+  XXL = '24px',
+  XXXL = '30px',
+  XXXXL = '36px',
+}
```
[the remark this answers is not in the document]
atomrc: I tend to agree that the values should live somewhere else. 
I think the webapp, as the consumer of the API, should be the one defining this enum.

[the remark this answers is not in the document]
V-Gira: The theming for font won't be using px values but REM (it'll be the pr on the webapp side). This enum is for a setting in the preferences that allow the user to choose a root font size

packages/api-client/src/user/data/UserPropertiesSetData.ts
atomrc: Not sure we should add a strict type here. It could just be a string. The consuming app is responsible for understanding the string it has put in there

packages/api-client/src/user/data/UserPropertiesSetData.ts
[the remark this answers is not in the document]
aweiss-dev: In this case I agree with Thomas. The enum should be defined in the webapp in this case.

packages/api-client/src/user/data/UserPropertiesSetData.ts
aweiss-dev: Sure, if the value will be defined and managed in the webapp, this can be a string type since the content has no relevance here.

packages/webapp-events/src/index.ts
```diff
@@ -186,6 +186,7 @@ export const WebAppEvents = {
         REPLACE_INLINE: 'wire.webapp.properties.update.emoji.replace_inline',
       },
       INTERFACE: {
+        FONTSIZE: 'wire.webapp.properties.update.interface.font_size',
```
V-Gira, line 189:
```suggestion
        FONT_SIZE: 'wire.webapp.properties.update.interface.font_size',
```

atomrc (approved):

PR #4613, opened 2022-11-14
fix: streamline DecryptionError

aweiss-dev:
## Pull Request Checklist


packages/core/src/messagingProtocols/proteus/EventHandler/ConversationEvent/events/otrMessageAdd/otrMessageAdd.ts
```diff
@@ -18,8 +18,8 @@
  */
 
 import {BackendEvent, ConversationOtrMessageAddEvent, CONVERSATION_EVENT} from '@wireapp/api-client/lib/event';
-import {DecryptionError} from '@wireapp/cryptobox/src/error';
 
+import {DecryptionError} from '../../../../../../errors/DecryptionError';
```
atomrc, line 22: hummm this looks wrong. It should be an error from cryptobox here right?

aweiss-dev: No, because we catch an error that we create ourselves in the CryptocraphyService, which departs from the cryptobox error through a "code" attribute.

PR #4614, opened 2022-11-15
feat: initialise proteus with core crypto [FS-1096]

PatrykBuniX:
Use core crypto instead of cryptobox for proteus (and mls) client initialisation.

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -25,6 +25,7 @@ import {
   Cookie,
   CookieStore,
   LoginData,
+  PreKey,
 } from '@wireapp/api-client/lib/auth';
 import {ClientClassification, ClientType, RegisteredClient} from '@wireapp/api-client/lib/client/';
 import * as Events from '@wireapp/api-client/lib/event';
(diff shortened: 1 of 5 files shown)


packages/core/src/Account.ts
```diff
@@ -356,12 +356,20 @@ export class Account<T = any> extends EventEmitter {
...
+    if (!this.service || !this.apiClient.context) {
       throw new Error('Services are not set.');
     }
 
     try {
-      const localClient = await this.loadAndValidateLocalClient(entropyData);
+      const localClient = await this.loadAndValidateLocalClient();
+      const coreCryptoClient = await this.initCoreCrypto(this.apiClient.context, entropyData);
+
+      await coreCryptoClient.proteusInit();
```
PatrykBuniX, line 367: do we want to always call `.proteusInit()` even for mls devices? or we should add an if statement here?

atomrc: I believe proteus is always going to be init (at least for now). 
MLS should be called on top if MLS is request by the consumer. 

Later (a lot later I believe) we will also condition proteus init, but it's not for now :)

packages/core/src/Account.ts
```diff
@@ -480,23 +512,14 @@ export class Account<T = any> extends EventEmitter {
...
-        this.apiClient.context!,
-        this.cryptoProtocolConfig,
-        entropyData,
-      );
-    }
-
     return loadedClient;
   }
 
+  /*
```
atomrc, line 522: I guess we can completely delete that function now

packages/core/src/client/ClientService.ts
```diff
@@ -115,16 +116,15 @@ export class ClientService {
...
     const prekeys: PreKey[] = [];
+
     for (let i = 0; i < nbPrekeys; i++) {
       const id = i;
       const key = await coreCryptoClient.proteusNewPrekey(i);
       prekeys.push({id, key: Encoder.toBase64(key).asString});
     }
-    const lastPrekeyId = 65535;
+
+    const lastPrekeyId = ProteusKeys.PreKey.MAX_PREKEY_ID;
```
PatrykBuniX, line 127: It stores exact same number

atomrc: Oh perfect, nice catch 👌

atomrc (approved):

PR #4619, opened 2022-11-16
refactor: Streamline creation of coreCrypto instance

atomrc:
Since entropy data can be fed to `coreCrypto` right before proteus/mls devices are initialised, we can actually build our coreCrypto instance very early and have the instance accessible


PatrykBuniX (approved):
love it! 🚀

PR #4624, opened 2022-11-17
feat: encrypt messages with core crypto

PatrykBuniX:
Encrypting messages for proteus with core crypto.


packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -173,4 +178,82 @@ export class ProteusService {
...
+        : Object.keys(users[userId])
+            // We filter out clients that have `null` prekey
+            .filter(clientId => !!users[userId][clientId]);
+      for (const clientId of clientIds) {
+        const sessionId = this.cryptographyService.constructSessionId(userId, clientId, domain);
+        const sessionExists = (await this.coreCryptoClient.proteusSessionExists(sessionId)) as unknown as boolean;
+        if (!sessionExists) {
+          const userQualifiedId = {id: userId, domain: domain ?? ''};
+          await this.createSession(userQualifiedId, clientId);
+        }
```
PatrykBuniX, line 226: Creating a session (if it does not exist already) before encrypting a message.

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -173,4 +178,82 @@ export class ProteusService {
...
+      const clientIds = isUserClients(users)
+        ? users[userId]
+        : Object.keys(users[userId])
+            // We filter out clients that have `null` prekey
+            .filter(clientId => !!users[userId][clientId]);
+      for (const clientId of clientIds) {
+        const sessionId = this.cryptographyService.constructSessionId(userId, clientId, domain);
+        const sessionExists = (await this.coreCryptoClient.proteusSessionExists(sessionId)) as unknown as boolean;
+        if (!sessionExists) {
+          const userQualifiedId = {id: userId, domain: domain ?? ''};
```
PatrykBuniX, line 224: What do we do when there's no domain?

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.test.ts
```diff
@@ -75,12 +80,11 @@ const buildProteusService = (federated = false): [ProteusService, {apiClient: AP
     proteusSessionExists: jest.fn(),
     proteusSessionFromPrekey: jest.fn(),
     proteusFingerprintRemote: jest.fn(),
+    proteusEncrypt: jest.fn(),
```
PatrykBuniX, line 83: I don't think we want to test actual implementation, do we?

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -76,17 +84,14 @@ export class ProteusService {
...
   }
 
-  private async createSession(
-    sessionId: string,
-    userId: QualifiedId,
-    clientId: string,
-    initialPrekey?: PreKey,
-  ): Promise<void> {
+  private async createSession(userId: QualifiedId, clientId: string, initialPrekey?: PreKey): Promise<void> {
+    const sessionId = this.cryptographyService.constructSessionId(userId, clientId);
```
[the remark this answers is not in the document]
atomrc: True, that makes sense 👌

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
atomrc: yep, empty string is the way to go 👌

atomrc: actually, you could just go `domain: string = ''` in the params of the function :)

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -173,4 +178,82 @@ export class ProteusService {
...
+      if ((error as any).code === notFoundErrorCode) {
+        // If the session is not in the database, we just return undefined. Later on there will be a mismatch and the session will be created
+        return undefined;
+      }
+      this.logger.error(`Could not encrypt payload: ${(error as Error).message}`);
+      encryptedPayload = new Uint8Array(Buffer.from('💣', 'utf-8'));
+    }
+
+    return {encryptedPayload, sessionId};
+  }
```
atomrc, line 204: actually we should try to use `coreCrypto.proteusEncryptBatched` this should simplify things a bit and should be way more performant (as we would avoid many back and forth between js and wasm)

atomrc (changes requested):

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
PatrykBuniX: Cool, didn't know that, will refactor the code then :D

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -173,4 +178,76 @@ export class ProteusService {
...
+          encrypted[userId][clientId] = payload.get(sessionId)!;
+        } else {
+          missing[userId] ||= [];
+          missing[userId].push(clientId);
+        }
+      }
+    }
+
+    return {encrypted, missing};
+  }
```
PatrykBuniX, line 233: Not sure how do we want to keep track of missed (`missing`) clients with `proteusEncryptBatched` method... For now I've assumed that the clients that were included in recipients but are not included in returned batched payload are missing. 🤔

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -173,4 +178,90 @@ export class ProteusService {
...
```
PatrykBuniX, line 247: Not sure how do we want to keep track of missed (missing) clients with `proteusEncryptBatched` method... For now I've assumed that the clients that were included in recipients but are not included in returned batched payload are missing. 🤔

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -173,4 +178,90 @@ export class ProteusService {
...
+    domain: string = '',
+  ): Promise<{missing: UserClients; encrypted: OTRRecipients<Uint8Array>}> {
+    const {sessions, userClients} = await this.mapRecipantsToSessionsAndClients(users, domain, true);
+
+    const payload = await this.coreCryptoClient.proteusEncryptBatched(sessions, plainText);
+
+    return this.extractEncryptedAndMissingFromBatchedPayload(payload, userClients);
+  }
+
+  private async mapRecipantsToSessionsAndClients(
```
atomrc, line 194:
```suggestion
  private async mapRecipientsToSessionsAndClients(
```

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
atomrc: I'm lacking the understanding of the behavior of `proteusEncryptBatched` when some sessions are missing. 
If it doesn't error and just returns the payload that it could encrypt, then this sounds like the perfect behavior to detect missings :)

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -173,4 +178,90 @@ export class ProteusService {
...
+
+      for (const clientId of clientIds) {
+        const sessionId = this.cryptographyService.constructSessionId(userId, clientId, domain);
+
+        if (createSession) {
+          const sessionExists = (await this.coreCryptoClient.proteusSessionExists(sessionId)) as unknown as boolean;
+
+          if (!sessionExists) {
+            const userQualifiedId = {id: userId, domain};
+            await this.createSession(userQualifiedId, clientId);
```
atomrc, line 213: hummm maybe we shouldn't create sessions one-by-one. We should probably implement batching session creation (I believe there is a backend endpoint to query multiple user/client prekeys at once)

atomrc: The endpoint is `POST users/list-prekeys`

atomrc:
Super nice PR, can't wait to test this one live :)

A few ideas of acceptance criteris for you to test before merging:

- [ ] sending a message in a conversation with a completely new device (there should be a query to `POST users/list-prekeys` with all the clients that are member of the conversation
- [ ] sending a message in a conversation where one of the member has a new device (there should be a query to `POST /users/list-prekeys` with a single device)
- [ ] sending a **second** message in a conversation where we already has sent a message (there should not be any call to `POST /users/list-prekeys` the payload sent to `/proteus/message` should be a `20x` code and not `412`

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -76,17 +85,81 @@ export class ProteusService {
...
+  }
+
+  private async _createSessionsBatchedNonQualified(userClientMap: UserClients): Promise<string[]> {
+    const prekeyBundleMap = await this.apiClient.api.user.postMultiPreKeyBundles(userClientMap);
+    const sessions = await this._createSessionsWithBundleMap(prekeyBundleMap);
+
+    return sessions;
+  }
+
+  private async _createSessionsBatchedQualified(userClientMap: QualifiedUserClients): Promise<string[]> {
```
atomrc, line 137:
```suggestion
  private async createQualifiedSessions(userClientMap: QualifiedUserClients): Promise<string[]> {
```

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -76,17 +85,81 @@ export class ProteusService {
...
+        }
+
+        sessions.push(sessionId);
+      }
+    }
+
+    return sessions;
+  }
+
+  private async _createSessionsBatchedNonQualified(userClientMap: UserClients): Promise<string[]> {
```
atomrc, line 130:
```suggestion
  private async createSessions(userClientMap: UserClients): Promise<string[]> {
```

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -76,17 +85,81 @@ export class ProteusService {
...
     return this.coreCryptoClient.proteusFingerprintRemote(sessionId);
   }
 
-  private async createSession(
-    sessionId: string,
-    userId: QualifiedId,
-    clientId: string,
-    initialPrekey?: PreKey,
-  ): Promise<void> {
+  private async _createSessionsWithBundleMap(
```
atomrc, line 93:
```suggestion
  private async createSessionsFromPreKeys(
```

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -76,17 +85,81 @@ export class ProteusService {
...
+      const domainUsers = prekeyBundleMap[domain];
+
+      const domainSessions = await this._createSessionsWithBundleMap(domainUsers, domain);
+      sessions.push(...domainSessions);
+    }
+
+    return sessions;
+  }
+
+  private async createSessionsBatched(userClientMap: UserClients | QualifiedUserClients): Promise<string[]> {
```
atomrc, line 152:
```suggestion
  private async createSessions(userClientMap: UserClients | QualifiedUserClients): Promise<string[]> {
```

packages/core/src/test/ProteusHelper.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const coreCrypto = {
+    getRemoteFingerprint: jest.fn(),
+    getLocalFingerprint: jest.fn(),
+    proteusSessionExists: jest.fn(),
+    proteusSessionFromPrekey: jest.fn(),
+    proteusFingerprintRemote: jest.fn(),
+    proteusEncryptBatched: jest.fn(),
+  } as unknown as CoreCrypto;
+
```
atomrc, line 71: maybe you can add those to the global `__mock/@wireapp/core-crypto.ts` mock ;)

packages/core/src/util/preKeyBundleToUserClients/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export * from './preKeyBundleToUserClients';
```
atomrc, line 20: Since this is probably only used inside of the proteus namespace, maybe we could move it there? WDYT?

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -173,4 +179,47 @@ export class ProteusService {
...
+    domain: string = '',
+  ): Promise<{missing: UserClients; encrypted: OTRRecipients<Uint8Array>}> {
+    const {sessions, userClients} = await getSessionsAndClientsFromRecipients({
+      recipients,
+      domain,
+      apiClient: this.apiClient,
+      coreCryptoClient: this.coreCryptoClient,
+      cryptographyService: this.cryptographyService,
+      logger: this.logger,
+    });
```
atomrc, line 195: Humm I might be wrong, but I have the feeling that we are creating sessions for all the users everytime we encrypt something. This feels a little wrong here 🤔 

Creating session should happen only for the sessions that do not exist yet. (the `getSessionAndClientsFromRecipients` seems to never be checking if the session already exists)

PatrykBuniX: The check is inside `createSessionsFromPreKeys` method

packages/core/src/client/ClientService.ts
```diff
@@ -67,7 +63,7 @@ export class ClientService {
   public async deleteClient(clientId: string, password?: string): Promise<unknown> {
     const userId: QualifiedId = {id: this.apiClient.userId as string, domain: this.apiClient.domain || ''};
     await this.backend.deleteClient(clientId, password);
-    return this.database.deleteClient(this.cryptographyService.constructSessionId(userId, clientId));
+    return this.database.deleteClient(constructSessionId({userId, clientId}));
```
atomrc, line 66: This change is not equivalent since the `useQualifiedId` is going to be out of sync. 
You will need to inject the `useQualifiedId` in the class and in the `constructSessionId` as well

packages/core/src/messagingProtocols/proteus/Utility/SessionHandler/SessionHandler.ts
```diff
@@ -0,0 +1,258 @@
...
+
+import {SessionId} from './SessionHandler.types';
+
+import {isQualifiedUserClients, isUserClients} from '../../../../util';
+import {preKeyBundleToUserClients} from '../PreKeyBundle';
+
+interface ConstructSessionIdParams {
+  userId: string | QualifiedId;
+  clientId: string;
+  useQualifiedIds?: boolean;
```
atomrc, line 37: Let's make this one mandatory to avoid mis-generated sessionIds (and since this one is only supposed to be used internally)

packages/core/src/client/ClientService.ts
atomrc, line 66:
```suggestion
    return this.database.deleteClient(this.proteusService.constructSessionId({userId, clientId}));
```

packages/core/src/user/UserService.ts
```diff
@@ -26,7 +26,7 @@ import {Availability, GenericMessage} from '@wireapp/protocol-messaging';
 import {AvailabilityType, BroadcastService} from '../broadcast/';
 import {ConnectionService} from '../connection';
 import {createId} from '../conversation/message/MessageBuilder';
-import {getPreKeyBundleMap} from '../messagingProtocols/proteus/Utility/PreKeyBundle';
+import {getPreKeyBundleMap} from '../messagingProtocols/proteus/Utility/PreKeyBundle/PreKeyBundle';
```
atomrc, line 29:
```suggestion
import {getPreKeyBundleMap} from '../messagingProtocols/proteus/Utility/PreKeyBundle';
```

I think this should work since you have a `index.ts` in the root of PreKeyBundle

PR #4625, opened 2022-11-18
feat: use coreCrypto to decrypt proteus messages

aweiss-dev:
## Pull Request Checklist


packages/core/tsconfig.json
```diff
@@ -4,6 +4,6 @@
     "esModuleInterop": true,
     "outDir": "lib"
   },
-  "exclude": ["node_modules", "src/test", "src/**/*.test.ts", "src/demo"],
```
aweiss-dev, line 7: ?? Problem because we include files into the build

packages/core/src/messagingProtocols/proteus/Utility/constructSessionId/constructSessionId.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import {QualifiedId} from '@wireapp/api-client/lib/user';
+
+interface ConstructSessionIdParams {
+  userId: string | QualifiedId;
+  clientId: string;
+  useQualifiedIds?: boolean;
+  domain?: string;
+}
+const constructSessionId = ({userId, clientId, useQualifiedIds, domain}: ConstructSessionIdParams): string => {
```
atomrc, line 28: This needs to be publicly accessible somehow since the webapp uses this method

packages/core/src/messagingProtocols/proteus/CryptMessages/CryptMessages.ts
```diff
@@ -0,0 +1,78 @@
+/*
```
atomrc, line 1: CryptMessage, this name doesn't make sense. We need to find something better there

packages/core/src/messagingProtocols/proteus/CryptMessages/decrypt/decrypt.ts
```diff
@@ -0,0 +1,37 @@
...
+  sessionId: string;
+  encodedCiphertext: string;
+  logger: logdown.Logger;
+}
+
+const decrypt = ({coreCryptoClient, encodedCiphertext, logger, sessionId}: DecryptParams): Promise<Uint8Array> => {
+  logger.log(`Decrypting message for session ID "${sessionId}"`);
+  const messageBytes = Decoder.fromBase64(encodedCiphertext).asBytes;
+  return coreCryptoClient.proteusDecrypt(sessionId, messageBytes);
+};
```
atomrc, line 35: I don't see the value in splitting this in its own file and injecting all those params.

atomrc (changes requested):
Scenarios to test:
- [x] receiving a message from a fresh new device (the session doesn't exist already)
- [x] receiving a message from a user we already have sent a message to (the session already exists)

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -32,18 +31,20 @@ import {
...
 import {EventHandlerParams, handleBackendEvent} from '../EventHandler';
+import {constructSessionId} from '../Utility/constructSessionId';
+import {createSession} from '../Utility/createSession';
+import {getGenericMessageParams} from '../Utility/getGenericMessageParams';
+import {isClearFromMismatch} from '../Utility/isClearFromMismatch';
 
 export class ProteusService {
   private readonly messageService: MessageService;
-  private readonly logger = logdown('@wireapp/core/ProteusService');
+  private readonly logger = logdown('@wireapp/core/messagingProtocols/proteus/ProteusService');
```
atomrc, line 47: let's avoid long names in the logger, it doesn't need to match the file path

packages/core/src/messagingProtocols/proteus/Utility/GenericMessageMapper.ts
```diff
@@ -0,0 +1,275 @@
...
+
+export class GenericMessageMapper {
+  private static readonly logger = logdown('@wireapp/core/GenericMessageMapper', {
+    logger: console,
+    markdown: false,
+  });
+
+  // TODO: Turn "any" into a specific type (or collection of types) and make the return type more specific based on the
+  // "genericMessage" input parameter.
+  public static mapGenericMessage(
```
atomrc, line 51: I'm actually thinking of removing any trace of mapping on the core. 
IMO the core should only transmit raw events to the consumer and the consumer decides what it want to do with it. 

I'll prepare a breaking change PR, that removes all those mappings, but in the meantime please revert this change to avoid conflicts in the future (and actually this function doesn't seem to be used in your PR)

packages/core/src/messagingProtocols/proteus/DecryptMessages/DecryptMessages.ts
```diff
@@ -0,0 +1,71 @@
+/*
+ * Wire
```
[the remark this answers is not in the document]
aweiss-dev: That sounds reasonable. Thanks mate

packages/core/src/messagingProtocols/proteus/Utility/GenericMessageMapper.ts
aweiss-dev: Its used in otrMessageAdd.ts under messagingProtocols/proteus/EventHandler

packages/core/src/messagingProtocols/proteus/MessageDecryption/MessageDecryption.ts
```diff
@@ -0,0 +1,71 @@
...
+import {Decoder} from 'bazinga64';
+import logdown from 'logdown';
+
+import {APIClient} from '@wireapp/api-client';
+import {GenericMessage} from '@wireapp/protocol-messaging';
+
+import {constructSessionId} from '../Utility/constructSessionId/constructSessionId';
+import {generateDecryptionError} from '../Utility/generateDecryptionError/generateDecryptionError';
+
+const logger = logdown('@wireapp/core/messagingProtocols/proteus/CryptMessages');
```
atomrc, line 31:
```suggestion
const logger = logdown('@wireapp/core/MessageDecryption');
```

packages/core/src/messagingProtocols/proteus/Utility/constructSessionId/constructSessionId.ts
```diff
@@ -0,0 +1,35 @@
+/*
```
atomrc, line 1: Let's avoid action-type module naming. 
How about `SessionIdBuilder` rather than `constructSessionId`?

packages/core/src/messagingProtocols/proteus/Utility/generateDecryptionError/index.ts
```diff
@@ -0,0 +1,20 @@
+/*
```
atomrc, line 1: Let's avoid action-type module naming.
How about `DecryptionErrorGenerator` rather than `generateDecryptionError`?

packages/core/src/messagingProtocols/proteus/Utility/GenericMessageMapper.ts
aweiss-dev: Its gone

atomrc (approved):

PR #4628, opened 2022-11-21
test: Fix Account tests

atomrc:
Since `@wireapp/core-crypto` is a esmodule only, we cannot, as of now, import it in our unit test, we need to manually mock it


PatrykBuniX (approved):

packages/core/src/Account.test.ts
```diff
@@ -88,6 +91,12 @@ describe('Account', () => {
     user: 'aaf9a833-ef30-4c22-86a0-9adc8a15b3b4',
   };
 
+  beforeEach(() => {
+    // @ts-ignore
+    global.crypto = nodeCrypto.webcrypto;
+    global.window = {crypto: nodeCrypto.webcrypto, btoa: i => i};
```
atomrc, line 97: There might be a better way to do this 🤔

aweiss-dev: That should be done in a setup file. Same for the indexdb polyfill

aweiss-dev (approved):

PR #4629, opened 2022-11-21
fix: Handle case of empty list of users added to a conversation

atomrc:
When no users are added to an MLS conversation, a simple update to the keying material can be done.

### DIFF
--- a/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
+++ b/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
@@ -259,9 +259,9 @@ export class MLSService {
       ciphersuite: 1, // TODO: Use the correct ciphersuite enum.
     };
 
-    this.coreCryptoClient.createConversation(groupIdBytes, configuration);
+    await this.coreCryptoClient.createConversation(groupIdBytes, configuration);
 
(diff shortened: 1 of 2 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -274,10 +274,11 @@ export class MLSService {
...
     );
 
-    let response;
-    if (coreCryptoKeyPackagesPayload.length !== 0) {
-      response = await this.addUsersToExistingConversation(groupIdBytes, coreCryptoKeyPackagesPayload);
-    }
+    const response =
+      keyPackages.length > 0
+        ? await this.addUsersToExistingConversation(groupIdBytes, keyPackages)
+        : // If there are no clients to add, just update the keying material
```
V-Gira, line 280:
```suggestion
        : // If there are no clients to add, just update the keyring material
```

atomrc: Not a typo ;)

V-Gira: Yup, became clear after a little googling and looking at the rest of the comments. Learning bit by bit :)

PatrykBuniX (approved):

PR #4630, opened 2022-11-22
feat(react-ui-kit): Prepare Indicator Range Input

przemvs:
Created Range Input with indicators.

image

### DIFF
--- a/packages/react-ui-kit/src/Form/IndicatorRangeInput.styles.ts
+++ b/packages/react-ui-kit/src/Form/IndicatorRangeInput.styles.ts
@@ -84,7 +84,7 @@ export const headingStyle = (listLength: number, theme: Theme): CSSObject => {
 
   return {
     color: theme.general.color,
-    fontSize: '14px',
+    fontSize: theme.fontSizes.medium,
     fontWeight: '400',
     lineHeight: '1.2em',
     width: optionWidth,
@@ -105,7 +105,7 @@ export const dataListOption = (listLength: number, theme: Theme): CSSObject => {
 
   return {
     color: theme.Input.labelColor,
-    fontSize: '16px',
+    fontSize: '12px',
     fontWeight: '400',
     lineHeight: '1.2em',
 


V-Gira (approved):
Tested locally, good to go!

packages/react-ui-kit/src/Form/IndicatorRangeInput.md
```diff
@@ -0,0 +1,37 @@
...
+  {value: 20, label: '20px'},
+  {value: 24, label: '24px', heading: 'Large'},
+  {value: 30, label: '30px'},
+  {value: 36, label: '36px', heading: 'Huge'},
+];
+
+const [fontValue, setFontValue] = useState(3);
+
+const handleChange = e => {
+  setFontValue(+e.target.value);
```
arjita-mitra, line 22: why we need a plus here? the value in dataListOptions is already a number. Is there a case where you are expecting it can be a string?

packages/react-ui-kit/src/Form/IndicatorRangeInput.md
```diff
@@ -0,0 +1,37 @@
...
+};
+
+<>
+  <div>
+    <IndicatorRangeInput
+      label="Font size"
```
arjita-mitra, line 28: Shouldn't be hard coded here. Don't we need to translate this?

packages/react-ui-kit/src/Form/IndicatorRangeInput.md
```diff
@@ -0,0 +1,37 @@
...
+      onChange={handleChange}
+      value={fontValue}
+      min={0}
+      step={1}
```
arjita-mitra, line 32: better to use constants for this. we are using no magic number rule.

packages/react-ui-kit/src/Form/IndicatorRangeInput.styles.ts
```diff
@@ -0,0 +1,151 @@
...
+    backgroundRepeat: 'no-repeat',
+    display: 'flex',
+    alignItems: 'center',
+    borderRadius: '8px',
+    marginBlock: '10px',
+    zIndex: '1',
+    position: 'relative',
+
+    ...manySelectors(thumbSelectors, {
+      '-webkit-appearance': 'none',
```
arjita-mitra, line 56: It's usually considered good practice to specify the vendor-prefixed version first and then the non-prefixed version, in order that the non-prefixed property will override the vendor-prefixed property-settings once it's implemented;

according to that, appearance should come after '-webkit-appearance'. In this case it doesn't matter much because both are having the same value but in future we may change it.

packages/react-ui-kit/src/Form/IndicatorRangeInput.styles.ts
```diff
@@ -0,0 +1,151 @@
...
+    }),
+  };
+};
+
+export const headingStyle = (listLength: number): CSSObject => {
+  const optionWidth = `calc((100% - 12px) / ${listLength})`;
+
+  return {
+    color: COLOR_V2.BLACK,
+    fontSize: '14px',
```
arjita-mitra, line 86: we are using rem/em everywhere. Do we really need to use px here for font size?

packages/react-ui-kit/src/Form/IndicatorRangeInput.styles.ts
```diff
@@ -0,0 +1,151 @@
...
+    borderRadius: '8px',
+    marginBlock: '10px',
+    zIndex: '1',
+    position: 'relative',
+
+    ...manySelectors(thumbSelectors, {
+      '-webkit-appearance': 'none',
+      height: '28px',
+      width: '10px',
+      borderRadius: '4px',
```
arjita-mitra, line 59: Can you update px to rem/em wherever possible?

packages/react-ui-kit/src/Form/IndicatorRangeInput.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    const listLength = dataListOptions.length - 1;
+    const valueNum = Number(value);
+
+    const maxNum = isCustomSlider ? Number(listLength) : Number(max);
+
+    const backgroundSize = useMemo(() => {
+      if (isCustomSlider && valueNum === 0) {
+        const firstOptionThumbPosition = `calc(((100% - 10px) / (${listLength} * 2) - 4px) / 2)`;
+
```
arjita-mitra, line 71: Can you use const here for value 2

packages/react-ui-kit/src/Form/IndicatorRangeInput.tsx
```diff
@@ -0,0 +1,119 @@
...
+                <div key={dataListOption.value} css={() => headingStyle(listLength)}>
+                  {dataListOption?.heading}
+                </div>
+              ))}
+            </div>
+          )}
+
+          <input
+            ref={ref}
+            css={() => rangeStyles(backgroundSize, valueNum, listLength, isCustomSlider)}
```
arjita-mitra, line 95: Have you tested the slider while playing the screen reader? The label should announce the slider range. Using keyboard a user should be able to increase/decrease the value and the current value should be announced.

I have found some example that may help you.

https://codepen.io/smhigley/pen/ObWbdy
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/slider_role

arjita-mitra (changes requested):

packages/react-ui-kit/src/Form/IndicatorRangeInput.styles.ts
V-Gira: This is happening at the same time as my refactor, we need one of the pr to be merged before @przemvs can use the new variable, or I can change this component in my pr

PR #4637, opened 2022-11-24
fix: Address type mismatch for fowarded ref components

atomrc:


V-Gira (approved):
Double checked, other components with `forwardRef` are already using `FC`. Looks good to me :+1:

PR #4639, opened 2022-11-24
refactor: Remove all event mapping

atomrc:
Mapping should not be part of the responsibilities of the core. The consumer should receive raw events and decide what to do with it. 

BREAKING CHANGE: the `on(eventType)` event emitter doesn't exist anymore. You should now directly use  the `onEvent` callback from the `account.listen` method


packages/core/src/Account.ts
```diff
@@ -83,40 +82,6 @@ export enum ConnectionState {
...
-  on(event: PayloadBundleType.MESSAGE_HIDE, listener: (payload: OtrMessage.HideMessage) => void): this;
-  on(event: PayloadBundleType.PING, listener: (payload: OtrMessage.PingMessage) => void): this;
-  on(event: PayloadBundleType.REACTION, listener: (payload: OtrMessage.ReactionMessage) => void): this;
-  on(event: PayloadBundleType.TEXT, listener: (payload: OtrMessage.TextMessage) => void): this;
-  on(
-    event: PayloadBundleType.TIMER_UPDATE,
-    listener: (payload: Events.ConversationMessageTimerUpdateEvent) => void,
-  ): this;
-  on(event: PayloadBundleType.TYPING, listener: (payload: Events.ConversationTypingEvent) => void): this;
-  on(event: PayloadBundleType.UNKNOWN, listener: (payload: any) => void): this;
```
atomrc, line 119: This is legacy behavior. We now get all events from the `onEvent` callback of the `listen` method

atomrc:
closed in favor of https://github.com/wireapp/wire-web-packages/pull/4640

PR #4640, opened 2022-11-24
refactor: Remove all event mapping

atomrc:
Mapping should not be part of the responsibilities of the core. The consumer should receive raw events and decide what to do with it.

BREAKING CHANGE: the on(eventType) event emitter doesn't exist anymore. You should now directly use the onEvent callback from the account.listen method


aweiss-dev (approved):
I like this alot

PR #4643, opened 2022-11-29
refactor: Improve SessionHandler API

atomrc:


packages/core/src/conversation/message/MessageService.ts
```diff
@@ -83,10 +83,7 @@ export class MessageService {
       cipherText = externalPayload.cipherText;
     }
 
-    const {encrypted, missing} = await this.proteusService.encrypt(plainTextPayload, recipients);
-    const encryptedPayload = Object.keys(missing).length
-      ? await this.reencryptAfterMismatch({missing, deleted: {}}, encrypted, plainText)
-      : encrypted;
+    const encryptedPayload = await this.proteusService.encrypt(plainTextPayload, recipients);
```
atomrc, line 86: no more missing encrypted payload since all the necessary sessions will be created beforehand

PatrykBuniX, line 86: We don't care about `missing` anymore?

PatrykBuniX: nevermind, I see the comment https://github.com/wireapp/wire-web-packages/pull/4643/#discussion_r1034517331

atomrc: There won't be any missing anymore (since we handle the case where sessions are missing and create them straight away)

PatrykBuniX (approved):

PR #4645, opened 2022-11-29
feat(react-ui-kit): Indicator range input voiceover

przemvs:
## Pull Request Checklist

### DIFF
--- a/packages/react-ui-kit/src/Form/IndicatorRangeInput.tsx
+++ b/packages/react-ui-kit/src/Form/IndicatorRangeInput.tsx
@@ -17,7 +17,7 @@
  *
  */
 
-import {FC, forwardRef, useMemo} from 'react';
+import {FC, HTMLProps, PropsWithRef, forwardRef, useMemo} from 'react';
 
 import {CSSObject} from '@emotion/react';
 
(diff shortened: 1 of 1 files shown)


V-Gira (approved):

packages/react-ui-kit/src/Form/IndicatorRangeInput.tsx
```diff
@@ -76,15 +78,35 @@ export const IndicatorRangeInput: FC<IndicatorRangeInputProps> = forwardRef<
       return `${((valueNum - minNum) * 100) / (maxNum - minNum)}% 100%`;
     }, [isCustomSlider, valueNum, minNum, maxNum, listLength]);
 
+    const currentSelectedValue = useMemo(() => {
+      if (typeof value !== 'string' && typeof value !== 'number') {
```
arjita-mitra, line 82: isStringOrNumber can be used here?

arjita-mitra (changes requested):

packages/react-ui-kit/src/Form/IndicatorRangeInput.tsx
PatrykBuniX: Why this check if even needed here? Can't we just explicitly set prop type to be a number? I wonder if `extends TextProps<T>` is needed I feel like we're losing some control here due to very wide prop types. Wdyt?

V-Gira: can't we use `valueNum` declared earlier instead of `value`?

PatrykBuniX (approved):

arjita-mitra (approved):

PR #4647, opened 2022-11-30
feat: trigger migration from cryptobox to corecrypto [FS-1095]

PatrykBuniX:
Trigger `cryptobox` -> `corecrypto` migration script after initialising proteus client.

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -88,6 +88,10 @@ export interface Account {
 
 export type CreateStoreFn = (storeName: string, context: Context) => undefined | Promise<CRUDEngine | undefined>;
 
+export interface DBMigrationHooks {
+  getState: () => {storeName: string} | undefined;
+  onSuccess: () => void;
+}
 interface AccountOptions<T> {
   /** Used to store info in the database (will create a inMemory engine if returns undefined) */
   createStore?: CreateStoreFn;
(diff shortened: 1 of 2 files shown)


packages/core/src/Account.ts
```diff
@@ -313,24 +323,28 @@ export class Account<T = any> extends EventEmitter {
...
    * @param clientInfo Will allow creating the client if the local client cannot be found (else will fail if local client is not found)
    * @param entropyData Additional entropy data
+   * @param dbMigrationHooks Cryptobox to CoreCrypto migration config object
    * @returns The local existing client or newly created client
    */
   public async initClient(
     loginData: LoginData,
     clientInfo?: ClientInfo,
     entropyData?: Uint8Array,
+    dbMigrationHooks?: DBMigrationHooks,
```
atomrc, line 338: Since all those are optional, let's gather them in a `option` object param (this will avoid the `initClient(info, undefined, undefined, migrationHooks)`)

atomrc:
Acceptance criteria (scenarios to tests):
- [ ] sending from a old device to people I have session with:
  - use a device from current `dev` branch
  - send a bunch of messages in a conversation
  - logout from this device
  - switch to your feature branch 
  - reload the webapp
  => the migration should happen at this point (there should be data in `corecrypto` database on the `proteus_identities` table)
  => you should be able to send messages and the other person should be able to decrypt them

- [ ] receiving from a old device
  - use a device from current `dev` branch
  - send a bunch of messages in a conversation
  - logout from this device
  - send a bunch of messages from a user that is connected to the user that will reload the webapp
  - switch to feature branch
  - reload the webapp
  => you should see the message that were sent while the user was offline

atomrc (approved):

PR #4650, opened 2022-11-30
refactor: Remove traces of cryptobox

atomrc:


aweiss-dev (approved):
Is Bazinga64 used somewhere else or can it be retired?

atomrc:
> Is Bazinga64 used somewhere else or can it be retired?

yes, it's still used in the core (for converting data to and from coreCrypto)

PR #4683, opened 2022-12-05
fix: update asset api (FS-1001)

tlebon:
Add error case for if v3 is called without domain

### DIFF
--- a/packages/api-client/src/asset/AssetAPI.ts
+++ b/packages/api-client/src/asset/AssetAPI.ts
@@ -291,9 +291,10 @@ export class AssetAPI {
     if (!isValidUUID(assetId)) {
       throw new TypeError(`Expected asset ID "${assetId}" to only contain alphanumeric values and dashes.`);
     }
-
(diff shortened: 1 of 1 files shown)


packages/api-client/src/asset/AssetAPI.ts
```diff
@@ -291,9 +291,10 @@ export class AssetAPI {
     if (!isValidUUID(assetId)) {
       throw new TypeError(`Expected asset ID "${assetId}" to only contain alphanumeric values and dashes.`);
     }
-
-    const assetBaseUrl = this.backendFeatures.version >= 2 ? ASSET_URLS.ASSETS_URL : ASSET_URLS.ASSET_V3_URL;
-    return this.getAssetShared(`${assetBaseUrl}/${assetId}`, token, forceCaching, progressCallback);
+    if (this.backendFeatures.version >= 2) {
```
arjita-mitra, line 294: make 2 as const? Avoid magic numbers.

arjita-mitra (changes requested):

atomrc (approved):

PR #4693, opened 2022-12-06
feat(api-client): api v3 deprecated endpoints [FS-1268]

PatrykBuniX:
Remove deprecated endpoints/fields due to api v3 updates (backwards compatibility):
- [x] The deprecated GET /conversations/ids endpoint was removed.
- [x] The deprecated endpoints GET /conversations/ids and GET /conversations have been removed. Use POST /conversations/list-ids followed by POST /conversations/list instead.
- [x] The endpoint PUT /conversations/:id/access has been removed. Use its qualified counterpart instead.
- [x] The field access_role_v2 in the Conversation type, in the request body of POST /conversations, and in the request body of PUT /conversations/:domain/:id/access has been removed. Its content is now contained in the access_role field instead. It replaces the legacy access role, previously contained in the access_role field.

### DIFF
--- a/packages/api-client/src/conversation/ConversationAPI/ConversationAPI.ts
+++ b/packages/api-client/src/conversation/ConversationAPI/ConversationAPI.ts
@@ -85,7 +85,6 @@ export class ConversationAPI {
     CODE_CHECK: '/code-check',
     CONVERSATIONS: '/conversations',
     MLS: '/mls',
-    IDS: 'ids',
     JOIN: '/join',
     LIST: 'list',
     LIST_IDS: 'list-ids',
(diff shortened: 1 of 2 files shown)


packages/api-client/src/conversation/ConversationAPI/ConversationAPI.ts
```diff
@@ -890,6 +867,34 @@ export class ConversationAPI {
    * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/conversations/updateConversationAccess
    */
   public async putAccess(
+    conversationId: QualifiedId | string,
```
atomrc, line 870: Let's force the consumer to use qualified ids from now on. 
```suggestion
    conversationId: QualifiedId,
```

atomrc: Actually, since only the url changes but the rest is exactly the same, let's just have a small helper that generates qualified/non-qualified base urls. 

```ts
  private generateBaseConversationUrl(conversationId: QualifiedId): string {
    return this.backendFeatures.federationEndpoints
      ? `${ConversationAPI.URL.CONVERSATIONS}/${conversationId.domain}/${conversationId.id}`
      : `${ConversationAPI.URL.CONVERSATIONS}/${conversationId.id}`;
  }
```

and we can use this helper everywhere we need to support both unqualified and qualified urls. WDYT?

packages/api-client/src/conversation/ConversationAPI/ConversationAPI.ts
```diff
@@ -282,27 +286,6 @@ export class ConversationAPI {
...
   }
 
-  /**
-   * Get all conversation IDs.
-   * @param limit Max. number of IDs to return
-   * @param conversationId Conversation ID to start from (exclusive)
-   * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/conversations/conversationIds
-   * @deprecated Use `getListConversations()` instead.
-   */
-  public async getConversationIds(limit: number, conversationId?: string): Promise<ConversationIds> {
```
PatrykBuniX, line 292: doesn't seem to be used anywhere

packages/api-client/src/conversation/data/ConversationAccessUpdateData.ts
```diff
@@ -22,7 +22,7 @@ import {CONVERSATION_ACCESS, CONVERSATION_LEGACY_ACCESS_ROLE, CONVERSATION_ACCES
 /**@deprecated */
 export interface ConversationAccessV2UpdateData {
   access: CONVERSATION_ACCESS[];
-  access_role: CONVERSATION_LEGACY_ACCESS_ROLE;
+  access_role?: CONVERSATION_LEGACY_ACCESS_ROLE;
```
PatrykBuniX, line 25: this field was not required in v2 (we were always sending data in `access_role_v2` field)

atomrc (approved):

PR #4696, opened 2022-12-08
feat: api v3 support for typing endpoint [FS-1311]

PatrykBuniX:
Add support for `/conversations/{domain}/{id}/typing` endpoint in api v3.

### DIFF
--- a/packages/api-client/src/conversation/ConversationAPI/ConversationAPI.ts
+++ b/packages/api-client/src/conversation/ConversationAPI/ConversationAPI.ts
@@ -104,7 +104,7 @@ export class ConversationAPI {
   constructor(protected readonly client: HttpClient, protected readonly backendFeatures: BackendFeatures) {}
 
   private generateBaseConversationUrl(conversationId: QualifiedId): string {
-    return this.backendFeatures.federationEndpoints
+    return this.backendFeatures.federationEndpoints && conversationId.domain
       ? `${ConversationAPI.URL.CONVERSATIONS}/${conversationId.domain}/${conversationId.id}`
       : `${ConversationAPI.URL.CONVERSATIONS}/${conversationId.id}`;
   }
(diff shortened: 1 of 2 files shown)


packages/api-client/src/conversation/ConversationAPI/ConversationAPI.ts
```diff
@@ -104,7 +104,7 @@ export class ConversationAPI {
   constructor(protected readonly client: HttpClient, protected readonly backendFeatures: BackendFeatures) {}
 
   private generateBaseConversationUrl(conversationId: QualifiedId): string {
-    return this.backendFeatures.federationEndpoints
+    return this.backendFeatures.federationEndpoints && conversationId.domain
       ? `${ConversationAPI.URL.CONVERSATIONS}/${conversationId.domain}/${conversationId.id}`
       : `${ConversationAPI.URL.CONVERSATIONS}/${conversationId.id}`;
```
PatrykBuniX, line 109: this way we're completely sure that in case of empty string for `domain` we won't get url like `${ConversationAPI.URL.CONVERSATIONS}/${''}/${conversationId.id}` => `/conversations}//conversation-id`

packages/api-client/src/conversation/ConversationAPI/ConversationAPI.ts
```diff
@@ -856,11 +856,11 @@ export class ConversationAPI {
    * @param typingData The typing status
    * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/conversations/isTyping
    */
-  public async postTyping(conversationId: string, typingData: ConversationTypingData): Promise<void> {
+  public async postTyping(conversationId: QualifiedId, typingData: ConversationTypingData): Promise<void> {
     const config: AxiosRequestConfig = {
       data: typingData,
       method: 'post',
-      url: `${ConversationAPI.URL.CONVERSATIONS}/${conversationId}/${ConversationAPI.URL.TYPING}`,
+      url: `${this.generateBaseConversationUrl(conversationId)}/${ConversationAPI.URL.TYPING}`,
```
atomrc, line 863: Wait, I'm not sure this endpoint will work with api v2?

atomrc (changes requested):

packages/api-client/src/conversation/ConversationAPI/ConversationAPI.ts
atomrc: Ok, seems to be fine with backend, the endpoint is going to be available for v2 as well, all good then 👌

PatrykBuniX: Yes, we should be in sync with what's documented in swagger 👍

PatrykBuniX:
Can confirm that after last change, the client is hitting non-qualified endpoint when using v2 😄

atomrc (approved):

PR #4697, opened 2022-12-09
fix: improve useTimeout hook with ref

tlebon:
## Pull Request Checklist


packages/react-ui-kit/src/Misc/useTimeout.ts
```diff
@@ -17,21 +17,21 @@
...
 
 const useTimeout = (onTimeoutStop: () => void, duration: number) => {
-  const [timerId, setTimerId] = useState(null);
+  const timerRef = useRef<number | null>(null);
 
   const removeTimeout = () => {
-    if (timerId) {
-      window.clearTimeout(timerId);
-      setTimerId(null);
+    if (timerRef) {
```
PatrykBuniX, line 26: Should be timerRef.current

PatrykBuniX (changes requested):

packages/react-ui-kit/src/Misc/useTimeout.ts
PatrykBuniX: Won't timerRef always be defined since it's object?

aweiss-dev: As far as I know they check for deep equality. So if current changes it should be recognized. But we can be more precise here and go for .current like you mentioned

aweiss-dev (changes requested):
Like Patrik mentioned

tlebon:
pushed ref correction

PatrykBuniX (approved):

PR #4706, opened 2022-12-12
feat(core): recover mls conversation from quick sync 404 error [FS-1172]

PatrykBuniX:
Adds logic for `MLS` conversation to recover from `404` error in quick sync flow.

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -637,6 +637,13 @@ export class Account<T = any> extends EventEmitter {
       }
     });
 
+    const handleMissedNotifications = async (notificationId: string) => {
+      if (this.cryptoProtocolConfig?.mls && this.backendFeatures.supportsMLS) {
+        await this.service?.conversation.handleEpochMissmatch();
+      }
+      return onMissedNotifications(notificationId);
+    };
+
     const processNotificationStream = async (abortHandler: AbortHandler) => {
       // Lock websocket in order to buffer any message that arrives while we handle the notification stream
       this.apiClient.transport.ws.lock();
@@ -648,7 +655,7 @@ export class Account<T = any> extends EventEmitter {
           await handleNotification(notification, source);
           onNotificationStreamProgress(progress);
         },
-        onMissedNotifications,
+        handleMissedNotifications,
         abortHandler,
       );
       this.logger.log(`Finished processing notifications ${JSON.stringify(results)}`, results);
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -639,7 +639,7 @@ export class Account<T = any> extends EventEmitter {
 
     const handleMissedNotifications = async (notificationId: string) => {
       if (this.cryptoProtocolConfig?.mls && this.backendFeatures.supportsMLS) {
-        await this.service?.conversation.handleEpochMissmatch();
+        await this.service?.conversation.handleEpochMismatch();
       }
       return onMissedNotifications(notificationId);
     };
--- a/packages/core/src/conversation/ConversationService/ConversationService.test.ts
+++ b/packages/core/src/conversation/ConversationService/ConversationService.test.ts
@@ -176,19 +176,6 @@ describe('ConversationService', () => {
       } as Conversation;
     };
 
-    it('tries to re-join when not established yet conversation found', async () => {
-      const [conversationService, {apiClient}] = buildConversationService();
-
-      const mlsConversation = createConversation(1);
-      const mockedDBResponse: Conversation[] = [mlsConversation];
-      jest.spyOn(apiClient.api.conversation, 'getConversationList').mockResolvedValueOnce({found: mockedDBResponse});
-
-      jest.spyOn(conversationService, 'isMLSConversationEstablished').mockResolvedValueOnce(false);
-
-      await conversationService.handleEpochMismatch();
-      expect(conversationService.joinByExternalCommit).toHaveBeenCalledWith(mlsConversation.qualified_id);
-    });
-
     it('re-joins multiple not-established conversations', async () => {
       const [conversationService, {apiClient}] = buildConversationService();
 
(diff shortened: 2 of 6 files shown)


packages/core/src/Account.ts
```diff
@@ -637,6 +637,13 @@ export class Account<T = any> extends EventEmitter {
       }
     });
 
+    const handleMissedNotifications = async (notificationId: string) => {
+      if (this.cryptoProtocolConfig?.mls && this.backendFeatures.supportsMLS) {
+        await this.service?.conversation.handleEpochMissmatch();
```
atomrc, line 642:
```suggestion
        await this.service?.conversation.handleEpochMismatch();
```

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -403,4 +404,38 @@ export class ConversationService {
   public async wipeMLSConversation(groupId: Uint8Array): Promise<void> {
     return this.mlsService.wipeConversation(groupId);
   }
+
+  public isMLSConversation(conversation: Conversation): conversation is MLSConversation {
+    const {protocol, epoch, group_id} = conversation;
+    return protocol === ConversationProtocol.MLS && epoch !== undefined && group_id !== undefined;
+  }
+
+  public async handleEpochMissmatch() {
```
atomrc, line 413:
```suggestion
  public async handleEpochMismatch() {
```

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -403,4 +406,48 @@ export class ConversationService {
   public async wipeMLSConversation(groupId: Uint8Array): Promise<void> {
     return this.mlsService.wipeConversation(groupId);
   }
+
+  public async handleEpochMismatch() {
+    this.logger.info(`There were some missed messages, handling possible epoch missmatch in MLS conversations.`);
```
atomrc, line 411:
```suggestion
    this.logger.info(`There were some missed messages, handling possible epoch mismatch in MLS conversations.`);
```

PatrykBuniX: I always missssss correct spelling here 😅

packages/core/src/messagingProtocols/mls/types.ts
```diff
@@ -74,3 +75,9 @@ export interface CryptoProtocolConfig<T = any> {
   /** if set to true, will use experimental proteus encryption/decryption library (core-crypto). If not set will fallback to the legacy proteus library (cryptobox) */
   proteus?: boolean;
 }
+
+export type MLSConversation = Conversation & {
```
atomrc, line 79: 👌

packages/core/src/conversation/ConversationService/ConversationService.test.ts
```diff
@@ -154,6 +162,98 @@ describe('ConversationService', () => {
...
+
+      const mlsConversation = createConversation(1);
+      const mockedDBResponse: Conversation[] = [mlsConversation];
+      jest.spyOn(apiClient.api.conversation, 'getConversationList').mockResolvedValueOnce({found: mockedDBResponse});
+
+      jest.spyOn(conversationService, 'isMLSConversationEstablished').mockResolvedValueOnce(false);
+
+      await conversationService.handleEpochMismatch();
+      expect(conversationService.joinByExternalCommit).toHaveBeenCalledWith(mlsConversation.qualified_id);
+    });
```
atomrc, line 190: I think this test is not necessary as it's also covered by the following test `'re-joins multiple not-established conversations'`

packages/core/src/conversation/ConversationService/ConversationService.test.ts
```diff
@@ -154,6 +162,98 @@ describe('ConversationService', () => {
...
+
+      const mockedDBResponse: Conversation[] = [mlsConversation];
+      jest.spyOn(apiClient.api.conversation, 'getConversationList').mockResolvedValueOnce({found: mockedDBResponse});
+
+      jest.spyOn(conversationService, 'isMLSConversationEstablished').mockResolvedValueOnce(true);
+      jest.spyOn(mlsService, 'getEpoch').mockResolvedValueOnce(2);
+
+      await conversationService.handleEpochMismatch();
+      expect(conversationService.joinByExternalCommit).toHaveBeenCalledWith(mlsConversation.qualified_id);
+    });
```
atomrc, line 221: Also this one is redundant with the next one

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -403,4 +406,48 @@ export class ConversationService {
...
+        const isEpochMismatched = BigInt(clientStoredEpoch) !== BigInt(epoch);
+
+        //if there's no match -> try to rejoin
+        if (isEpochMismatched) {
+          this.logger.info(
+            `MLS conversation with id ${
+              qualifiedId.id
+            } was found in the database, epoch mismatch detected {client: ${String(
+              clientStoredEpoch,
+            )}, backend: ${String(epoch)}}, joining via external commit`,
```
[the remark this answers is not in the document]
PatrykBuniX: Cool 😄

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -403,4 +406,48 @@ export class ConversationService {
...
+      for (const {qualified_id: qualifiedId, group_id: groupId, epoch} of mlsConversations) {
+        const isEstablished = await this.isMLSConversationEstablished(groupId);
+        if (!isEstablished) {
+          this.logger.info(
+            `MLS conversation with id ${qualifiedId.id} was found in the database, but not established yet, joining via external commit`,
+          );
+          //if conversation was found bot not yet established, join
+          await this.joinByExternalCommit(qualifiedId);
+          continue;
+        }
```
atomrc, line 430: I wonder if we could simplify the logic a tiny bit. Since we always want to call `joinByExternalCommit` whether the conversation is a mismatch OR is not established. 

I'm wondering: what does `getEpoch` returns if the conversation is not established? Does it return some special value (like -1) or does it throw?

PatrykBuniX: Good point, I'll check it out 👍

atomrc: we could probably do something along those lines

```js
if (!isEstablished(conv) || isEpochMismatch(conv, newEpoch) {
  joinByExternal(conv)
}
```
WDYT?

PatrykBuniX: It seems that `getEpoch` throws an error when it cannot find conversation with given `groupId`

PR #4730, opened 2022-12-22
fix: Correctly init new mls client

atomrc:
## Pull Request Checklist

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -62,7 +62,6 @@ import {CoreDatabase, deleteDB, openDB} from './storage/CoreDB';
 import {TeamService} from './team/';
 import {UserService} from './user/';
 import {createCustomEncryptedStore, createEncryptedStore, EncryptedStore} from './util/encryptedStore';
-import {constructFullyQualifiedClientId} from './util/fullyQualifiedClientIdUtils';
 
 export type ProcessedEventPayload = HandledEventPayload;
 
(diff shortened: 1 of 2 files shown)


packages/core/src/Account.ts
```diff
@@ -477,10 +474,13 @@ export class Account<T = any> extends EventEmitter {
...
 
-    if (createMlsClient && this.backendFeatures.supportsMLS) {
+    if (createMlsClient) {
       const {userId, domain} = this.apiClient.context;
       const clientId = constructFullyQualifiedClientId(userId, registeredClient.id, domain || '');
-      await this.service.mls.initClient(clientId);
+      const {publicKey, keyPackages} = await this.service.mls.createClient(clientId);
+      // If the device is new, we need to upload keypackages and public key to the backend
+      await this.service.mls.uploadMLSPublicKeys(publicKey, registeredClient.id);
+      await this.service.mls.uploadMLSKeyPackages(keyPackages, registeredClient.id);
```
atomrc, line 483: Actually I could hide this inside of the `createClient` method :)

packages/core/src/Account.ts
```diff
@@ -296,7 +296,11 @@ export class Account<T = any> extends EventEmitter {
       //call /access endpoint with client_id after client initialisation
       await this.apiClient.transport.http.associateClientWithSession(localClient.id);
 
-      if (this.cryptoProtocolConfig?.mls && this.backendFeatures.supportsMLS) {
+      await this.service.proteus.init();
+      if (this.backendFeatures.supportsMLS && this.cryptoProtocolConfig?.mls) {
+        const {userId, domain} = this.apiClient.context;
+        const clientId = constructFullyQualifiedClientId(userId, localClient.id, domain || '');
```
PatrykBuniX, line 302: not sure about this `domain || ''`, what should we do if domain is not defined? throw?

atomrc: The case of the empty domain should never happen on any of our env. I think it's fine to keep it like this for now

PatrykBuniX (approved):

PR #4734, opened 2022-12-28
fix: Always run cryptobox to corecrypto migration before proteus init

atomrc:
As of now, there is a possibility to login to an existing device but get a new proteus identity. 
This PR moves all the migration logic to the core in order to trigger that migration before **any** call to `proteus.init` is done


packages/core/src/Account.ts
```diff
@@ -228,126 +222,99 @@ export class Account<T = any> extends EventEmitter {
...
-      const notFoundInDatabase =
-        error instanceof StoreEngineError.RecordNotFoundError ||
-        (error as Error).constructor.name === StoreEngineError.RecordNotFoundError.name;
-      const notFoundOnBackend = axios.isAxiosError(error) ? error.response?.status === HTTP_STATUS.NOT_FOUND : false;
-
-      if (notFoundInDatabase) {
-        this.logger.log(`Could not find valid client in database "${this.storeEngine?.storeName}".`);
-        return this.registerClient(loginData, clientInfo, entropyData);
-      }
+    //call /access endpoint with client_id after client initialisation
```
tlebon, line 300:
```suggestion
    // Call /access endpoint with client_id after client initialisation
```

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -266,7 +277,34 @@ export class ProteusService {
...
+    // We want sessions to be fully qualified from now on
+    migrateToQualifiedSessionIds(storeEngine, context.domain ?? '');
+
+    this.logger.log(`Migrating data from cryptobox store (${dbName}) to corecrypto.`);
+    try {
+      await this.coreCryptoClient.proteusCryptoboxMigrate(dbName);
+
+      // We can clear 3 stores (keys - local identity, prekeys and sessions) from wire db.
+      // They will be stored in corecrypto database now.
+      /* TODO uncomment this code when we are sure migration for wire.com has happened successfully for enough users
```
tlebon, line 296: are we creating a follow-up ticket or some tracking for this topic?

atomrc: Created 👍

tlebon (approved):

PR #4735, opened 2022-12-29
fix: Improve granularity of coreCrypto decryption error handling

atomrc:
We do not have perfect handling for decryption error code on coreCrypto side, but we can still handle a bunch of different scenarios that will improve user facing errors on the front-end

### DIFF
--- a/packages/core/src/messagingProtocols/proteus/ProteusService/DecryptionErrorGenerator/DecryptionErrorGenerator.ts
+++ b/packages/core/src/messagingProtocols/proteus/ProteusService/DecryptionErrorGenerator/DecryptionErrorGenerator.ts
@@ -21,9 +21,6 @@ import {QualifiedId} from '@wireapp/api-client/lib/user';
 
 import {DecryptionError} from '../../../../errors/DecryptionError';
 
-type ErrorWithCode = Error & {code?: number};
-const hasErrorCode = (error: any): ErrorWithCode => error && error.code;
-
 export const ProteusErrors = {
   InvalidMessage: 201,
   RemoteIdentityChanged: 204,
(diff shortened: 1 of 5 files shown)


packages/core/src/messagingProtocols/proteus/ProteusService/DecryptionErrorGenerator/DecryptionErrorGenerator.ts
```diff
@@ -18,46 +18,60 @@
  */
 
 import {QualifiedId} from '@wireapp/api-client/lib/user';
-import logdown from 'logdown';
 
 import {DecryptionError} from '../../../../errors/DecryptionError';
 
 type ErrorWithCode = Error & {code?: number};
 const hasErrorCode = (error: any): ErrorWithCode => error && error.code;
```
tlebon, line 25: i know this isnt something you changed but i find this function and its usage sort of confusing. in my opinion it should either be a boolean or it shouldn't be used in ternaries like on line 45. wdyt?

packages/core/src/messagingProtocols/proteus/ProteusService/DecryptionErrorGenerator/DecryptionErrorGenerator.ts
```diff
@@ -18,46 +18,60 @@
...
-  const isInvalidSignature = false; // TODO when coreCrypto returns ganular errors
-  const isRemoteIdentityChanged = false; // TODO when coreCrypto returns ganular errors
-  // Session is broken, let's see what's really causing it...
-  if (isInvalidMessage || isInvalidSignature) {
-    message = `Session with user '${remoteUserId}' (${remoteClientId}) is broken.\nReset the session for possible fix.`;
-  } else if (isRemoteIdentityChanged) {
-    message = `Remote identity of client '${remoteClientId}' from user '${remoteUserId}' changed`;
-  }
+const mapCoreCryptoError = (error: any): ProteusErrorCode => {
+  const code = hasErrorCode(error) ? error.code : CoreCryptoErrorMapping[error.message as CoreCryptoErrors];
```
tlebon, line 45: ie: this seems more correct (referencing earlier comment)
```suggestion
  const code = hasErrorCode(error) ?? CoreCryptoErrorMapping[error.message as CoreCryptoErrors];
```

tlebon: i personally prefer it returning a boolean but if we would like to keep the current function i think this is nicer. (i would rename the function in that case however)

tlebon (approved):

PR #4745, opened 2023-01-04
feat(api-client): base for subconversations and conference calls [FS-1408]

PatrykBuniX:
- adds subconversation requests' handlers 
- adds return data type


atomrc (approved):
Let's kick it off 🚀

PR #4773, opened 2023-01-11
feat: Have both CoreCrypto and Cryptobox as crypto client options

atomrc:
Will add the possibility to chose which crypto system should be used to handle encrypted messages. 
By default it will use `Cryptobox` but `CoreCrypto` can be chosen by using the parameter `useCoreCrypto`  in the Account's constructor config. 

## Design choices

Since we need to handle both `Cryptobox` and `CoreCrypto` and they both have a very similar API, we decided to go for a wrapping interface that would expose a common API and wrapper both Cryptobox and CoreCrypto into 2 implementation classes. 

This way all the logic to handle encrypted messages, prekeys... is hidden inside those classes


packages/core/src/Account.ts
```diff
@@ -301,7 +297,7 @@ export class Account<T = any> extends EventEmitter {
     await this.apiClient.transport.http.associateClientWithSession(validClient.id);
 
     await this.service.proteus.initClient(this.storeEngine, this.apiClient.context);
-    if (this.backendFeatures.supportsMLS && this.cryptoProtocolConfig?.mls) {
+    if (this.service.mls) {
```
atomrc, line 300: from now on, we know that if the mls service has been instantiated we are able to use MLS everywhere. No need to check for the config or the backend abilities anymore

packages/core/package.json
```diff
@@ -14,6 +14,7 @@
     "@wireapp/api-client": "workspace:^",
     "@wireapp/commons": "workspace:^",
     "@wireapp/core-crypto": "0.6.0-rc.3",
+    "@wireapp/cryptobox": "^12.8.0",
```
PatrykBuniX, line 17: Don't we want to use exact version?

atomrc: good call 👌

packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper.ts
```diff
@@ -0,0 +1,126 @@
...
+  deleteSession(sessionId: string) {
+    return this.coreCrypto.proteusSessionDelete(sessionId);
+  }
+
+  consumePrekey() {
+    return this.prekeyTracker.consumePrekey();
+  }
+
+  async newPrekey(id: number) {
+    // no need to generate prekey for cryptobox as they are generate internally
```
PatrykBuniX, line 104: This comment is not relevant, probably a leftover after splitting two apis

PatrykBuniX (approved):
Nice 🚀

PR #4783, opened 2023-01-12
refactor: Lazy load crypto clients 

atomrc:
This will avoid loading both CoreCrypto and Cryptobox in the same instance of the webapp (since only one of them is going to be used at a time).

This also cleans up which components are aware of CoreCrypto/Cryptobox. Now only the `CryptoClient` are aware of those classes and know how to build them

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -342,7 +342,7 @@ export class Account<T = any> extends EventEmitter {
     return key;
   }
 
-  private async buildCryptoClient(storeEngine: CRUDEngine, enableMLS: boolean) {
+  private async buildCryptoClient(context: Context, storeEngine: CRUDEngine, db: CoreDatabase, enableMLS: boolean) {
     const clientType =
       enableMLS || !!this.cryptoProtocolConfig?.useCoreCrypto
         ? CryptoClientType.CORE_CRYPTO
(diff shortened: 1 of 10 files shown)


packages/core/src/Account.ts
```diff
@@ -434,7 +442,6 @@ export class Account<T = any> extends EventEmitter {
    */
   public async logout(clearData: boolean = false): Promise<void> {
     this.db?.close();
-    await this.secretsDb?.close();
```
atomrc, line 437: The secret DB can be closed as soon as the key has been generated/fetched for CoreCrypto, no need to wait that long with the DB open. 
It's now closed here https://github.com/wireapp/wire-web-packages/pull/4783/files#diff-0a7e6d4a8ee3b5e3153dbc2336fecd1741ea0f39e7226d6b7a5d63be31b2747fR330

packages/core/src/Account.ts
```diff
@@ -443,9 +450,7 @@ export class Account<T = any> extends EventEmitter {
   }
 
   private async wipe(): Promise<void> {
-    if (this.coreCryptoClient) {
-      await this.coreCryptoClient.wipe();
-    }
+    await this.service?.proteus.wipe();
```
atomrc, line 453: Now the proteus service will call the `wipe` method of the CryptoClientWrapper

PatrykBuniX (approved):

PR #4789, opened 2023-01-17
fix: Do call onNewPrekeys handler on cryptobox wrapper

atomrc:
We currently do not upload new prekeys to the backend when they are generated. This fixes it


PatrykBuniX (approved):
Hard to spot 👍

aweiss-dev (approved):

PR #4806, opened 2023-01-23
refactor: Simplify MLS message handling

atomrc:


packages/core/src/messagingProtocols/mls/EventHandler/events/messageAdd/messageAdd.ts
```diff
@@ -70,9 +68,7 @@ const handleMLSMessageAdd = async ({mlsService, event, logger}: HandleMLSMessage
   }
 
   if (!message) {
-    const newEpoch = await mlsService.getEpoch(groupIdBytes);
-    logger.log(`Received commit message for group "${groupId}". New epoch is "${newEpoch}"`);
-    return;
+    return {event};
```
atomrc, line 71: This will be reimplemented in a better way in a follow up PR

V-Gira (approved):

packages/core/src/messagingProtocols/mls/EventHandler/events/messageAdd/messageAdd.ts
V-Gira: That's the only change right? The rest is just refactorin I'm not missing anything?

atomrc: No, you are right, the rest is only refactoring :)

PR #4808, opened 2023-01-23
feat: expose get subconversation in conversation service

PatrykBuniX:
- Exposes getSubconversation to `ConversationService`, 
- changes key bytesize to 32,
- improves exportsecretkey method

### DIFF
--- a/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
+++ b/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
@@ -234,14 +234,19 @@ export class MLSService {
     storeSubconversationGroupId(conversationId, subconversation.subconv_id, subconversation.group_id);
 
     const keyLength = 32;
-    const groupIdBytes = Decoder.fromBase64(subconversation.group_id).asBytes;
-    const secretKey = await this.coreCryptoClient.exportSecretKey(groupIdBytes, keyLength);
+    const secretKey = await this.exportSecretKey(subconversation.group_id, keyLength);
 
     return {
       subconversation,
-      secretKey: Encoder.toBase64(secretKey).asString,
+      secretKey,
       keyLength,
     };
+  }
+
+  public async exportSecretKey(groupId: string, keyLength: number): Promise<string> {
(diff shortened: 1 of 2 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -224,6 +224,7 @@ export class MLSService {
     if (subconversation.epoch === 0) {
       await this.registerConversation(subconversation.group_id, []);
     } else {
+      //TODO: check timestamp, if subconv is older than 24h -> delete and re-create subconversation
```
atomrc, line 227: 👍

packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -232,17 +233,22 @@ export class MLSService {
...
 
     return {
       subconversation,
-      secretKey: new TextDecoder().decode(secretKey),
+      secretKey,
       keyLength,
     };
   }
 
+  public async exportSecretKey(groupId: string, keyLength: number): Promise<string> {
```
atomrc, line 246: maybe we could make it `private` first, WDYT?

PatrykBuniX: We will need it on webapp side when updating epoch info when handling `mls-message-add` backend event including `subconv: "conference"`. (Unless we will provide any other method that will have exporting this key included)

atomrc: I'm not yet sure how this update it going to happen, but there might be a async way that will warn the webapp that a new epoch and a new secret key was generated. 

let's keep it private for now and we could change it later on, as we go. 


Rule of thumb: Method should be private by default and it's trivial to make them public later on :)

PatrykBuniX: makes sense 👍

atomrc (approved):

PR #4811, opened 2023-01-24
feat(core, api-client): mls conference improvements

PatrykBuniX:
Improvements in mls conferencing logic.


atomrc (approved):

packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -155,6 +161,9 @@ export class MLSService extends TypedEventEmitter<Events> {
       clientIsExistingGroupUser: (client, otherClients) => {
         const decoder = new TextDecoder();
         const {user} = parseFullQualifiedClientId(decoder.decode(client));
+        //FIXME: handle the case to return true for subconversations if the client is a member of parent conversation
+        // (not possible with corecrypto for now)
+        return true;
```
atomrc, line 166: Let's keep this change local only (for dev). This should never make it to the main branch (as it could easily be let there forever!)

atomrc (approved):

PR #4836, opened 2023-02-01
refactor: simplify remove clients from mls convo api

PatrykBuniX:


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -91,6 +91,8 @@ export class MLSService extends TypedEventEmitter<Events> {
       keyingMaterialUpdateThreshold = defaultConfig.keyingMaterialUpdateThreshold,
       nbKeyPackages = defaultConfig.nbKeyPackages,
     }: Partial<MLSServiceConfig>,
+    private readonly textEncoder = new TextEncoder(),
+    private readonly textDecoder = new TextDecoder(),
```
atomrc, line 95: No need to put them as params of the constructor, you could move them a few lines above

```js
class MLSService {
    logger = logdown(..)
    //...
    private readonly textEncoder = new TextEncoder();
    private readonly textDecoder = new TextDecoder();

   constructor() //..
}
```

packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -371,9 +371,19 @@ export class MLSService extends TypedEventEmitter<Events> {
     return response;
   }
 
-  public removeClientsFromConversation(conversationId: ConversationId, clientIds: Uint8Array[]) {
-    return this.processCommitAction(conversationId, () =>
-      this.coreCryptoClient.removeClientsFromConversation(conversationId, clientIds),
+  /**
+   * Will send a commit with removal proposal for given clients
```
atomrc, line 375: I'm not entirely sure it sends a proposal, I believe it a commit, isn't it?

PatrykBuniX: I'm not sure either 😅 Let's call it removal commit 👍

atomrc (approved):

PR #4842, opened 2023-02-02
refactor: Cleanup session unit tests to prepare for partial message sending

atomrc:
This cleans up a bit the unit tests in preparation of handling partial message sending. 
The idea is that when a federated backend is down and some messages could not go through, we will need to warn the sender that some recipients will not be able to receive those messages. 

When creating a session, we will need to detect sessions that could not be created because the remote backend is down, this is an effort to prepare for this

### DIFF
--- a/packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.mocks.ts
+++ b/packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.mocks.ts
@@ -20,8 +20,8 @@
 import {ClientType} from '@wireapp/api-client/lib/client';
 
 import {APIClient} from '@wireapp/api-client';
-import {CoreCrypto} from '@wireapp/core-crypto';
 
+import {CryptoClient} from './CryptoClient';
 import {CoreCryptoWrapper} from './CryptoClient/CoreCryptoWrapper';
 import {ProteusService} from './ProteusService';
 
@@ -29,7 +29,7 @@ import {getUUID} from '../../../test/PayloadHelper';
 
 export const buildProteusService = async (
   federated = false,
-): Promise<[ProteusService, {apiClient: APIClient; coreCrypto: CoreCrypto}]> => {
+): Promise<[ProteusService, {apiClient: APIClient; cryptoClient: CryptoClient}]> => {
   const apiClient = new APIClient({urls: APIClient.BACKEND.STAGING});
 
   apiClient.context = {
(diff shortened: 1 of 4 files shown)


packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.mocks.ts
```diff
@@ -38,11 +38,11 @@ export const buildProteusService = async (
     clientId: getUUID(),
   };
 
-  const coreCrypto = await CoreCrypto.deferredInit('store-name', 'key');
+  const cryptoClient = new CoreCryptoWrapper({} as any, {} as any, {} as any);
```
atomrc, line 41: We can now use the generic wrapper around CoreCrypto and Cryptobox, no need for a specific implementation here :)

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.test.ts
```diff
@@ -124,35 +123,41 @@ const prepareDataForEncryption = async (useQualifiedIds: boolean = true) => {
...
-      const getPrekeyMock = jest.spyOn(apiClient.api.user, 'getClientPreKey').mockResolvedValue({
-        client: 'client1',
-        prekey: {
-          id: 123,
-          key: 'pQABARhIAqEAWCCaJpFa9c626ORmjj1aV6OnOYgmTjfoiE3ynOfNfGAOmgOhAKEAWCD60VMzRrLfO+1GSjgyhnVp2N7L58DM+eeJhZJi1tBLfQT2',
-        },
-      });
-      jest.spyOn(coreCrypto, 'proteusFingerprintRemote').mockResolvedValue(expectedFingerprint);
-      jest.spyOn(coreCrypto as any, 'proteusSessionExists').mockResolvedValue(false);
-
```
atomrc, line 139: Those scenarios (creating the new session and getting prekeys) are already covered by the SessionHandler tests

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.test.ts
```diff
@@ -225,36 +237,24 @@ describe('ProteusService', () => {
...
       const encryptedPayload = new Map([
         [firstUser.sessions.first, encryptedMessageBuffer],
         [firstUser.sessions.second, encryptedMessageBuffer],
       ]);
 
-      jest
-        .spyOn(services.apiClient.api.user, 'postMultiPreKeyBundles')
-        .mockImplementationOnce(() => Promise.resolve(preKeyBundleMap));
-
```
atomrc, line 250: Doing the request to create the session from prekeys and getting the prekeys from backend are already covered in the SessionHandler tests

packages/core/src/messagingProtocols/proteus/Utility/SessionHandler/SessionHandler.ts
```diff
@@ -61,26 +61,6 @@ const parseSessionId = (sessionId: string): SessionId => {
...
-  apiClient,
-}: CreateSessionParams): Promise<void> => {
-  const {userId, clientId, domain} = parseSessionId(sessionId);
-  const prekey =
-    initialPrekey ?? (await apiClient.api.user.getClientPreKey({id: userId, domain: domain ?? ''}, clientId)).prekey;
-  const prekeyBuffer = Decoder.fromBase64(prekey.key).asBytes;
-  await cryptoClient.sessionFromPrekey(sessionId, prekeyBuffer);
-  await cryptoClient.saveSession(sessionId);
-};
-
```
atomrc, line 83: Instead of having a method that create a single session (and duplicated code) we could instead use the `createSessions` (plural) with an array of a single session. 
This way we don't branch out and have a single piece of logic for creating sessions

thisisamir98 (approved):

PR #4867, opened 2023-02-09
refactor(core): remove unused deprecated method and improve fetchAll api

PatrykBuniX:
Removes the method that was used to fetch all conversation's participants in a hacky way + removes ConversationService dependencies that are not needed anymore.

Improves the api of the method we're using instead.

### DIFF
--- a/packages/core/src/conversation/ConversationService/ConversationService.ts
+++ b/packages/core/src/conversation/ConversationService/ConversationService.ts
@@ -80,53 +80,16 @@ export class ConversationService {
 
   /**
    * Get a fresh list from backend of clients for all the participants of the conversation.
-   * This is a hacky way of getting all the clients for a conversation.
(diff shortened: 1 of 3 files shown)


packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -54,21 +52,17 @@ import {isMLSConversation} from '../../util';
...
-import {MessageService} from '../message/MessageService';
 
 export class ConversationService {
   public readonly messageTimer: MessageTimer;
-  private readonly messageService: MessageService;
   private readonly logger = logdown('@wireapp/core/ConversationService');
 
   constructor(
     private readonly apiClient: APIClient,
-    private readonly config: {useQualifiedIds?: boolean},
```
PatrykBuniX, line 66: These 2 are not being used anymore (were used only by the method we removed)

atomrc (approved):
Nice 😎

PR #4882, opened 2023-02-13
docs(license-collector): Fix broken licenses link

bennycode:
The website https://wire.com/legal/licenses/ shows a "Page not found" error. Possible fixes are to update all links in this monorepo to use https://wire.com/legal/#legal5 or to fix the redirect of the website.


tlebon:
@bennycode does this mean you are ready to rejoin? 😛 
unfortunately, #legal5 doesn't actually navigate via hash because of how the page renders.

bennycode:
I've also noticed the issue with `#legal5`. It might be better to just link to https://wire.com/legal/ instead, as this would be a more effective solution than the current one, which is:

image

PR #4885, opened 2023-02-15
fix: Improve secretKey generator to handle multiple system crypto

atomrc:
We have a full history of different system crypto that we had in the past:

- a system that would `TextEncoder.encode/TextDecoder.decode` (that was inherently broken because of data loss when decoding a buffer)
- a system that used `base64 encode/decode` (which was fine but should not have been opiniated about this)
- the new system that returns a string (instead of a `Uint8Array`) and doesn't manipulate the content at all

This PR will handle all the different systems and wipe the DB in case there is a secret key that is corrupted


packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/secretKeyGenerator.ts
```diff
@@ -0,0 +1,98 @@
...
+          const decrypted = await baseCrypto.decrypt(value);
+          if (isBase64.test(decrypted)) {
+            return Decoder.fromBase64(decrypted).asBytes;
+          }
+          // Between June 2022 and October 2022, the systemCrypto returned a string encoded in UTF-8
+          const encoder = new TextEncoder();
+
+          return encoder.encode(decrypted);
+        },
+      }
```
atomrc, line 69: This is the meat of this PR: handling different type of encryption system (the new one and those from the past).

packages/core/src/Account.ts
```diff
@@ -312,40 +310,17 @@ export class Account<T = any> extends TypedEventEmitter<Events> {
...
-      ? await createCustomEncryptedStore(dbName, systemCrypto)
-      : await createEncryptedStore(dbName);
-
-    let key = await this.secretsDb.getsecretValue(coreCryptoKeyId);
-    if (!key) {
-      key = crypto.getRandomValues(new Uint8Array(16));
-      await this.secretsDb.saveSecretValue(coreCryptoKeyId, key);
-    }
-    await this.secretsDb?.close();
-    return key;
```
atomrc, line 330: This has been moved to a specific handler covered by unit tests

packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/secretKeyGenerator.test.ts
```diff
@@ -0,0 +1,145 @@
...
+} as const;
+
+describe('SecretKeyGenerator', () => {
+  beforeEach(async () => {
+    return new Promise(resolve => {
+      Object.values(systemCryptos).forEach(version => {
+        jest.spyOn(version, 'encrypt');
+        jest.spyOn(version, 'decrypt');
+      });
+      const deleteReq = indexedDB.deleteDatabase('test');
```
PatrykBuniX, line 66: Love the thoroughness of the tests 🚀  One thing: what about moving this `'test'` db name to const variable?

PatrykBuniX (approved):

PR #4889, opened 2023-02-16
feat(core): access last resort prekey and its id from corecrypto [FS-1542]

PatrykBuniX:
https://wearezeta.atlassian.net/browse/FS-1542

Access last resort prekey from `corecrypto` directly - no need to generate it using hardcoded last key id number (`65535`).

Last prekey and its id remained the same after changes:
Screenshot 2023-02-16 at 09 59 43

I've also removed `LAST_PREKEY_ID` from our types as it did already exist in `@wireapp/proteus` package and will be consumed by `cryptobox` implementation only (Did a `console.log` to make sure the value is exactly the same). 
Screenshot 2023-02-16 at 10 25 44

### DIFF
--- a/packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CryptoboxWrapper.ts
+++ b/packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CryptoboxWrapper.ts
@@ -20,9 +20,10 @@
 import {PreKey} from '@wireapp/api-client/lib/auth';
 
 import {Cryptobox} from '@wireapp/cryptobox';
+import {keys as ProteusKeys} from '@wireapp/proteus';
 import {CRUDEngine} from '@wireapp/store-engine';
 
-import {CryptoClient, LAST_PREKEY_ID} from './CryptoClient.types';
+import {CryptoClient} from './CryptoClient.types';
 
 type Config = {
   onNewPrekeys: (prekeys: PreKey[]) => void;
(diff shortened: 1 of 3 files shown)


packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CryptoboxWrapper.ts
```diff
@@ -20,9 +20,10 @@
 import {PreKey} from '@wireapp/api-client/lib/auth';
 
 import {Cryptobox} from '@wireapp/cryptobox';
+import {keys as ProteusKeys} from '@wireapp/proteus';
```
atomrc, line 23: Doesn't coreCrypto expose this value? I think at some point we would like to remove `@wirapp/proteus` at all

atomrc: Oh wait, this is the cryptobox wrapper, all good you can ignore my comment :)

atomrc (approved):

packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CryptoboxWrapper.ts
PatrykBuniX: I don't think we want to use coreCrypto type in cryptobox wrapper, do we? So maybe let me just revert it and keep the value on our side for cryptobox?

atomrc:
solid 🚀

PR #4909, opened 2023-02-21
feat: add generic request for backend calls

aweiss-dev:
A generic way to access the backend and receive data via api-client

### DIFF
--- a/packages/api-client/src/generic/GenericAPI.ts
+++ b/packages/api-client/src/generic/GenericAPI.ts
@@ -17,7 +17,7 @@
  *
  */
 
-import {AxiosRequestConfig} from 'axios';
+import {AxiosRequestConfig, AxiosResponse} from 'axios';
 
 import {HttpClient} from '../http';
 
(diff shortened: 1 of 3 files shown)


packages/api-client/src/generic/GenericAPI.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import {HttpClient} from '../http';
+
+export class GenericAPI {
+  constructor(private readonly client: HttpClient) {}
+
+  /**
+   * Get Data from a given URL.
+   */
+  public async get<T = unknown>(url: string): Promise<T> {
```
atomrc, line 30: Don't we want something even more generic? Something like a `request` method that could do a GET/POST... and take a full `AxiosRequestConfig` object?

packages/api-client/src/generic/GenericAPI.ts
```diff
@@ -0,0 +1,43 @@
...
+      url,
+    };
+
+    try {
+      const response = await this.client.sendJSON<T>(config);
+      return response.data;
+    } catch (error) {
+      throw error;
+    }
+  }
```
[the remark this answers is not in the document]
aweiss-dev: I was thinking about logging the error but removed it in the end. So yes, it is not needed anymore

packages/api-client/src/generic/GenericAPI.ts
aweiss-dev: I would instead add post, put, and delete calls when needed. I was thinking about taking in a config, but I didn't have the use case for it yet.

atomrc: ok, fair enough, let's let it evolve as we go 👌

atomrc (approved):

PR #4912, opened 2023-02-22
feat: leave stale subconversations after crash [FS-1564]

PatrykBuniX:
Enables leaving subconversations after client crash (or reload).


packages/core/src/messagingProtocols/mls/MLSService/stores/subconversationGroupIdStore/subconversationGroupIdStore.ts
```diff
@@ -0,0 +1,78 @@
...
+  currentMap.delete(subconversationId);
+  localStorage.setItem(storageKey, JSON.stringify(Array.from(currentMap.entries())));
+};
+
+function storeGroupId(parentConversation: QualifiedId, subconversation: string, subgroupId: string): void {
+  const subconversationId = generateSubconversationId(parentConversation, subconversation);
+  addItemToMap(subconversationId, subgroupId);
+}
+
+function getGroupId(parentConversation: QualifiedId, subconversation: string): string | undefined {
```
PatrykBuniX, line 56: This is what we were doing before see two deleted files in this PR (they should be renamed not recreated - thanks git 😓). What I've added is storing the map in localStorage + removal option, so we can keep track of "known" subconversations -> we can access list of subconversations after client crash and leave them.

atomrc: maybe we should rename this module adding the idea that those subconv are only temporary. 

I fear if we make it too general purposes it's going to be used in the future for subconv that should stay long lived and then it's going to backfire by deleting the conversation on load. 

Maybe `TemporarySubconversationGroupIdStore` or something like this. WDYT?

PatrykBuniX: Yeah, good idea, even though we're using `localStorage` they're not eternal. Good call 👌

PatrykBuniX: Resolved on 1:1 chat: We've ended up keeping the storage as it is, we changed `getAllGroupIds` to `getAllGroupIdsBySubconversationId` to be sure only subconversations of specified id (eg. `"conference"`) are returned (and cleared on app reload).

With this implementation store has only one responsibility: persistent storage of groupIds. Consumer is able to remove/add entries whenever it want.

atomrc (approved):

packages/core/src/messagingProtocols/mls/EventHandler/events/messageAdd/messageAdd.ts
```diff
@@ -38,10 +38,17 @@ const handleMLSMessageAdd = async (
...
+
+  const groupId = await mlsService.getGroupIdFromConversationId(qualifiedConversationId, event.subconv);
+
+  // We should not receive a message for a group the client is not aware of
+  if (!groupId) {
+    throw new Error(
+      `Could not find a group_id for conversation ${qualifiedConversationId.id}@${qualifiedConversationId.domain}`,
+    );
+  }
+
```
PatrykBuniX, line 51: Since `getGroupIdFromConversationId` does not throw anymore, we have to throw an error here (check out the comment in the code above)

packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -234,19 +234,46 @@ export class MLSService extends TypedEventEmitter<Events> {
...
+   */
   public async leaveConferenceSubconversation(conversationId: QualifiedId): Promise<void> {
     const subconversationGroupId = await this.getGroupIdFromConversationId(
       conversationId,
       SUBCONVERSATION_ID.CONFERENCE,
     );
 
+    if (!subconversationGroupId) {
+      return;
+    }
```
PatrykBuniX, line 250: We don't do anything when client tried to leave a call it's not aware of

atomrc (approved):

PR #4927, opened 2023-02-28
refactor: Sunset non-domain-aware message sending

atomrc:
At wire we plan on sunsetting old backend api version every 6 month.
The api v0 (that doesn't support domain-prefixed endpoints) is more that a year old now so we can remove all the specific code that generates non-domain entities and query non-domain endpoints. 

BREAKING CHANGE: Will not be compatible with a backend that only supports api v0 (backends that are 1 year old). You will now be forced to pass fully qualified ID when calling method that need conversation or user identification


packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -256,51 +243,30 @@ export class ProteusService {
...
-    const missingRecipients: QualifiedUserClients = {};
-
-    for (const [domain, preKeyBundleMap] of Object.entries(preKeyBundles)) {
-      const {unknowns, payloads} = await this.encrypt(plainText, preKeyBundleMap, domain);
-      qualifiedOTRRecipients[domain] = payloads;
-      if (unknowns) {
-        missingRecipients[domain] = unknowns;
-      }
-    }
+    const payloads = await this.cryptoClient.encrypt(sessions, plainText);
```
atomrc, line 265: encryption becomes quite trivial now that we don't need to account for non-qualified session. We can just bulk encrypt all the payloads at once (and avoid a few back and forth between JS and wasm lands)

PatrykBuniX (approved):
Dreams come true 😅 🚀

PR #4932, opened 2023-03-02
fix: Do not consume prekeys when listing members-clients of a conversation

atomrc:
When trying to send a message in a conversation and not giving any users to send to, the core will fetch all the prekeys of the members. 
This is not optimal since it will **consume** those prekeys. 

Instead we list the clients of those members (then no prekey is consumed and we might just download the prekeys that we need for encryption)


packages/core/src/conversation/message/UserClientsUtil.ts
```diff
@@ -19,26 +19,31 @@
...
- * @param userClients The UserClients (qualified or not) to flatten
+ * @param userMap The qualified UserMap to flatten
  * @return An array containing the qualified user Ids and the clients info
  */
-export function flattenQualifiedUserClients<T = unknown>(
-  userClients: QualifiedUserClientsContainer<T>,
-): {data: T; userId: QualifiedId}[] {
-  return Object.entries(userClients).reduce((ids, [domain, userClients]) => {
-    return [...ids, ...flattenUserClients(userClients, domain)];
+export function flattenUserMap<T = unknown>(userMap: QualifiedUserMap<T>): {data: T; userId: QualifiedId}[] {
```
atomrc, line 31: Now that `qualified` is the default, we can simplify some naming here and there :)

V-Gira (approved):

PatrykBuniX (approved):

PR #4945, opened 2023-03-07
refactor(core): subconversation group id store

PatrykBuniX:
Improves naming inside `subconversationGroupId` store module. 
Covers one extra case where subconversation entry can be removed from the store.

### DIFF
--- a/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
+++ b/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
@@ -272,8 +272,8 @@ export class MLSService extends TypedEventEmitter<Events> {
       SUBCONVERSATION_ID.CONFERENCE,
     );
 
-    for (const {parentConversation} of conversationIds) {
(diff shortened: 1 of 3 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -240,7 +240,7 @@ export class MLSService extends TypedEventEmitter<Events> {
    * @param conversationId Id of the parent conversation which subconversation we want to leave
    */
   public async leaveConferenceSubconversation(conversationId: QualifiedId): Promise<void> {
-    const subconversationGroupId = await this.getGroupIdFromConversationId(
+    const subconversationGroupId = subconversationGroupIdStore.getGroupId(
```
PatrykBuniX, line 243: We can access store directly here (this is what `getGroupIdFromConversationId` would do under the hood)

thisisamir98 (approved):

PR #4949, opened 2023-03-10
feat: Forward users for which message could not be encrypted to sender

atomrc:
When a prekey could not be retrieved from backend, we need to warn the consumer that sending the message failed for those users

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -21,6 +21,7 @@
     "@wireapp/store-engine-dexie": "workspace:^",
     "axios": "1.3.4",
     "bazinga64": "6.0.4",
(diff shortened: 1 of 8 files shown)


packages/core/package.json
```diff
@@ -21,6 +21,7 @@
     "@wireapp/store-engine-dexie": "workspace:^",
     "axios": "1.3.4",
     "bazinga64": "6.0.4",
+    "deepmerge-ts": "^4.3.0",
```
PatrykBuniX, line 24: Exact version? 😄

atomrc: good call 👍  Fixed :)

PR #4967, opened 2023-03-13
fix: Improve prekeys tracking between client and backend

atomrc:
Currently the number of prekeys that are available are only computed by the client. 
Which can easily lead to an out-of-sync state between client and backend. 

In order to make sure we are in sync, when we init the device, we query the backend for all the prekey ids we have for this client and store this in a local state. 

This way backend is always the single source of truth

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -259,7 +259,6 @@ export class Account extends TypedEventEmitter<Events> {
     // we reset the services to re-instantiate a new CryptoClient instance
     await this.initServices(this.apiClient.context);
     const initialPreKeys = await this.service.proteus.createClient(entropyData);
-    await this.service.proteus.initClient(this.storeEngine, this.apiClient.context);
 
     const client = await this.service.client.register(loginData, clientInfo, initialPreKeys);
 
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -271,8 +271,7 @@ export class Account extends TypedEventEmitter<Events> {
     await this.service.notification.initializeNotificationStream();
     await this.service.client.synchronizeClients();
 
-    await this.initClient(client);
-    return client;
+    return this.initClient(client);
   }
 
   /**
@@ -280,6 +279,7 @@ export class Account extends TypedEventEmitter<Events> {
    *
    * @returns The local existing client or undefined if the client does not exist or is not valid (non existing on backend)
    */
+  public async initClient(client: RegisteredClient): Promise<RegisteredClient>;
   public async initClient(client?: RegisteredClient): Promise<RegisteredClient | undefined> {
     if (!this.service || !this.apiClient.context || !this.storeEngine) {
       throw new Error('Services are not set.');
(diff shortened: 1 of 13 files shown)


packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
```diff
@@ -154,8 +152,8 @@ export class CoreCryptoWrapper implements CryptoClient {
     return this.prekeyTracker.consumePrekey();
   }
 
-  async newPrekey(id: number) {
-    const key = await this.coreCrypto.proteusNewPrekey(id);
+  async newPrekey() {
+    const [id, key] = await this.coreCrypto.proteusNewPrekeyAuto();
```
atomrc, line 156: ⚠️  This will need a coreCrypto update https://github.com/wireapp/core-crypto/blob/develop/CHANGELOG.md#070---2023-03-

packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/PrekeysTracker/PrekeysTracker.store.ts
```diff
@@ -17,49 +17,32 @@
...
   nbPrekeys: number;
-  highestId: number;
 };
-export class PrekeysTrackerStore {
-  constructor(private readonly db: CoreDatabase) {}
 
-  private async getState(): Promise<PrekeysState> {
-    return (await this.db.get('prekeys', STATE_PRIMARY_KEY)) ?? {nbPrekeys: 0, highestId: 0};
-  }
+let state: PrekeysState = {nbPrekeys: 0};
```
atomrc, line 24: Now the state is an in memory state only. No need for database access anymore (since it will be synced at load time with backend)

packages/core/src/Account.ts
```diff
@@ -259,7 +259,6 @@ export class Account extends TypedEventEmitter<Events> {
     // we reset the services to re-instantiate a new CryptoClient instance
     await this.initServices(this.apiClient.context);
     const initialPreKeys = await this.service.proteus.createClient(entropyData);
-    await this.service.proteus.initClient(this.storeEngine, this.apiClient.context);
```
atomrc, line 262: This was unecessary as it's done in the `this.initClient` at the end (so we were doing it twice)

PatrykBuniX (approved):
Seems solid 🚀

packages/core/src/Account.ts
```diff
@@ -272,15 +271,15 @@ export class Account extends TypedEventEmitter<Events> {
...
+    return this.initClient(client);
   }
 
   /**
    * Will initiate all the cryptographic material of the device and setup all the background tasks.
    *
    * @returns The local existing client or undefined if the client does not exist or is not valid (non existing on backend)
    */
+  public async initClient(client: RegisteredClient): Promise<RegisteredClient>;
   public async initClient(client?: RegisteredClient): Promise<RegisteredClient | undefined> {
```
atomrc, line 283: This improves the return type of the function (if a client is given, then the return type is not going to be undefined)

PR #4968, opened 2023-03-14
fix: Do not try to fetch clients when no users are given

atomrc:


PatrykBuniX (approved):

packages/core/src/messagingProtocols/proteus/Utility/Recipients.ts
```diff
@@ -49,7 +49,10 @@ const getRecipientsForConversation = async ({
     return userIds;
   }
 
-  const recipientIds = userIds || (await getConversationQualifiedMembers({apiClient: apiClient, conversationId}));
+  const recipientIds =
+    userIds && Object.keys(userIds).length > 0
```
arjita-mitra, line 53: adding userIds && Object.keys(userIds).length > 0 in a const will make the code more redable

PR #4972, opened 2023-03-17
fix(core): wait until corecrypto is ready to call .wipe method [FS-1562]

PatrykBuniX:
## Cause
User logged in on removed device ended up crushing on conversation view and not getting logged out properly (device was not completely deleted, storage was not cleared). For more details see [ticket](https://wearezeta.atlassian.net/browse/FS-1562).

## Problem
Due to rust lang specificity, only 1 "strong reference" is being allowed to be alive at a time - reference gets created every time we call any `corecrypto` method (and gets cleared after the method is done executing). It was possible that in case of deleting other self device, it was not logged out properly because client tried to call `.wipe()` method while it was decrypting a message (`.decryptMessage()` was still executing).

## Solution
CoreCrypto implemented new api - `.strongRefCount()` method, allowing us to check whether corecrypto is ready to call new method.

See https://github.com/wireapp/core-crypto/pull/272/files#diff-d259ed7cc527064fee3325e07d38c1a4921ebc42ec92fddfc36e947e4c898c08R679 for more details.

**NOTE:**
CoreCrypto team will release better api in the future that hides implementation details (consumer won't need to be aware of strong refs related logic)

### DIFF
--- a/packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
+++ b/packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
@@ -177,31 +177,35 @@ export class CoreCryptoWrapper implements CryptoClient {
     return this.coreCrypto.proteusCryptoboxMigrate(dbName);
   }
 
-  async handleSafeWipe() {
-    //See https://github.com/wireapp/wire-web-packages/pull/4972 for more details
-    const wipeWhenReady = async (onReady: () => void) => {
+  /**
+   * Will call the callback once corecrypto is ready.
+   * @param callback - Function to be called once corecrypto is ready.
+   * @see https://github.com/wireapp/wire-web-packages/pull/4972
+   */
+  private async onReady(callback: () => Promise<void>) {
+    const isCoreCryptoReady = async () => {
       const strongRefCount = this.coreCrypto.strongRefCount();
       const isCoreCryptoReady = strongRefCount <= 1;
-      if (isCoreCryptoReady) {
-        await this.coreCrypto.wipe();
-        return onReady();
-      }
+      return isCoreCryptoReady;
     };
 
-    return new Promise<void>(async resolve => {
-      await wipeWhenReady(resolve);
+    if (await isCoreCryptoReady()) {
+      return callback();
+    }
 
+    return new Promise<void>(async resolve => {
       const intervalId = setInterval(async () => {
-        await wipeWhenReady(() => {
+        if (await isCoreCryptoReady()) {
           clearInterval(intervalId);
+          await callback();
           return resolve();
-        });
-      }, 500);
+        }
+      }, 100);
     });
   }
 
   async wipe() {
     await this.config.onWipe();
-    return this.handleSafeWipe();
+    await this.onReady(() => this.coreCrypto.wipe());
   }
 }
--- a/packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
+++ b/packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
@@ -177,8 +177,29 @@ export class CoreCryptoWrapper implements CryptoClient {
     return this.coreCrypto.proteusCryptoboxMigrate(dbName);
   }
 
+  async handleSafeWipe() {
+    const wipeWhenReady = async (onReady: () => void) => {
+      const strongRefCount = this.coreCrypto.strongRefCount();
+      if (strongRefCount <= 1) {
+        await this.coreCrypto.wipe();
+        return onReady();
+      }
+    };
+
+    return new Promise<void>(async resolve => {
+      await wipeWhenReady(resolve);
+
+      const intervalId = setInterval(async () => {
+        await wipeWhenReady(() => {
+          clearInterval(intervalId);
+          return resolve();
+        });
+      }, 500);
+    });
+  }
+
   async wipe() {
     await this.config.onWipe();
-    return this.coreCrypto.wipe();
+    return this.handleSafeWipe();
   }
 }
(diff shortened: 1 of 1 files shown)


packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
```diff
@@ -177,8 +177,31 @@ export class CoreCryptoWrapper implements CryptoClient {
     return this.coreCrypto.proteusCryptoboxMigrate(dbName);
   }
 
+  async handleSafeWipe() {
+    //See https://github.com/wireapp/wire-web-packages/pull/4972 for more details
+    const wipeWhenReady = async (onReady: () => void) => {
+      const strongRefCount = this.coreCrypto.strongRefCount();
+      const isCoreCryptoReady = strongRefCount <= 1;
+      if (isCoreCryptoReady) {
```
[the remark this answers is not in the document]
PatrykBuniX: For me `strongRefCount <= CONST_VALUE` has no meaning if you're not in the context, with `isCoreCryptoReady` you know what is being checked here 🤔 (actually the best solution would be corecrypto to do the check internally and provide sth like `this.coreCrypto.isStrongRefLocked()` or sth like this)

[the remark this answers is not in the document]
PatrykBuniX: Aligned with cc team already, they will hide strong ref logic and expose nicer api 😄

packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
```diff
@@ -177,8 +177,31 @@ export class CoreCryptoWrapper implements CryptoClient {
     return this.coreCrypto.proteusCryptoboxMigrate(dbName);
   }
 
+  async handleSafeWipe() {
```
atomrc, line 180:
```suggestion
  private async handleSafeWipe() {
```

packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
```diff
@@ -177,8 +177,31 @@ export class CoreCryptoWrapper implements CryptoClient {
...
+          return resolve();
+        });
+      }, 500);
+    });
+  }
+
   async wipe() {
     await this.config.onWipe();
-    return this.coreCrypto.wipe();
+    return this.handleSafeWipe();
```
atomrc, line 205: How about an API like `this.onReady(() => this.coreCrypto.wipe())`

And implementing this generic `onReady` and avoid mixing wiping concerns with readiness concerns. 
WDYT?

packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
```diff
@@ -177,8 +177,35 @@ export class CoreCryptoWrapper implements CryptoClient {
...
+   * Will call the callback once corecrypto is ready.
+   * @param callback - Function to be called once corecrypto is ready.
+   * @see https://github.com/wireapp/wire-web-packages/pull/4972
+   */
+  private async onReady(callback: () => Promise<void>) {
+    const isCoreCryptoReady = async () => {
+      const strongRefCount = this.coreCrypto.strongRefCount();
+      const isCoreCryptoReady = strongRefCount <= 1;
+      return isCoreCryptoReady;
+    };
```
atomrc, line 190:
```suggestion
    const isCoreCryptoReady = () => {
      const strongRefCount = this.coreCrypto.strongRefCount();
      const isCoreCryptoReady = strongRefCount <= 1;
      return isCoreCryptoReady;
    };
```
There is no await in there it seems

packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
```diff
@@ -177,8 +177,35 @@ export class CoreCryptoWrapper implements CryptoClient {
...
+
+    if (await isCoreCryptoReady()) {
```
atomrc, line 192:
```suggestion
    if (isCoreCryptoReady()) {
```

packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
```diff
@@ -177,8 +177,35 @@ export class CoreCryptoWrapper implements CryptoClient {
     return this.coreCrypto.proteusCryptoboxMigrate(dbName);
   }
 
+  /**
+   * Will call the callback once corecrypto is ready.
+   * @param callback - Function to be called once corecrypto is ready.
+   * @see https://github.com/wireapp/wire-web-packages/pull/4972
+   */
+  private async onReady(callback: () => Promise<void>) {
```
atomrc, line 185:
```suggestion
  private onReady(callback: () => Promise<void>) {
```

packages/core/src/messagingProtocols/proteus/ProteusService/CryptoClient/CoreCryptoWrapper/CoreCryptoWrapper.ts
```diff
@@ -177,8 +177,35 @@ export class CoreCryptoWrapper implements CryptoClient {
...
+      return isCoreCryptoReady;
+    };
+
+    if (await isCoreCryptoReady()) {
+      return callback();
+    }
+
+    return new Promise<void>(async resolve => {
+      const intervalId = setInterval(async () => {
+        if (await isCoreCryptoReady()) {
```
atomrc, line 198:
```suggestion
        if (isCoreCryptoReady()) {
```

atomrc (approved):
I think there are some extra unneeded async/await but looks good man 😎

atomrc (approved):

PR #4973, opened 2023-03-19
feat: update reaction type

arjita-mitra:
## Pull Request Checklist

- update reaction type to accept string


V-Gira (approved):

atomrc (approved):
This can also be removed from the core if you feel it makes sense. 
I don't think having an alias to `string` is really worth it here 🤷

PR #4991, opened 2023-03-20
feat(api-client): Truncate domain name

przemvs:
## Pull Request Checklist


packages/api-client/src/APIClient.ts
```diff
@@ -356,7 +356,8 @@ export class APIClient extends EventEmitter {
     try {
       const self = await this.api.self.getSelf();
       selfDomain = self.qualified_id?.domain;
-      this.logger.info(`Got self domain "${selfDomain}"`);
+
+      this.logger.info(`Got self domain "${selfDomain?.substring(0, 3) || ''}***}"`);
```
atomrc, line 360: Let's remove that log line at all, it doesn't add any value
```suggestion
```

PR #4992, opened 2023-03-21
fix: Make sure current client is not duplicated in database

atomrc:
This will ensure that we don't try to send messages for a device we will never have a session with (namely the current device)

### DIFF
--- a/packages/core/src/client/ClientService.ts
+++ b/packages/core/src/client/ClientService.ts
@@ -133,9 +133,13 @@ export class ClientService {
     return this.database.createLocalClient(client, domain);
   }
 
-  public async synchronizeClients(): Promise<MetaClient[]> {
+  /**
+   * Will download all the clients of the self user (excluding the current client) and will store them in the database
+   * @param currentClient - the id of the current client (to be excluded from the list)
+   */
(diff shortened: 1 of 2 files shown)


packages/core/src/client/ClientService.ts
```diff
@@ -133,9 +133,13 @@ export class ClientService {
...
 
-  public async synchronizeClients(): Promise<MetaClient[]> {
+  /**
+   * Will download all the clients of the self user (excluding the current client) and will store them in the database
+   * @param currentClient - the id of the current client (to be excluded from the list)
+   */
+  public async synchronizeClients(currentClient: string): Promise<MetaClient[]> {
     const registeredClients = await this.backend.getClients();
-    const filteredClients = registeredClients.filter(client => client.id !== this.apiClient.context!.clientId);
+    const filteredClients = registeredClients.filter(client => client.id !== currentClient);
```
atomrc, line 142: This was a pure temporal coupling (depending on when the method is called, the outcome would be different). 
In this case we force a client Id to be given explicitely removing any kind of temporal coupling

PatrykBuniX (approved):

PR #4994, opened 2023-03-22
feat: allow partial user listing for federated backends

atomrc:
This will adapt the API client to receive partial list of users (for when one federated backend is down). 

BREAKING CHANGE: The return type of `postListUser` will now always wrap the result in a full `UsersResponse`

### DIFF
--- a/packages/api-client/src/user/UserAPI.ts
+++ b/packages/api-client/src/user/UserAPI.ts
@@ -51,9 +51,17 @@ type PrekeysResponse = {
   qualified_user_client_prekeys: QualifiedUserPreKeyBundleMap;
   failed_to_list?: QualifiedId[];
 };
-
 function isPrekeysResponse(object: any): object is PrekeysResponse {
   return object.qualified_user_client_prekeys && object.failed_to_list;
+}
+
+type UsersReponse = {
+  found?: User[];
+  failed?: QualifiedId[];
+  not_found?: QualifiedId[];
+};
+function isUsersResponse(object: any): object is UsersReponse {
+  return object.found || object.failed || object.not_found;
 }
 
 export class UserAPI {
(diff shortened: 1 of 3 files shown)


packages/api-client/src/user/UserAPI.ts
```diff
@@ -51,11 +51,19 @@ type PrekeysResponse = {
...
   return object.qualified_user_client_prekeys && object.failed_to_list;
 }
 
+type UsersReponse = {
+  found?: User[];
+  failed?: QualifiedId[];
+  not_found?: QualifiedId[];
+};
+function isUsersResponse(object: any): object is UsersReponse {
+  return object.found || object.failed || object.not_found;
```
arjita-mitra, line 64: why the param named object? can be give it a meaningful name like?

packages/api-client/src/user/UserAPI.ts
```diff
@@ -519,15 +527,18 @@ export class UserAPI {
...
+  ): Promise<UsersReponse> {
     const config: AxiosRequestConfig = {
       data: users,
       method: 'post',
       url: UserAPI.URL.LIST_USERS,
     };
 
-    const response = await this.client.sendJSON<User[]>(config);
-    return response.data;
+    const {data} = await this.client.sendJSON<User[] | UsersReponse>(config);
```
arjita-mitra, line 537: can we rename data too like userData?

packages/api-client/src/user/UserAPI.ts
atomrc: That's because it's a typeguard predicate. So we basically give it something, and we don't know what that thing is, and in return it tells us if that thing is a `UserResponse`

packages/api-client/src/user/UserAPI.ts
atomrc: This is the basic axios response, the `data` represents the raw payload. I can sure rename it if you feel it's going to be clearer

packages/api-client/src/user/UserAPI.ts
```diff
@@ -51,11 +51,19 @@ type PrekeysResponse = {
...
 function isPrekeysResponse(object: any): object is PrekeysResponse {
   return object.qualified_user_client_prekeys && object.failed_to_list;
 }
 
+type UsersReponse = {
+  found?: User[];
+  failed?: QualifiedId[];
+  not_found?: QualifiedId[];
+};
+function isUsersResponse(object: any): object is UsersReponse {
```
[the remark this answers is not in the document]
atomrc: Since it's very specific to the UserAPI, I'd rather keep it private for now and not publish it to the higher scope

PR #5035, opened 2023-04-06
fix: make cli-client compatible with latest core

atomrc:
## Pull Request Checklist

### DIFF
--- a/packages/cli-client/overrides/core-crypto/index.js
+++ b/packages/cli-client/overrides/core-crypto/index.js
@@ -1,0 +1,25 @@
+/*
+ * Wire
+ * Copyright (C) 2023 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+module.exports = {
+  RatchetTreeType: {
+    FULL: 0,
+  },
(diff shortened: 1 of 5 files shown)


packages/cli-client/overrides/core-crypto/index.js
```diff
@@ -0,0 +1,25 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+module.exports = {
```
atomrc, line 20: Since `@wireapp/core-crypto` is not compatible with nodejs, we just mock it (it's not really needed, we will need to find a solution when we want to handle MLS in the future)

packages/cli-client/package.json
```diff
@@ -8,8 +8,9 @@
   ],
   "main": "lib/index",
   "dependencies": {
-    "@wireapp/api-client": "workspace:^",
-    "@wireapp/core": "workspace:^",
+    "@wireapp/api-client": "24.3.0",
+    "@wireapp/core": "40.1.6",
+    "@wireapp/core-crypto": "file:./overrides/core-crypto",
```
atomrc, line 13: This is overriding corecrypto (and let the node js runtime run the actual cli)

packages/cli-client/overrides/core-crypto/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@wireapp/core-crypto",
+  "version": "1.0.0",
+  "description": "",
+  "main": "./index.js",
+  "author": "",
+  "license": "ISC"
```
thisisamir98, line 7: Shouldn’t this be gpl3?

atomrc: This is not published anywhere anyway. This is a totally fake package

atomrc: fixed with https://github.com/wireapp/wire-web-packages/pull/5035/commits/78f6c46333d9d01d9db7c161169f1ce35f9b196c

thisisamir98 (approved):

PatrykBuniX (approved):

PR #5048, opened 2023-04-12
fix: Improve detection of valid prekey responses

atomrc:


packages/api-client/src/user/UserAPI.ts
```diff
@@ -52,7 +52,7 @@ type PrekeysResponse = {
   failed_to_list?: QualifiedId[];
 };
 function isPrekeysResponse(object: any): object is PrekeysResponse {
-  return object.qualified_user_client_prekeys && object.failed_to_list;
```
atomrc, line 55: Now that `failed_to_list` is optional on backend side, we need to also consider a response without `failed_to_list` as a valid response

PatrykBuniX (approved):

V-Gira (approved):

PR #5051, opened 2023-04-13
fix: ensure fetching users is backward compatible [FS-1784]

V-Gira:
### Issue
this change https://github.com/wireapp/wire-webapp/pull/14923 renders the webapp incompatible wth older versions of the api

### Solution
Reintrodruce the previous webapp logic in `api-client`: if the response correspond to the old api, map users from the same backend in `found` and all federated users in `failed`

### DIFF
--- a/packages/api-client/src/user/UserAPI.ts
+++ b/packages/api-client/src/user/UserAPI.ts
@@ -534,12 +534,20 @@ export class UserAPI {
       url: UserAPI.URL.LIST_USERS,
     };
     try {
+      /**
+       * We expect two differents responses depending on which version of the API the back-end is running
+       * of type UserResponse in the case of a newer back-end
+       * of type User[] in the case of an older back-end
+       */
       const {data: userData} = await this.client.sendJSON<User[] | UsersReponse>(config);
+      /* If the response is of type UserResponse, the webapp consumes it as is */
       if (isUsersResponse(userData)) {
         return userData;
       }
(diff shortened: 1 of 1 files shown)


packages/api-client/src/user/UserAPI.ts
```diff
@@ -533,12 +533,40 @@ export class UserAPI {
...
+      }
+      return {found: userData};
+    } catch (error: any) {
+      if (
+        error.label === BackendErrorLabel.FEDERATION_NOT_AVAILABLE ||
+        error.label === BackendErrorLabel.FEDERATION_BACKEND_NOT_FOUND ||
+        error.label === BackendErrorLabel.FEDERATION_REMOTE_ERROR ||
+        error.label === BackendErrorLabel.FEDERATION_TLS_ERROR
+      ) {
+        const selfDomain = this.backendFeatures.domain;
```
atomrc, line 549: How about a comment that explains what happens there ;) ?

V-Gira:
```suggestion
      ) {
        /* UserAPI,backendFeatures.domain returns the domain of the self user*/
        const selfDomain = this.backendFeatures.domain;
```
something like this or more detail on what `UserAPI.backendFeature` is?

atomrc: No I mean a general comments that explain how we handle federation errors (sorry I didn't mean to comment on this line in particular 😅 )

PR #5095, opened 2023-04-26
fix: Add missing lerna packages

atomrc:
This should fix the publish job (broken since https://github.com/wireapp/wire-web-packages/pull/5072)


package.json
```diff
@@ -6,6 +6,8 @@
     "@lerna-lite/changed": "2.1.0",
     "@lerna-lite/cli": "2.1.0",
     "@lerna-lite/run": "2.1.0",
+    "@lerna-lite/version": "^2.1.0",
+    "@lerna/publish": "^6.6.1",
```
PatrykBuniX, line 10: Are you sure about non-exact versions (`^`)?

atomrc: good call 👌

PatrykBuniX (approved):

V-Gira (approved):

thisisamir98 (approved):

PR #5185, opened 2023-05-29
runfix(core): emit new epoch event only if the group still exists

PatrykBuniX:
- fixes types for onEpochChanged callback to async
- does check if group does exist, before checking epoch number and emitting an event (when `corecrypto.getEpoch` is called with `groupId` that does no longer exists, an error is thrown)

### DIFF
--- a/packages/core/src/conversation/ConversationService/ConversationService.ts
+++ b/packages/core/src/conversation/ConversationService/ConversationService.ts
@@ -367,6 +367,10 @@ export class ConversationService {
   }
 
   public async wipeMLSConversation(groupId: string): Promise<void> {
+    const isMLSConversationEstablished = await this.isMLSConversationEstablished(groupId);
+    if (!isMLSConversationEstablished) {
+      return;
+    }
     return this.mlsService.wipeConversation(groupId);
   }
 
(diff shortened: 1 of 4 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -558,6 +558,10 @@ export class MLSService extends TypedEventEmitter<Events> {
 
   public async handleEvent(params: Omit<EventHandlerParams, 'mlsService'>): EventHandlerResult {
     return handleBackendEvent({...params, mlsService: this}, async groupId => {
+      const conversationExists = await this.conversationExists(groupId);
+      if (!conversationExists) {
+        return;
+      }
```
PatrykBuniX, line 564: This is the main change, we make sure conversation still exist before checking epoch number and emitting an event. It's possible that upon decrypting an mls message (inside `handleBackendEvent`, see https://github.com/wireapp/wire-web-packages/blob/a68052626eea553fcd0b5142ac15b6ae92da5aff/packages/core/src/messagingProtocols/mls/EventHandler/events/messageAdd/messageAdd.ts#L60) we'll be removed from a group and it will be wiped.

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -367,6 +367,10 @@ export class ConversationService {
   }
 
   public async wipeMLSConversation(groupId: string): Promise<void> {
+    const isMLSConversationEstablished = await this.isMLSConversationEstablished(groupId);
+    if (!isMLSConversationEstablished) {
+      return;
+    }
```
PatrykBuniX, line 373: We want to make sure the group exists before wiping it (otherwise an error is thrown)

thisisamir98 (approved):

PR #5186, opened 2023-05-29
feat: Return failed users when sending mls message (FS-2030)

thisisamir98:
Needed to do the UI in webapp


PatrykBuniX (approved):
Looks fine 👌

PR #5187, opened 2023-05-31
fix: Correct type for mls failed_to_send (FS-2030)

thisisamir98:
backend response type for failed_to_send on mls is actually QualifiedId[] not QualifiedUserClients

### DIFF
--- a/packages/core/src/conversation/ConversationService/ConversationService.ts
+++ b/packages/core/src/conversation/ConversationService/ConversationService.ts
@@ -274,24 +274,19 @@ export class ConversationService {
 
     const encrypted = await this.mlsService.encryptMessage(groupIdBytes, GenericMessage.encode(payload).finish());
 
-    let response: PostMlsMessageResponse = {
-      time: new Date().toISOString(),
-      events: [],
-      failed_to_send: [],
-      failed: undefined,
-    };
+    let response: PostMlsMessageResponse | null = null;
 
     try {
       response = await this.apiClient.api.conversation.postMlsMessage(encrypted);
     } catch {}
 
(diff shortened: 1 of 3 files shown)


PatrykBuniX (approved):

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -288,7 +288,7 @@ export class ConversationService {
     const sentAt = response.time;
 
     const failedToSend =
-      response.failed || Object.keys(response.failed_to_send ?? {}).length > 0
```
tlebon, line 291: is response.failed an array or a boolean?

thisisamir98: array but here we only care if it is there or not, backend does not send it when no failure is there.

tlebon: bit confusing to me when we include it on 281 before overwriting it from the BE

thisisamir98: right, changed it to undefined
https://github.com/wireapp/wire-web-packages/pull/5187/commits/4c783e3b00096d86e9a6a262a6abed9c484362bb

thisisamir98: https://github.com/wireapp/wire-web-packages/pull/5187/files#diff-4c1562146ae07c9ddbe8479827807af4a87be8cc13ae150557ad72cf6065f254

tlebon (approved):

PR #5189, opened 2023-06-02
runfix: improve periodic keymaterial update logic

PatrykBuniX:
Improves periodic key material update:
- even though `recurringTaskScheduler` was storing last fire date in `localStorage`, the timers were never restarted after reloading the app
- `keyMaterialUpdatesStore` is not needed anymore - it's `.store` method was never called after we've migrated to `recurringTaskScheduler`
- Made `schedulePeriodicKeyMaterialRenewals` method to be called by consumer - when app is reloaded it will provide the list of groupIds of MLS-capable conversations we should resume the timers for
- periodic key update was never scheduled when user has joined the group itself via external commit, in case it was added via welcome message. This PR covers these two cases.
- from now we cancel the renewal task prematurely when wiping conversation (e.g after we've left, were removed or deleted a conversation).


packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -384,11 +384,6 @@ export class ConversationService {
   }
 
   public async wipeMLSConversation(groupId: string): Promise<void> {
-    const isMLSConversationEstablished = await this.isMLSConversationEstablished(groupId);
-    if (!isMLSConversationEstablished) {
-      //if the mls group does not exist, we don't need to wipe it
-      return;
-    }
```
PatrykBuniX, line 391: This part was moved to `MLSService.wipeConversation` to make sure we always check if conversation exists before wiping it.

packages/core/src/Account.ts
```diff
@@ -304,9 +304,6 @@ export class Account extends TypedEventEmitter<Events> {
       // initialize schedulers for pending mls proposals once client is initialized
       await this.service.mls.checkExistingPendingProposals();
 
-      // initialize schedulers for renewing key materials
-      this.service.mls.checkForKeyMaterialsUpdate();
```
PatrykBuniX, line 308: Now consumer will be the one that initiates intervals and provides the list of `groupIds`

packages/core/src/messagingProtocols/mls/EventHandler/events/welcomeMessage/welcomeMessage.ts
```diff
@@ -35,6 +35,9 @@ const handleWelcomeMessage = async ({mlsService, event}: HandleWelcomeMessagePar
   const newGroupId = await mlsService.processWelcomeMessage(data);
   const groupIdStr = Encoder.toBase64(newGroupId).asString;
   // The groupId can then be sent back to the consumer
+
+  // After we were added to the group we need to schedule a periodic key material renewal
+  mlsService.scheduleKeyMaterialRenewal(groupIdStr);
```
PatrykBuniX, line 40: We were not handling this case previously, we should also maintain the timer when being added to a group (not only when creating one) - basically we should maintain the timer for all the groups we're currently in (the way we've joined it does not matter)

PR #5237, opened 2023-06-15
feat: add workflow for manual release of package beta version

PatrykBuniX:
Should add ability to manually publish beta version of packages.


lerna.json
```diff
@@ -1,7 +1,6 @@
 {
   "command": {
     "publish": {
-      "allowBranch": "main",
```
PatrykBuniX, line 4: This would not allow us to publish a package from a branch other than `main` We already do the check in our pipeline workflow, see https://github.com/wireapp/wire-web-packages/blob/main/.github/workflows/test_build_release.yml#L76, so only packages merged to main will get deployed automatically

tlebon (approved):
gogogogo

PR #5238, opened 2023-06-15
runfix: constant version tag, configurable preid

PatrykBuniX:
Changes how version are tagged when releasing beta version of package:
- version tag will always equal to `@beta`
- prerelease identifier (preid) will be configurable via workflow config

So instead of:
Screenshot 2023-06-15 at 12 37 59

It will be e.g. 
`40.5.3-draft-20-cc.0` with tag `@beta`


tlebon (approved):
good changes

PR #5239, opened 2023-06-15
runfix: remove canceling tasks

PatrykBuniX:
I feel that this might fix the issue with task being auto-cancelled, maybe this cancel-action does not play well with `workflow_dispatch` trigger.


tlebon (approved):
Will this deploy correctly if it's a run fix?

PatrykBuniX:
> Will this deploy correctly if it's a run fix?

it doesn't matter, we just need it to be on main

PR #5255, opened 2023-06-21
runfix: skip changelog update when releasing beta version of package

PatrykBuniX:
Don't update changelog files of packages when beta version of package is released.


V-Gira (approved):
makes sense to me

PR #5338, opened 2023-07-26
refactor: method to register empty mls conversation

PatrykBuniX:
Some refactoring in mls service:
- created a separate method to establish empty mls conversation (just register it with epoch 0 without adding other users or sending empty commit)
- refactored some methods to expect string group ids instead of uint8array

### DIFF
--- a/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
+++ b/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
@@ -400,7 +400,7 @@ export class MLSService extends TypedEventEmitter<Events> {
     });
   }
 
-  public updateKeyingMaterial(groupId: string) {
+  private updateKeyingMaterial(groupId: string) {
     const groupIdBytes = Decoder.fromBase64(groupId).asBytes;
     return this.processCommitAction(groupIdBytes, () => this.coreCryptoClient.updateKeyingMaterial(groupIdBytes));
   }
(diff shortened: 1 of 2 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -396,21 +400,16 @@ export class MLSService extends TypedEventEmitter<Events> {
     });
   }
 
-  public updateKeyingMaterial(conversationId: ConversationId) {
-    return this.processCommitAction(conversationId, () => this.coreCryptoClient.updateKeyingMaterial(conversationId));
+  private updateKeyingMaterial(groupId: string) {
+    const groupIdBytes = Decoder.fromBase64(groupId).asBytes;
```
arjita-mitra, line 404: no error handling here if decoding fails? Can we add a try/catch block here?

arjita-mitra (changes requested):

packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
PatrykBuniX: Why would it fail? It would mean that `groupId` was somehow malformed what would be programming error.

arjita-mitra: if groupId is malformed then don't we want to throw some exception?

PatrykBuniX: It will be thrown by decrypt method itself, this is not something we should handle

PR #5393, opened 2023-08-16
feat(core): establish mls 1to1

PatrykBuniX:
Adds a method to establish (register) MLS conversation, when registering a group fails, check the epoch number of remote conversation and:

- if it's `0` it means the group was not established yet, wipe and retry
- if it's more than `0`, it means that the group was already established by you / the other user, we should join with external commit.

### DIFF
--- a/packages/api-client/src/conversation/Conversation.ts
+++ b/packages/api-client/src/conversation/Conversation.ts
@@ -109,3 +109,10 @@ export interface Conversation {
 
   protocol: ConversationProtocol;
 }
+
+export interface MLSConversation extends Conversation {
+  group_id: string;
+  epoch: number;
+  cipher_suite: number;
+  protocol: ConversationProtocol.MLS;
+}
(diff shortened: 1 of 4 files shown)


packages/api-client/src/conversation/Conversation.ts
```diff
@@ -109,3 +109,10 @@ export interface Conversation {
 
   protocol: ConversationProtocol;
 }
+
+export interface MLSConversation extends Conversation {
+  group_id: string;
+  epoch: number;
+  cipher_suite: number;
+  protocol: ConversationProtocol.MLS;
+}
```
PatrykBuniX, line 118: We are sure that methods like `getMLS1to1Conversations` will return an mls conversation, we can extend the type to tell typescript that groupid and epoch are defined in this case.

packages/core/src/conversation/ConversationService/ConversationService.test.ts
```diff
@@ -214,6 +216,72 @@ describe('ConversationService', () => {
     });
   });
 
+  describe('establishMLS1to1Conversation', () => {
```
atomrc, line 219: solid test cases 🙏

atomrc (approved):

PR #5396, opened 2023-08-16
feat: Handle adding users that are on an unreachable backend

atomrc:
This will add the ability to handle backend errors when adding users to a conversation. 

If users from an unreachable backend are added, then those users are filtered out and marked as `failedToAdd` in the response.

### DIFF
--- a/packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.test.ts
+++ b/packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.test.ts
@@ -38,6 +38,7 @@ import {constructSessionId} from '../Utility/SessionHandler';
 import {NotificationSource} from '../../../notification';
 import {CONVERSATION_EVENT} from '@wireapp/api-client/lib/event';
 import {GenericMessage} from '@wireapp/protocol-messaging';
+import {ProteusService} from './ProteusService';
 
 jest.mock('./CryptoClient/CoreCryptoWrapper/PrekeysTracker', () => {
   return {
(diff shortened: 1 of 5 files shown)


packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.ts
```diff
@@ -172,15 +174,15 @@ export class ProteusService {
...
-          const {backends} = backendError;
-          const users = payload.qualified_users ?? [];
-          const availableUsers: QualifiedId[] = [];
-          const unreachableUsers: QualifiedId[] = [];
-          users.forEach(user =>
-            backends.includes(user.domain) ? unreachableUsers.push(user) : availableUsers.push(user),
+          const {backends} = error;
+          const {excludedUsers: unreachableUsers, includedUsers: availableUsers} = filterUsersFromDomains(
+            payload.qualified_users ?? [],
+            backends,
```
atomrc, line 185: I extracted this logic to an external handler and used it in the àddUsersToConversation` method as well :)

packages/core/src/messagingProtocols/proteus/ProteusService/ProteusService.test.ts
```diff
@@ -572,6 +573,61 @@ describe('ProteusService', () => {
...
+  describe('addUsersToConversation', () => {
+    const baseResponse = {
+      event: {
+        conversation: '',
+        from: '',
+        time: Date.now().toString(),
+        data: {users: [], user_ids: []},
+        type: CONVERSATION_EVENT.MEMBER_JOIN,
+      },
+    } satisfies Awaited<ReturnType<typeof ProteusService.prototype.addUsersToConversation>>;
```
atomrc, line 585: pretty proud of my first usage of `satisfies` \o/

V-Gira (approved):
Looking good, only non-federating users left and we're done :muscle:

PR #5398, opened 2023-08-17
refactor: harmonize failedToAdd response in conversation creation and user addition

atomrc:
This will make sure all the methods that add users to a conversation return the same format.

```js
{
  failedToAdd: {reason: ..., users: ...}
  // method specific data
}
```

### DIFF
--- a/packages/api-client/src/conversation/Conversation.ts
+++ b/packages/api-client/src/conversation/Conversation.ts
@@ -98,15 +98,6 @@ export interface Conversation {
   epoch?: number;
   cipher_suite?: number;
 
-  /**
-   * List of users that were originaly requested to be in the conversation
-   * but could not be added due to their backend not being available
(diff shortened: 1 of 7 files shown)


packages/api-client/src/conversation/Conversation.ts
```diff
@@ -98,15 +98,6 @@ export interface Conversation {
...
 
-  /**
-   * List of users that were originaly requested to be in the conversation
-   * but could not be added due to their backend not being available
-   * @note This field comes only on conversation creation response
-   * @note Added since version 4: https://staging-nginz-https.zinfra.io/v4/api/swagger-ui/#/default/post_conversations
-   * @note Federation only
-   */
-  failed_to_add: QualifiedId[];
-
```
atomrc, line 109: This metadata did not belong in the conversation object.

PR #5409, opened 2023-08-21
refactor: handle decryption backend events

PatrykBuniX:
Simplifies the flow of handling proteus/mls decrypted events. 
- check the event type first and then run function responsible for handling it. Previously we were always running the method and returning undefined if type was not matching.
- `ConversationService` is now responsible for handling conversation events (`conversation.otr-message-add`, `conversation.mls-message-add`, `conversation.mls-welcome`). It will forward handling the event to proteus/mls services accordingly, therefore `NotificationService` is not aware of mls/proteus services anymore and it's only dependency is `ConversationService`
- check if `notificationService.handleEvent` is called with dryRun before running any mls/proteus handlers
- adds test scenarios for `handleMLSMessageAdd` handler


packages/core/src/notification/NotificationService.ts
```diff
@@ -231,42 +229,25 @@ export class NotificationService extends TypedEventEmitter<Events> {
...
-          throw new Error('no conv');
-        }
+      case CONVERSATION_EVENT.MLS_MESSAGE_ADD:
+        return this.conversationService.handleMLSMessageAddEvent(event);
+
+      case CONVERSATION_EVENT.MLS_WELCOME_MESSAGE:
+        return this.conversationService.handleMLSWelcomeMessageEvent(event);
+
+      case CONVERSATION_EVENT.OTR_MESSAGE_ADD:
+        return this.conversationService.handleOtrMessageAddEvent(event);
```
PatrykBuniX, line 250: It makes more sense to me to check the event type first and then handle the particular event type, than run a function responsible for handling it without even knowing the type yet.

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -477,4 +483,16 @@ export class ConversationService {
       return this.establishMLS1to1Conversation(groupId, selfUser, otherUserId);
     }
   };
+
+  public async handleMLSMessageAddEvent(event: ConversationMLSMessageAddEvent) {
+    return handleMLSMessageAdd({event, mlsService: this.mlsService});
+  }
```
PatrykBuniX, line 489: comparing to `proteusService.handleOtrMessageAddEvent` function call below, `handleMLSMessageAdd` function (module) is not called directly in MLSService as we will need to inject `ConversationService` there -we need it for recovery from epoch mismatch when decrypting a message - conversation service knows how to join with external proposal by fetching group info first with its id.

packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -631,17 +629,6 @@ export class MLSService extends TypedEventEmitter<Events> {
...
 
-  public async handleEvent(params: Omit<EventHandlerParams, 'mlsService'>): EventHandlerResult {
-    return handleBackendEvent({...params, mlsService: this}, async groupId => {
-      const conversationExists = await this.conversationExists(groupId);
-      if (!conversationExists) {
-        return;
-      }
-      const newEpoch = await this.getEpoch(groupId);
-      this.emit('newEpoch', {groupId, epoch: newEpoch});
-    });
```
PatrykBuniX, line 642: `newEpoch` event is now emmited in `handleMLSMessageAdd` directly, no need to pass it as callback function

packages/core/src/notification/NotificationService.ts
atomrc: hummm actually, I'd think the other way around, the `NotificationService` should not be aware of the different types of events that come it. It should be as dumb as possible and the different handlers should be the one responsible for switching accros the different cases. 

I'd move this logic to the `conversationService` and make the `handle*` methods private (so that it doesn't leak the implementation). 

WDYT?

packages/core/src/notification/NotificationService.ts
```diff
@@ -231,42 +229,25 @@ export class NotificationService extends TypedEventEmitter<Events> {
...
+      // We just return the raw event to the caller
+      return {event};
     }
 
-    // Fallback to other events
     switch (event.type) {
-      // Meta events
-      case CONVERSATION_EVENT.MEMBER_JOIN:
-        // As of today (07/07/2022) the backend sends `WELCOME` message to the user's own conversation (not the actual conversation that the welcome should be part of)
-        // So in order to map conversation Ids and groupId together, we need to first fetch the conversation and get the groupId linked to it.
```
atomrc, line 263: This is outdated, right?

PatrykBuniX: correct

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -478,4 +487,29 @@ export class ConversationService {
       return this.establishMLS1to1Conversation(groupId, selfUser, otherUserId);
     }
   };
+
+  private async handleMLSMessageAddEvent(event: ConversationMLSMessageAddEvent) {
+    return handleMLSMessageAdd({event, mlsService: this.mlsService});
```
atomrc, line 492: In the future, we should remove this injection of the entire `mlsService` to the message handler, there is something a little wrong there. (not for this PR though). 

I feel we have an architectural problem, the handler should not access the entire MLS Service, we should try to find a way to limit what we inject in those. 

It feels weird that handling mls event doesn't go **first** to the `mlsService` that then redirects to the correct handler, that should be the ideal data flow

atomrc (approved):
Nice job 💪

PR #5412, opened 2023-08-22
feat: mls decrypt error recovery [WPB-523]

PatrykBuniX:
Adds recovery mechanism to mls message decryption failure. 

When `corecrypto.decryptMessage` fails with "WrongEpoch" error, it means that the client got out of sync - missed some handshake messages - and is receiving messages from a new epoch it is not a member of. 

At this point we should fetch conversation from backend compare local epoch number with the epoch number on the remote conversation and if there's no match - rejoin conversation with external commit.

This PR also extends `ConversationService` with event emitter. `MLSConversationRecovered` event is emitted after client has successfully rejoined a group. Consumer should subscribe to this event and react to it by inserting a system message into the chat (this will be handled by the webapp)

### DIFF
--- a/packages/core/src/conversation/ConversationService/ConversationService.ts
+++ b/packages/core/src/conversation/ConversationService/ConversationService.ts
@@ -26,6 +26,7 @@ import {
   ConversationProtocol,
   RemoteConversations,
   PostMlsMessageResponse,
+  MLSConversation,
 } from '@wireapp/api-client/lib/conversation';
 import {CONVERSATION_TYPING, ConversationMemberUpdateData} from '@wireapp/api-client/lib/conversation/data';
 import {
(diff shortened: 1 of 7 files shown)


packages/core/src/messagingProtocols/mls/types.ts
```diff
@@ -84,9 +83,3 @@ export interface CryptoProtocolConfig {
   /** if set to true, will use experimental proteus encryption/decryption library (core-crypto). If not set will fallback to the legacy proteus library (cryptobox) */
   proteus?: boolean;
 }
-
-export type MLSConversation = Conversation & {
```
PatrykBuniX, line 88: This one was duplicated, see https://github.com/wireapp/wire-web-packages/blob/main/packages/api-client/src/conversation/Conversation.ts#L104

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -430,21 +437,43 @@ export class ConversationService {
...
-          this.logger.log(
-            `Conversation (id ${qualifiedId.id}) was not established or it's epoch number was out of date, joining via external commit`,
-          );
-          await this.joinByExternalCommit(qualifiedId);
-        }
-      } catch (error) {
-        this.logger.error(
-          `There was an error while handling epoch mismatch in MLS conversation (id: ${qualifiedId.id}):`,
-          error,
+    await Promise.allSettled(
```
atomrc, line 440: Is there a particular reason why we use `allSettled`. Do we expect some to fail?

atomrc (approved):
One little question, but looks quite solid overall 👌

packages/core/src/conversation/ConversationService/ConversationService.ts
PatrykBuniX: I didn't want one failure stop handling other conversations, but actually all logic inside `handleConversationEpochMismatch` is wrapped with try/catch, we can simply use `Promise.all` ;) will update it.

PR #5448, opened 2023-08-31
chore: reexport types from corecrypto

PatrykBuniX:
Re-export `Ciphersuite` and `CredentialType` types from `core-crypto` so consumer (webapp) doesn't need to be aware of `core-crypto` lib.


packages/core/src/messagingProtocols/mls/index.ts
```diff
@@ -18,3 +18,4 @@
  */
 
 export * from './MLSService/';
+export * from './types';
```
PatrykBuniX, line 21: So we can import it 
`import {...} from '@wireapp/core/messagingProtocols/mls'`
instead of 
`import {...} from '@wireapp/core/messagingProtocols/mls/types'`

PR #5463, opened 2023-09-04
refactor: mls conversation exists

PatrykBuniX:
Splits the concerns of mls group just existing and being established:
- Group exists if `coreCrypto.conversationExists` returns true, the epoch number doesn't matter, could be `0` (initial number after group creation) or more (after some users were added or just key material was updated).
- We say that the group is established when it exists (see point above) and epoch number is at least 1 - it means that at least one commit was sent in the group. There's a new method that checks that on `MLSService`: `mlsService.isConversationEstablished`.

If group is established (epoch number is at least `1`) we know that backend is aware of this group since the commit bundle needed to be successfully uploaded, processed by backend and accepted by one of the members. If group only exists (with epoch `0`) it's possible that backend does not know about such a group.

A check for mls group being already established will for example help in establishing 1:1 conversation to decide whether we should register a conversation or we should join it with external commit as it was already established by somebody else.


atomrc (approved):
Seems like a valuable distinction 🙏

PR #5495, opened 2023-09-13
runfix: upload kp after upgrade to mls device

PatrykBuniX:
### Problem
It can happen that the client the user is using was using only proteus initially, and later after MLS feature was enabled, it has upgraded to mls (by calling `CoreCrypto.mlsInit`). 

Previous flow of registering/initialising mls client goes as follows:
- check if corecrypto client exists (doesn't matter if it support mls already or not)
    - if it exists, call mlsService.initClient()
    - if it does not exist, call mlsService.registerClient()

Only `.registerClient()` method included uploading a public key and key packages to backend. When client did already exist previously, but mls client was not yet created, we were calling only `.mlsInit()` without uploading key packages / public key to backend. Such client ended up being mls device without any key packages uploaded to backend, therefore other clients (users) were not able to send a welcome message to this client.

---

### Solution
Now `.initClient()` method contain the logic of uploading key packages and updating a client entity with public key. Since we initialise the client every time webapp is reloaded, there's now check included before uploads, whether the client has not previously uploaded public key already and if it has enough available key packages uploaded to backend.

So every time we initialise mls client:
- public key will get uploaded, but **only if it was not uploaded previously**
- key packages will get uploaded, but only if the number of uploaded key packages is below minimum required number, which is now set to half of the `nbKeyPackages` value in mls service config.

### DIFF
--- a/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
+++ b/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
@@ -618,7 +618,7 @@ export class MLSService extends TypedEventEmitter<Events> {
   }
 
   /**
-   * Will update current client on backend with its public key.
+   * Will update the given client on backend with its public key.
    *
    * @param mlsClient Intance of the coreCrypto that represents the mls client
    * @param clientId The id of the client
--- a/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
+++ b/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
@@ -117,11 +117,7 @@ export class MLSService extends TypedEventEmitter<Events> {
     await this.initClient(userId, clientId);
     // If the device is new, we need to upload keypackages and public key to the backend
     const publicKey = await this.coreCryptoClient.clientPublicKey(this.config.defaultCiphersuite);
-    const keyPackages = await this.coreCryptoClient.clientKeypackages(
-      this.config.defaultCiphersuite,
-      this.config.defaultCredentialType,
-      this.config.nbKeyPackages,
-    );
+    const keyPackages = await this.clientKeypackages(this.config.nbKeyPackages);
     await this.uploadMLSPublicKeys(publicKey, clientId);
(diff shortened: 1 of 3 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -66,6 +66,13 @@ export const optionalToUint8Array = (array: Uint8Array | []): Uint8Array => {
   return Array.isArray(array) ? Uint8Array.from(array) : array;
 };
 
+interface LocalMLSServiceConfig extends MLSServiceConfig {
+  /**
+   * minimum number of key packages client should have available (configured to half of nbKeyPackages)
+   */
+  minRequiredNumberOfAvailableKeyPackages: number;
+}
```
PatrykBuniX, line 74: This is the config used by MLSService, it contains field that are not configurable by consumer (we set them internally).

packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -607,42 +606,49 @@ export class MLSService extends TypedEventEmitter<Events> {
...
   /**
-   * Will make the given client mls capable (generate and upload key packages)
+   * Will update the given client on backend with its public key.
    *
    * @param mlsClient Intance of the coreCrypto that represents the mls client
    * @param clientId The id of the client
    */
-  private async uploadMLSPublicKeys(publicKey: Uint8Array, clientId: string) {
+  private async uploadMLSPublicKeys(clientId: string) {
+    const existingPublicKey = await this.apiClient.api.client.getClient(clientId);
```
atomrc, line 627: question: Since we have the client at this point https://github.com/wireapp/wire-web-packages/pull/5495/files#diff-0a7e6d4a8ee3b5e3153dbc2336fecd1741ea0f39e7226d6b7a5d63be31b2747fR267
maybe it would be better to inject it, rather than re-fetching (but maybe there is another codepath I'm not seeing)?

atomrc (approved):

PR #5497, opened 2023-09-14
feat: recover from mls stale message error

PatrykBuniX:
Adds a recovery mechanism to mls message sending when failed with `"mls-stale-message"` error. When error is catched, we should rejoin the group with external commit, retry to send a message once and notify consumer (webapp) that mls group has recovered from epoch mismatch. Consumer will then insert system message into the recovered conversation.

### DIFF
--- a/packages/core/src/conversation/ConversationService/ConversationService.ts
+++ b/packages/core/src/conversation/ConversationService/ConversationService.ts
@@ -37,8 +37,7 @@ import {
   ConversationMemberLeaveEvent,
   ConversationOtrMessageAddEvent,
 } from '@wireapp/api-client/lib/event';
-import {BackendError} from '@wireapp/api-client/lib/http';
-import {BackendErrorLabel} from '@wireapp/api-client/lib/http';
+import {BackendError, BackendErrorLabel} from '@wireapp/api-client/lib/http';
 import {QualifiedId} from '@wireapp/api-client/lib/user';
 import {XOR} from '@wireapp/commons/lib/util/TypeUtil';
 import {Decoder} from 'bazinga64';
--- a/packages/core/src/conversation/ConversationService/ConversationService.ts
+++ b/packages/core/src/conversation/ConversationService/ConversationService.ts
@@ -37,6 +37,8 @@ import {
   ConversationMemberLeaveEvent,
   ConversationOtrMessageAddEvent,
 } from '@wireapp/api-client/lib/event';
+import {BackendError} from '@wireapp/api-client/lib/http';
+import {BackendErrorLabel} from '@wireapp/api-client/lib/http';
 import {QualifiedId} from '@wireapp/api-client/lib/user';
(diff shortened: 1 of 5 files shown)


aweiss-dev (approved):

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -37,6 +37,8 @@ import {
   ConversationMemberLeaveEvent,
   ConversationOtrMessageAddEvent,
 } from '@wireapp/api-client/lib/event';
+import {BackendError} from '@wireapp/api-client/lib/http';
+import {BackendErrorLabel} from '@wireapp/api-client/lib/http';
```
tlebon, line 41: combine these lines

PatrykBuniX: https://github.com/wireapp/wire-web-packages/pull/5497/commits/b6229be4e2f2e7589063a7cd27f2e55cf7bd11c2

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -570,21 +581,25 @@ export class ConversationService extends TypedEventEmitter<Events> {
...
   }
 
+  private async recoverMLSConversationFromEpochMismatch(conversationId: QualifiedId) {
+    const mlsConversation = await this.apiClient.api.conversation.getConversation(conversationId);
+
+    if (!isMLSConversation(mlsConversation)) {
+      throw new Error('Conversation is not an MLS conversation');
+    }
+
+    return this.handleConversationEpochMismatch(mlsConversation, () =>
```
tlebon, line 598: should this still await also?

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -570,21 +581,25 @@ export class ConversationService extends TypedEventEmitter<Events> {
...
-
-        if (!isMLSConversation(mlsConversation)) {
-          throw new Error('Conversation is not an MLS conversation');
-        }
-
-        await this.handleConversationEpochMismatch(mlsConversation, () =>
-          this.emit('MLSConversationRecovered', {conversationId}),
-        );
+        await this.recoverMLSConversationFromEpochMismatch(conversationId);
         return;
```
tlebon, line 585: same comment here. return await?

PatrykBuniX: Simple await is enough, we don't expect to return anything. Will remove redundant `return` keyword

packages/core/src/conversation/ConversationService/ConversationService.ts
PatrykBuniX: Nope, simple return is enough. Since we're returning async method here, it will get awaited anyway when parent method is awaited.

tlebon (approved):

atomrc (approved):
Looking good :)

PR #5517, opened 2023-09-20
feat: refill check on welcome message [WPB-4773]

PatrykBuniX:
We (and other platforms) were missing a check for amount of local key packages after receiving a welcome message. We were performing a check only once every 24h. This PR adds this functionality.

Explanation:
When somebody tries to add you to a MLS group, they need to first claim one of your key packages you've uploaded to backend when registering MLS device. After successfully claiming your key package, you're being sent a welcome message. After welcome message is processed, corecrypto "consumes" your key package. Minimal number of key packages client should have uploaded to backend is half of the default amount of keys which is 100 by default (so minimal required number is 50 by default).

We need to keep track of the local amount of keypackages, to know when we should generate new keys and upload them to backend.

Example:
- we have 50 keys locally, they are all uploaded to backend
- somebody adds us to mls group, using our key package
- number of keys on backend reduces to 49
- we receive a welcome message, corecrypto processes it and reduces number of local keys to 49
- we detect that we are below minimal required number of keys (50)
- we generate new keys to top up to 100
- we upload all the newly generated keys to backend
- we have 100 keys both locally and on backend

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -305,7 +305,7 @@ export class Account extends TypedEventEmitter<Events> {
       await this.service.mls.checkExistingPendingProposals();
 
       // initialize scheduler for syncing key packages with backend
-      this.service.mls.checkForKeyPackagesBackendSync();
+      this.service.mls.schedulePeriodicKeyPackagesBackendSync(validClient.id);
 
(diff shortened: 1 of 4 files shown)


atomrc (approved):
seems solid 😎

PR #5519, opened 2023-09-22
refactor(api-client): infer feature update event data object from feature key

PatrykBuniX:
This will allow us to narrow down the data property with feature name (key) on team feature config update event. 

Also we don't need to maintain `TeamFeatureConfigurationUpdateEventData` type anymore - we can just relay on FeatureList type that contains all the features already.

Before:
We cannot even access data property on the event.
Screenshot 2023-09-22 at 11 48 39

After:
We can type-safely access `data.config` property on the event. It contains actual feature config based on `event.name`.
Screenshot 2023-09-22 at 11 49 16


thisisamir98 (approved):

atomrc (commented):
super nice one 😎 
Good job man

PR #5520, opened 2023-09-22
feat: lock websocket when uploading a commit bundle to backend

PatrykBuniX:
### Problem

There's undesired behaviour on backend when sending a commit bundle with external commit (e.g when joining mls groups you are member of after registering completely new client). When request is being processed (we've not received a response from backend yet), it can happen that we will receive a bunch of messages on different a "channel" (websocket in this case). Those messages are usually mls messages we cannot decrypt anyway since external join (external commit) was not fully processed, and we're not yet a group member locally.

Current behaviour:
- we send a commit bundle to backend
- while request is being processed we receive some messages via websocket we cannot decrypt anyway, we're thrown a bunch of errors from core-crypto
- response from backend finally arrives, but we don't have access to messages that were received in the meantime (it's possible we're out of sync due to lost mesages)

### Solution

The solution is to simply lock websocket when we're awaiting commit bundle request. When websocket is locked, messages are buffered in the queue and waiting for websocket to be unlocked. After it's unlocked it processes them one by one in order they were received.

New behaviour:
- we lock websocket (buffer incoming messages)
- we send a commit bundle to backend (there're messages arriving but we do not process them yet, they're buffered)
- we got a response from backend
- we accept/clear the commit we've sent locally based on the response code
- we unlock websocket
- we process messages that were queued
- we're all in sync

### DIFF
--- a/packages/api-client/src/APIClient.ts
+++ b/packages/api-client/src/APIClient.ts
@@ -408,4 +408,17 @@ export class APIClient extends EventEmitter {
     }
     throw new Error('No valid client ID.');
   }
+
+  /**
+   * Will lock the websocket before executing the callback and unlock it after
+   * @param callback The callback to execute
+   */
+  public withLockedWebSocket<T>(callback: () => T): T {
+    this.transport.ws.lock();
+    try {
+      return callback();
+    } finally {
+      this.transport.ws.unlock();
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


packages/api-client/src/APIClient.ts
```diff
@@ -408,4 +408,21 @@ export class APIClient extends EventEmitter {
...
+  ): (...args: Parameters<T>) => ReturnType<T> {
+    return (...args: Parameters<T>): ReturnType<T> => {
+      this.transport.ws.lock();
+      try {
+        return callback(...args);
+      } finally {
+        this.transport.ws.unlock();
+      }
+    };
+  }
```
atomrc, line 427: oooh pretty nice higher order function. I like the solution 👏

atomrc (approved):

PR #5551, opened 2023-09-26
feat: move TypedEventEmitter to commons package

PatrykBuniX:
Move TypedEventEmitter module from core package's utils to `commons` package.


packages/commons/src/TypedEventEmitter/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export * from './TypedEventEmitter';
```
atomrc, line 20: Suggestion: How about also exporting it at the root of the `commons` package? 
So that we could import it directly with 

```
import {TypedEventEmitter} from '@wireapp/commons'
```

atomrc (approved):
sweet

packages/commons/src/TypedEventEmitter/index.ts
PatrykBuniX: added with https://github.com/wireapp/wire-web-packages/pull/5551/commits/2df53eae5d7b27cb67d7984a4ca403951d80d84d

PR #5572, opened 2023-10-02
feat: specify initial nb of key packages in corecrypto

PatrykBuniX:
Bumps core-crypto with version that:
- fixes one edge case with handling one edge case with rejoining a conversation with external commit after getting out of sync with epoch number
- allows consumer (core in our case) to specify initial number of key packages to be created when calling `CoreCrypto.mlsInit` method (previously it was always `100`)

### DIFF
--- a/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
+++ b/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
@@ -118,7 +118,11 @@ export class MLSService extends TypedEventEmitter<Events> {
 
   public async initClient(userId: QualifiedId, client: RegisteredClient) {
     const qualifiedClientId = constructFullyQualifiedClientId(userId.id, client.id, userId.domain);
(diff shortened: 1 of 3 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -118,7 +118,11 @@ export class MLSService extends TypedEventEmitter<Events> {
 
   public async initClient(userId: QualifiedId, client: RegisteredClient) {
     const qualifiedClientId = constructFullyQualifiedClientId(userId.id, client.id, userId.domain);
-    await this.coreCryptoClient.mlsInit(this.textEncoder.encode(qualifiedClientId), [this.config.defaultCiphersuite]);
+    await this.coreCryptoClient.mlsInit(
+      this.textEncoder.encode(qualifiedClientId),
+      [this.config.defaultCiphersuite],
+      this.config.nbKeyPackages,
```
PatrykBuniX, line 124: Now consumer can specify how many keys corecrypto should generate initially when registering a new client.

atomrc (approved):

PR #5573, opened 2023-10-03
runfix: store pending proposals in core store

PatrykBuniX:
Moves responsibility of storing pending proposals data from `localStorage` to core's indexedDB instance.

Pending proposals data mentioned above contains the list of objects with data:
- `groupId` - id of a mls group
- `firingDate` - time when pending proposals should be committed

Previously we were storing this information in `localStorage` under `pendingProposals` key. The problem with this solution was that there's one localStorage for all the users (clients) that log-in into the same browser. It could happen that after logging off from user A (that had some pending proposals stored in localStorage) to user B (participating completely different conversations), user B would try to commit pending proposals of user A's groups. Since `B` knows anything about those groups, corecrypto would throw error `"Conversation not found"`.

Since indexedDB is tied to a user, this problem will not occur anymore.


packages/core/src/storage/CoreDB.ts
```diff
@@ -35,6 +39,9 @@ export async function openDB(dbName: string): Promise<CoreDatabase> {
       switch (oldVersion) {
         case 0:
           db.createObjectStore('prekeys');
+        case 1:
+          db.deleteObjectStore('prekeys');
```
PatrykBuniX, line 43: We've never been using this store.

atomrc: neat 👌

packages/core/src/messagingProtocols/mls/MLSService/MLSService.test.ts
atomrc, line 1: question: Do we have any test that covers pending proposal in there? If not, maybe could be worth writing one

atomrc (approved):
Looks solid 👌

packages/core/src/messagingProtocols/mls/MLSService/MLSService.test.ts
PatrykBuniX: I don't think we do, let me prepare little something :D

PatrykBuniX: https://github.com/wireapp/wire-web-packages/pull/5573/commits/d7b47d6e225f3dca7b881525c627c112ab94d860 added two scenarios 😉

atomrc (approved):

PR #5575, opened 2023-10-04
runfix: recurring task scheduler not resumed after reload

PatrykBuniX:
### Problem
Recurring task scheduler was never resuming a task after app was reloaded. We were storing only last fire date, which simply didn't exist before the task was run for the first time.

### Solution
We store `firingDate` instead and make sure it's stored even before the task runs for the first time. If `firingDate` already exists, we just load it from the store.

### DIFF
--- a/packages/core/src/util/LowPrecisionTaskScheduler/LowPrecisionTaskScheduler.ts
+++ b/packages/core/src/util/LowPrecisionTaskScheduler/LowPrecisionTaskScheduler.ts
@@ -52,11 +52,11 @@ const addTask = ({key, firingDate, task, intervalDelay}: ScheduleLowPrecisionTas
     const tasks = intervals[intervalDelay]?.tasks;
     if (tasks?.length !== 0) {
       for (const taskData of tasks) {
-        if (nowTime >= firingDate) {
+        if (nowTime >= taskData.firingDate) {
           const {task, key} = taskData;
           logger.info(`Executing task with key "${key}"`);
-          task();
           cancelTask({intervalDelay, key});
+          task();
         }
       }
     }
--- a/packages/core/src/util/RecurringTaskScheduler/RecurringTaskScheduler.ts
+++ b/packages/core/src/util/RecurringTaskScheduler/RecurringTaskScheduler.ts
@@ -19,7 +19,7 @@
 
 import {TimeUtil} from '@wireapp/commons';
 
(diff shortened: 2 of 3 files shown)


packages/core/src/util/LowPrecisionTaskScheduler/LowPrecisionTaskScheduler.ts
```diff
@@ -52,11 +52,11 @@ const addTask = ({key, firingDate, task, intervalDelay}: ScheduleLowPrecisionTas
     const tasks = intervals[intervalDelay]?.tasks;
     if (tasks?.length !== 0) {
       for (const taskData of tasks) {
-        if (nowTime >= firingDate) {
+        if (nowTime >= taskData.firingDate) {
```
PatrykBuniX, line 55: We should compare with `firingDate` of each task separately, not with with `firingDate` that is passed via props when adding a new task (it would be the same for each task we check in the loop)

packages/core/src/util/LowPrecisionTaskScheduler/LowPrecisionTaskScheduler.ts
```diff
@@ -52,11 +52,11 @@ const addTask = ({key, firingDate, task, intervalDelay}: ScheduleLowPrecisionTas
...
           const {task, key} = taskData;
           logger.info(`Executing task with key "${key}"`);
-          task();
           cancelTask({intervalDelay, key});
+          task();
```
PatrykBuniX, line 59: Moved execution of task after canceling the current one in case where the task itself contained the logic of re-scheduling itself.

packages/core/src/util/RecurringTaskScheduler/RecurringTaskScheduler.ts
```diff
@@ -31,13 +31,14 @@ interface TaskParams {
 }
 
 export function registerRecurringTask({every, task, key}: TaskParams) {
-  const lastFireDate = loadState(key) ?? Date.now();
+  const firingDate = loadState(key) || Date.now() + every;
+  saveState(key, firingDate);
```
PatrykBuniX, line 35: Previously we were never storing fire date before the task was fired for the first time. Therefore it would restart to `Date.now() + every` on each app reload. User reloading their app frequently could make the task never being executed.

PR #5576, opened 2023-10-04
runfix: recurring task scheduler with indexeddb store

PatrykBuniX:
Makes recurring task scheduler more generic and allows its consumer to specify storage where events will be stored.

Now `core` has its own recurring task scheduler that is using core's database (indexedDB) to store the events.


atomrc (approved):
Seems solid. Now is a good time to migrate all those to indexeddb, cause that means we won't have to write a migration script (since no production code runs it currently) 😎

PR #5578, opened 2023-10-06
feat: try establishing mls group for mixed conversation

PatrykBuniX:
Adds a method for trying to establish a mls group for mixed conversation. This logic did previously live in webapp, but since it's plain core/mls logic, it can be moved to core package's `ConversationService`.

### DIFF
--- a/packages/core/src/conversation/ConversationService/ConversationService.ts
+++ b/packages/core/src/conversation/ConversationService/ConversationService.ts
@@ -576,7 +576,7 @@ export class ConversationService extends TypedEventEmitter<Events> {
    * @param selfUserId - id of the self user
    * @param qualifiedUsers - list of qualified users to add to the group (should not include the self user)
    */
-  public async tryEstablishingMLSGroupWithUsers({
+  public async tryEstablishingMLSGroup({
     groupId,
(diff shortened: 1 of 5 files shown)


packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -566,6 +567,53 @@ export class ConversationService extends TypedEventEmitter<Events> {
...
+  /**
+   * Will try to register mls group by sending an empty commit to establish it.
+   * After group was successfully established, it will try to add other users to the group.
+   *
+   * @param groupId - id of the MLS group
+   * @param conversationId - id of the conversation
+   * @param selfUserId - id of the self user
+   * @param qualifiedUsers - list of qualified users to add to the group (should not include the self user)
+   */
+  public async tryEstablishingMLSGroupWithUsers({
```
atomrc, line 579:
```suggestion
  public async tryEstablishingMLSGroup({
```

Let's shorten the naming a little. It's explicit that you can add users since it's in the param payload

atomrc (approved):

atomrc (approved):

PR #5637, opened 2023-10-17
feat: ignore events from a blacklisted conversation

PatrykBuniX:
Ignores events from conversations that were blacklisted.


packages/core/src/conversation/ConversationService/Utility/isConversationEvent.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+import {BackendEvent, ConversationEvent} from '@wireapp/api-client/lib/event';
+
+const CONVERSATION_EVENT_CATEGORY = 'conversation';
+
+export const isConversationEvent = (event: BackendEvent): event is ConversationEvent => {
+  const [category] = event.type.split('.');
+  return category === CONVERSATION_EVENT_CATEGORY;
+};
```
atomrc, line 27: I know we do those kind of checks in the webapp, but I think they actually can be harmful. 
We don't have any guarantee that there are events in the future that do not start with `conversation` that still need to be processed by the conversation service. 

Let's not try to filter events before hands, let's just react at all the ones we want to without any pre-filter. This is going to scale way better in the future :)

atomrc (approved):

PR #5662, opened 2023-10-23
feat: check if removal key is available on backend

PatrykBuniX:
Before initialising mls service, check whether a removal key is available on backend.

### DIFF
--- a/packages/api-client/src/APIClient.ts
+++ b/packages/api-client/src/APIClient.ts
@@ -427,7 +427,7 @@ export class APIClient extends EventEmitter {
   }
 
   /**
-   * Will check if MLS is supported by backend (whether used api version supports MLS and backend removal key is present).
+   * Will check if MLS is supported by backend (whether the api version used supports MLS, and whether a backend removal key is present).
    */
   public async isMLSSupportedByBackend(): Promise<boolean> {
     const isMLSSupportedByBackend = this.backendFeatures.supportsMLS;
--- a/packages/api-client/src/APIClient.ts
+++ b/packages/api-client/src/APIClient.ts
@@ -429,10 +429,10 @@ export class APIClient extends EventEmitter {
   /**
    * Will check if MLS is supported by backend (whether the api version used supports MLS, and whether a backend removal key is present).
    */
-  public async isMLSSupportedByBackend(): Promise<boolean> {
-    const isMLSSupportedByBackend = this.backendFeatures.supportsMLS;
+  public async supportsMLS(): Promise<boolean> {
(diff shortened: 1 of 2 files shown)


packages/core/src/Account.ts
```diff
@@ -582,7 +583,7 @@ export class Account extends TypedEventEmitter<Events> {
     });
 
     const handleMissedNotifications = async (notificationId: string) => {
-      if (this.cryptoProtocolConfig?.mls && this.backendFeatures.supportsMLS) {
+      if (this.service?.mls) {
```
PatrykBuniX, line 586: It's enough to check if mls service exists - it wouldn't get initialised if mls was disabled.

packages/api-client/src/APIClient.ts
```diff
@@ -425,4 +425,23 @@ export class APIClient extends EventEmitter {
       }
     };
   }
+
+  /**
+   * Will check if MLS is supported by backend (whether used api version supports MLS and backend removal key is present).
```
V-Gira, line 430: Suggestion:
```suggestion
   * Will check if MLS is supported by backend (whether the api version used supports MLS, and whether a backend removal key is present).
```

packages/api-client/src/APIClient.ts
```diff
@@ -425,4 +425,23 @@ export class APIClient extends EventEmitter {
...
+      return false;
+    }
+
+    let isBackendRemovalKeyPresent = false;
+    try {
+      const backendRemovalKey = (await this.api.client.getPublicKeys()).removal;
+      isBackendRemovalKeyPresent = !!backendRemovalKey;
+    } catch {}
+
+    return isBackendRemovalKeyPresent;
```
atomrc, line 445: suggestion :) 

```suggestion
    try {
      const backendRemovalKey = (await this.api.client.getPublicKeys()).removal;
      return !!backendRemovalKey;
    } catch {}

    return false;
```

packages/api-client/src/APIClient.ts
```diff
@@ -425,4 +425,23 @@ export class APIClient extends EventEmitter {
       }
     };
   }
+
+  /**
+   * Will check if MLS is supported by backend (whether used api version supports MLS and backend removal key is present).
+   */
+  public async isMLSSupportedByBackend(): Promise<boolean> {
```
atomrc, line 432:
```suggestion
  public async supportsMLS(): Promise<boolean> {
```

We know we are dealing with the APIClient so no need for mentionning the backend at all in the name of the method. It's already explicit from the context

atomrc (approved):
LGTM. Just a few suggestions

PR #5664, opened 2023-10-24
refactor: move subconversations logic to separate service

PatrykBuniX:
Moves subconversations related logic to new `SubconversationService`. 
Also adds some logic that previously lived on webapp side.

Added a bunch of test cases.

### DIFF
--- a/packages/core/src/conversation/SubconversationService/SubconversationService.ts
+++ b/packages/core/src/conversation/SubconversationService/SubconversationService.ts
@@ -1,0 +1,165 @@
+/*
+ * Wire
+ * Copyright (C) 2023 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -284,100 +276,6 @@ export class MLSService extends TypedEventEmitter<Events> {
     return mlsResponse;
   }
 
-  public async getConferenceSubconversation(conversationId: QualifiedId): Promise<Subconversation> {
```
PatrykBuniX, line 287: This all was moved to `SubconversationService`. All the logic related with subconversaitons lives there now.

packages/core/src/conversation/SubconversationService/SubconversationService.ts
```diff
@@ -0,0 +1,332 @@
...
+        epoch: subconversationEpoch,
+        epoch_timestamp: subconversationEpochTimestamp,
+        subconv_id: subconversationId,
+      } = await this.getConferenceSubconversation(conversationId);
+
+      if (subconversationEpoch === 0) {
+        const doesConversationExistsLocally = await this.mlsService.conversationExists(subconversationGroupId);
+        if (doesConversationExistsLocally) {
+          await this.mlsService.wipeConversation(subconversationGroupId);
+        }
```
PatrykBuniX, line 83: This is the only part that was added - we need to make sure local group is wiped before we try to register it again.

atomrc (approved):

PR #5676, opened 2023-10-30
refactor: core callbacks and subconversations store

PatrykBuniX:
Moves subconversation storage to core's database (indexedDB). Changes the way core callbacks are initialised by webapp.


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -120,6 +119,13 @@ export class MLSService extends TypedEventEmitter<Events> {
       this.config.nbKeyPackages,
     );
 
+    await this.coreCryptoClient.registerCallbacks({
+      // All authorization/membership rules are enforced on backend
+      clientIsExistingGroupUser: async () => true,
+      authorize: async () => true,
+      userAuthorize: async () => true,
```
PatrykBuniX, line 126: These were previously provided by webapp, but since the rules are enforced by backend anyway, we can move them to the core.

atomrc (approved):

PR #5735, opened 2023-11-22
feat: block 'default mls key packages' - upload for new devices when E2EI is enabled

aweiss-dev:
With enabled E2EI, uploading default (non x509) key packages would degrade all conversations that the user is part of.

We need to block the upload of these and force the user to enroll.

Hint: Forcing the user to enroll is not part of this branch, that will be done in the webapp


packages/core/src/Account.ts
```diff
@@ -328,6 +334,19 @@ export class Account extends TypedEventEmitter<Events> {
     return this.initClient(client);
   }
 
+  /**
+   * Will create a new MLS Client for the current user
+   */
+  private async initMLSClient(userId: QualifiedId, client: RegisteredClient): Promise<void> {
+    if (!this.service || !this.service.mls) {
```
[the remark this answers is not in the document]
aweiss-dev: Sure

packages/core/src/Account.ts
```diff
@@ -221,6 +222,11 @@ export class Account extends TypedEventEmitter<Events> {
     return storeEngine.updateOrCreate(AUTH_TABLE_NAME, AUTH_COOKIE_KEY, entity);
   }
 
+  private async getE2EIStatus(): Promise<boolean> {
```
PatrykBuniX, line 225: **suggestion**: wouldn't `isE2EIFeatureEnabled` method name make more sense since `boolean` is what we return, not the actual status.

PatrykBuniX (approved):

packages/core/src/Account.ts
aweiss-dev: I think it sounds like a boolean then. But it is a async function that we need to call.

PR #5736, opened 2023-11-23
feat: Map coreCrypto userIds to regular userIds

atomrc:
CoreCrypto is using base64 encoded strings to identity users. 
In order for consumers of the `core` not to have to implement the logic of mapping base64 encoded userIds to regular userIds, we map the userIds right before returning the identity.

### DIFF
--- a/packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
+++ b/packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
@@ -26,25 +26,13 @@ import {Ciphersuite, CoreCrypto, E2eiConversationState, WireIdentity} from '@wir
 import {getE2EIClientId, uuidTobase64url} from './Helper';
 import {E2EIStorage} from './Storage/E2EIStorage';
 
+import {parseFullQualifiedClientId} from '../../../util/fullyQualifiedClientIdUtils';
+
 // This export is meant to be accessible from the outside (e.g the Webapp / UI)
 export class E2EIServiceExternal {
-  private static instance: E2EIServiceExternal;
   private readonly logger = logdown('@wireapp/core/E2EIdentityServiceExternal');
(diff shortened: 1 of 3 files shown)


packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
```diff
@@ -78,21 +68,49 @@ export class E2EIServiceExternal {
...
+    const mappedUserIdentities = new Map();
+    for (const [base64Id, identities] of userIdentities) {
+      // remapping coreCrypto user ids to regular userIds
+      const userId = hashedIdMap.get(base64Id);
+      if (userId) {
+        mappedUserIdentities.set(
+          userId.id,
+          identities.map(identity => ({
+            ...identity,
+            deviceId: parseFullQualifiedClientId((identity as any).client_id).client,
```
atomrc, line 89: The types from CoreCrypto are wrong. The interface says the property is `clientId` but the payload actually contains `client_id` 🙃

packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
```diff
@@ -78,21 +68,49 @@ export class E2EIServiceExternal {
...
     );
-    return this.coreCryptoClient.getDeviceIdentities(Decoder.fromBase64(groupId).asBytes, clientIds);
+    const deviceIdentities = await this.coreCryptoClient.getDeviceIdentities(
+      Decoder.fromBase64(groupId).asBytes,
+      clientIds,
+    );
+
+    return deviceIdentities.map(identity => ({
+      ...identity,
+      deviceId: parseFullQualifiedClientId((identity as any).client_id).client,
```
atomrc, line 113: The idea here is to parse the `clientId` from coreCrypto and convert it to a format that the consumer (namely the webapp) can understand

PatrykBuniX (approved):

PR #5756, opened 2023-11-28
feat: Allow consumer to generate and store secret keys

atomrc:
This will add, for the consumer, the possibility to generate and store secret values (for encrypting database entries for example). 

The API works as follow:

```js
import {generateSecretKey} from '@wireapp/core/lib/secretStore/secretKeyGenerator';
const key = await generateSecretKey({
    keyId,
    keySize,
    dbName: `secrets-${storeName}`,
});
```

BREAKING CHANGE: the config object of `coreCrypto` has changed


packages/core/src/Account.ts
```diff
@@ -406,30 +406,30 @@ export class Account extends TypedEventEmitter<Events> {
...
-     * 1. MLS is enabled
-     * 2. The user has enabled CoreCrypto in the config
-     * 3. The user has already used CoreCrypto in the past (cannot rollback to using cryptobox)
-     */
-    const clientType =
-      (await this.isMlsEnabled()) ||
-      !!this.cryptoProtocolConfig?.useCoreCrypto ||
-      cryptoMigrationStore.coreCrypto.isReady(storeEngine.storeName)
-        ? CryptoClientType.CORE_CRYPTO
-        : CryptoClientType.CRYPTOBOX;
```
atomrc, line 419: This could be simplified, basically we init `CoreCrypto` as soon as this object is defined (no other conditions)

packages/core/src/messagingProtocols/mls/types.ts
```diff
@@ -58,25 +60,19 @@ export type CommitPendingProposalsParams = {
...
    * It, thus, needs to know where, on the server, the file can be found
    */
-  coreCrypoWasmFilePath: string;
+  wasmFilePath: string;
 
   /** If set will create an MLS capable device from the current device */
   mls?: Partial<MLSServiceConfig>;
-
-  /** if set to true, will use experimental proteus encryption/decryption library (core-crypto). If not set will fallback to the legacy proteus library (cryptobox) */
-  proteus?: boolean;
```
atomrc, line 81: This was not used anymore

PatrykBuniX (approved):

PR #5759, opened 2023-11-28
runfix: include clientId on last notification api call

PatrykBuniX:
Client id was not properly passed to `getLastNotification` method. `this.apiClient.clientId` was `undefined` at this point. Client was trying to fetch and process all the notifications (including mls messages) directed to user, not this new client. 

It could result in `"You have missed some messages"` system message being injected into all the conversations, event though the client is completely new and it should not receive those messages since the device is new and was not listed in message's recipients when the message was sent.

### DIFF
--- a/packages/core/src/notification/NotificationService.ts
+++ b/packages/core/src/notification/NotificationService.ts
@@ -96,8 +96,7 @@ export class NotificationService extends TypedEventEmitter<Events> {
   }
 
   /** Should only be called with a completely new client. */
-  public async initializeNotificationStream(): Promise<string> {
-    const clientId = this.apiClient.clientId;
+  public async initializeNotificationStream(clientId?: string): Promise<string> {
     await this.setLastEventDate(new Date(0));
     const latestNotification = await this.backend.getLastNotification(clientId);
     return this.setLastNotificationId(latestNotification);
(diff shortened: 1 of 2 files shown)


packages/core/src/notification/NotificationService.ts
```diff
@@ -96,8 +96,7 @@ export class NotificationService extends TypedEventEmitter<Events> {
   }
 
   /** Should only be called with a completely new client. */
-  public async initializeNotificationStream(): Promise<string> {
-    const clientId = this.apiClient.clientId;
```
PatrykBuniX, line 100: It was `undefined` at this point.

packages/core/src/notification/NotificationService.ts
```diff
@@ -96,8 +96,7 @@ export class NotificationService extends TypedEventEmitter<Events> {
...
+  public async initializeNotificationStream(clientId?: string): Promise<string> {
```
atomrc, line 99: Does it need to be optional?

atomrc (approved):

packages/core/src/notification/NotificationService.ts
PatrykBuniX: We can make it required, we call this method only once seems like and we do pass the id, good call 👌

PR #5764, opened 2023-11-29
fix(E2EI): stop removing necessary data while clearing temporary data

aweiss-dev:


packages/core/src/messagingProtocols/mls/E2EIdentityService/Storage/E2EIStorage.ts
```diff
@@ -88,11 +88,14 @@ const getCertificateData = (): string => {
   return atob;
 };
 
+const removeIntialData = () => {
```
atomrc, line 91:
```suggestion
const removeInitialData = () => {
```
typo?

PR #5782, opened 2023-12-06
fix: Use correct credential type according to e2ei enrollment status

atomrc:


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -134,6 +130,12 @@ export class MLSService extends TypedEventEmitter<Events> {
     }
   }
 
+  private async getCredentialType() {
+    return (await this.coreCryptoClient.e2eiIsEnabled(this.config.cipherSuite))
+      ? CredentialType.X509
+      : CredentialType.Basic;
+  }
```
atomrc, line 137: This fixes an issue where we need to dynamically change the creds type according to the e2ei enrollment status (this could not be static)

PR #5814, opened 2023-12-20
fix: Add MLS basic devices to the return value of user identities

atomrc:
## Description

the method `getUserIdentities` from CoreCrypto doesn't list the MLS basic devices in the return value. But we need to indicate those devices as `not downloaded` so we need to be aware of those. 

To circumvent this issue, we:
- ask CoreCrypto to give us all the devices known in the conversation (both with identities and basics)
- ask CoreCrypto the devices that have identities
- merge both values in order to add basic devices to the return value

## Pull Request Checklist


PatrykBuniX (approved):
Looks good 😄

PR #5815, opened 2023-12-21
feat: replace refreshActiveCertificate with hasActiveCertificate and initialise MLSService with e2eServiceExternal

arjita-mitra:
remove the refreshActiveCertificate dependency instead use CC hasActiveCertificate and initialise MLSService with e2eServiceExternal

## Pull Request Checklist


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -97,6 +97,7 @@ export class MLSService extends TypedEventEmitter<Events> {
       nbKeyPackages = defaultConfig.nbKeyPackages,
       cipherSuite = defaultConfig.cipherSuite,
     }: Partial<MLSServiceConfig>,
+    private readonly e2eServiceExternal: E2EIServiceExternal,
```
atomrc, line 100: as a convention we usually pass service early in the constructor's parameters. at least it should be before the config

atomrc (approved):

PR #5822, opened 2023-12-27
feat: add a queue for incoming mls messages

PatrykBuniX:
When uploading a commit bundle to backend, there's a chance we will receive a bunch of mls message before the response from this call is returned. Before that happens, we cannot decrypt those messages.

A previous solution for that was locking the whole websocket and unlocking it after receiving a response from backend and merging a commit locally. The downside of this approach was that while loading the app initially for the first time, when client was joining all the mls groups with external commit, we were spamming with lock/unlock websocket methods.

This PR introduces a better solution - an incoming MLS messages queue for each conversation. Instead of processing `conversation.mls-message-add` straight away, we put it in the queue which is unlocked by default (so we do not change the initial behaviour). 

Whenever we send a commit bundle to backend, we simply lock the queue of a particular mls group (allowing the app to receive all other messages) until we receive the response from backend. After queue is unlocked, all the messages that were received while backend was processing the post request, will get handled immediately in order they were received.

### DIFF
--- a/packages/api-client/src/APIClient.ts
+++ b/packages/api-client/src/APIClient.ts
@@ -410,26 +410,6 @@ export class APIClient extends EventEmitter {
   }
 
   /**
-   * Will lock the websocket before executing the callback and unlock it after
-   * @param callback The callback to execute
-   */
-  public withLockedWebSocket<T extends (...args: any[]) => any>(
-    callback: T,
-  ): (...args: Parameters<T>) => Promise<ReturnType<T>> {
-    return async (...args: Parameters<T>): Promise<ReturnType<T>> => {
-      this.transport.ws.lock();
-      try {
-        const result = await callback(...args);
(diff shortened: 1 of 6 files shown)


packages/api-client/src/APIClient.ts
```diff
@@ -409,26 +409,6 @@ export class APIClient extends EventEmitter {
     throw new Error('No valid client ID.');
   }
 
-  /**
-   * Will lock the websocket before executing the callback and unlock it after
-   * @param callback The callback to execute
-   */
-  public withLockedWebSocket<T extends (...args: any[]) => any>(
```
PatrykBuniX, line 416: This one is not used anymore

aweiss-dev (approved):

atomrc (approved):
Amazing job man. Pretty clean solution with a nice api design 👏

PR #5840, opened 2024-01-03
feat(e2ei): remove device storage for e2ei cert

aweiss-dev:
This cert storage was a leftover from early development. Certificates are stored by core-crypto already and should be retrieved from there.

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -238,13 +238,11 @@ export class Account extends TypedEventEmitter<Events> {
     displayName,
     handle,
     discoveryUrl,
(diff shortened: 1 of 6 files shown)


PatrykBuniX (approved):

packages/core/src/Account.ts
```diff
@@ -238,11 +238,13 @@ export class Account extends TypedEventEmitter<Events> {
     displayName,
     handle,
     discoveryUrl,
+    hasActiveCertificate,
```
atomrc, line 241: why not querying coreCrypto down the line rather than expecting the consumer to give that information?

PR #5857, opened 2024-01-08
fix: Delete secrets store when account is wiped

atomrc:
This will make sure the `secets-db ` is wiped when the account is wiped. 

For that, we need to keep track of the secretDB instance in order to wipe it. So we need to change a bit how secret keys are generated. 

Previously generating a secret key would fire up a DB that would be closed right after (and potentially re-openned later on). 
Now we always create the secret DB and eventually use it to generate a key

### DIFF
--- a/packages/core/src/secretStore/secretKeyGenerator.test.ts
+++ b/packages/core/src/secretStore/secretKeyGenerator.test.ts
@@ -17,135 +17,50 @@
  *
  */
 
-import {Decoder, Encoder} from 'bazinga64';
+import {EncryptedStore, createCustomEncryptedStore, createEncryptedStore} from './encryptedStore';
+import {generateSecretKey} from './secretKeyGenerator';
 
-import {generateSecretKey, CorruptedKeyError} from './secretKeyGenerator';
-
-const systemCryptos = {
-  v0: {
-    encrypt: async (value: Uint8Array) => {
-      return value;
-    },
-    decrypt: async (value: Uint8Array) => {
-      return value;
-    },
-    version: undefined,
-  },
-
-  v01: {
-    encrypt: async (value: Uint8Array) => {
-      return Encoder.toBase64(value).asBytes;
-    },
-    decrypt: async (value: Uint8Array) => {
-      const str = new TextDecoder().decode(value);
-      return Decoder.fromBase64(str).asBytes;
-    },
-    version: undefined,
-  },
-
-  v1: {
-    encrypt: async (value: string) => {
-      const encoder = new TextEncoder();
-      return encoder.encode(value);
-    },
-    decrypt: async (value: Uint8Array) => {
-      const decoder = new TextDecoder();
-      return decoder.decode(value);
-    },
-    version: 1,
-  },
+const customCrypto = {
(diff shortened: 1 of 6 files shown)


packages/core/src/secretStore/secretKeyGenerator.test.ts
```diff
@@ -17,135 +17,50 @@
...
-    encrypt: async (value: string) => {
-      const encoder = new TextEncoder();
-      return encoder.encode(value);
-    },
-    decrypt: async (value: Uint8Array) => {
-      const decoder = new TextDecoder();
-      return decoder.decode(value);
-    },
-    version: 1,
-  },
```
atomrc, line 56: Those tests for the handling of different system crypto version has been moved to the webapp (PR coming soon). 
The core should not be aware of any system crypto (use for Electron).

packages/core/src/secretStore/secretKeyGenerator.ts
```diff
@@ -38,50 +32,15 @@ export type GeneratedKey = {
...
-          if (isBase64.test(decrypted)) {
-            return Decoder.fromBase64(decrypted).asBytes;
-          }
-          // Between June 2022 and October 2022, the systemCrypto returned a string encoded in UTF-8
-          const encoder = new TextEncoder();
-
-          return encoder.encode(decrypted);
-        },
-      }
-    : undefined;
```
atomrc, line 79: Those tests for the handling of different system crypto version has been moved to the webapp (PR coming soon).
The core should not be aware of any system crypto (use for Electron).

packages/core/src/secretStore/secretKeyGenerator.test.ts
```diff
@@ -17,135 +17,50 @@
...
-
-      expect(key).toBeDefined();
-
-      try {
-        await generateSecretKey({dbName, systemCrypto: crypto2, keyId});
-      } catch (e) {
-        expect(e).toBeInstanceOf(CorruptedKeyError);
-      }
-    },
-  );
```
atomrc, line 124: see https://github.com/wireapp/wire-web-packages/pull/5857/files#r1444722319

packages/core/src/secretStore/secretKeyGenerator.test.ts
```diff
@@ -17,135 +17,50 @@
...
  */
 
-import {Decoder, Encoder} from 'bazinga64';
+import {EncryptedStore, createCustomEncryptedStore, createEncryptedStore} from './encryptedStore';
+import {generateSecretKey} from './secretKeyGenerator';
 
-import {generateSecretKey, CorruptedKeyError} from './secretKeyGenerator';
-
-const systemCryptos = {
-  v0: {
```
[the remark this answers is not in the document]
atomrc: correct, this logic now completely lives in the webapp. since the webapp version and the core version are always corrolated, we will not need to handle backward compat (see https://github.com/wireapp/wire-webapp/pull/16491/files#diff-5d5af779079c6b10bb91ed68e976c0bad494d65f6fdca4b1c4c30a85707363ecR1)
But that is an excellent question :)

PatrykBuniX (approved):
Looks good 🚀

PR #5860, opened 2024-01-11
fix: Do not upload public keys when renewing certificate

atomrc:
Public keys need to be uploaded only once the certification process is first done. Then it should not change after renewal


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -594,10 +596,6 @@ export class MLSService extends TypedEventEmitter<Events> {
    */
   private async uploadMLSPublicKeys(client: RegisteredClient) {
     // If we've already updated a client with its public key, there's no need to do it again.
-    if (typeof client.mls_public_keys.ed25519 === 'string' && client.mls_public_keys.ed25519.length > 0) {
```
atomrc, line 597: The problem was that the client was not necessarily refreshed with the latest data at this point.

aweiss-dev (approved):

PR #5863, opened 2024-01-12
feat: fetch local certificate root [WPB-3338]

PatrykBuniX:
Before the first enrolment, we need to fetch the local root certificate and pass it to core-crypto for it to be validated.

This should be done only once.

### DIFF
--- a/packages/core/src/messagingProtocols/mls/E2EIdentityService/Connection/AcmeServer/AcmeService.ts
+++ b/packages/core/src/messagingProtocols/mls/E2EIdentityService/Connection/AcmeServer/AcmeService.ts
@@ -48,6 +48,7 @@ import {
   OidcChallengeResponseSchema,
   GetCertificateResponseData,
   GetCertificateResponseSchema,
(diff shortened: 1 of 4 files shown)


packages/core/src/messagingProtocols/mls/E2EIdentityService/Connection/AcmeServer/AcmeService.ts
```diff
@@ -57,10 +58,16 @@ export class AcmeService {
...
   private readonly url = {
     DIRECTORY: '/directory',
+    ROOTS: '/roots.pem',
   };
 
   constructor(private discoveryUrl: string) {}
 
+  private get acmeBaseUrl() {
+    const hostname = new URL(this.discoveryUrl).hostname;
+    return `https://${hostname}`;
```
aweiss-dev, line 68: You can get the protocol from
The url as well

aweiss-dev (approved):

aweiss-dev (approved):

PR #5896, opened 2024-01-23
refactor: Improve E2EI enrollment error handling

atomrc:
This is preparation of installing core-crypto 1.0.0-rc.31 that will change the format of the challenges returned.
It improves a few areas:
- error reporting (avoid, as much as possible, catching errors in the deep layers of the flow, and let them bubble up)
- types refactoring (avoid having duplicated definitions)
- minor code improvements


PatrykBuniX (approved):
Looks good 👌

PR #5897, opened 2024-01-23
feat: Upgrade CoreCrypto to 1.0.0-rc.31

atomrc:
In core-crypto rc31, now challenges are created in 2 steps

### DIFF
--- a/packages/core/src/messagingProtocols/mls/E2EIdentityService/Steps/Authorization.ts
+++ b/packages/core/src/messagingProtocols/mls/E2EIdentityService/Steps/Authorization.ts
@@ -20,46 +20,55 @@
 import {AcmeService} from '../Connection';
 import {E2eiEnrollment, NewAcmeAuthz, Nonce} from '../E2EIService.types';
 import {jsonToByteArray} from '../Helper';
+import {AuthData} from '../Storage/E2EIStorage.schema';
 
 interface GetAuthorizationParams {
   nonce: Nonce;
-  authzUrl: string;
+  authzUrls: string[];
   identity: E2eiEnrollment;
   connection: AcmeService;
 }
-export type AuthorizationChallenge = {authorization: NewAcmeAuthz; nonce: Nonce};
 
 export const getAuthorizationChallenges = async ({
-  authzUrl,
+  authzUrls,
   nonce,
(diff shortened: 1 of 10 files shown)


packages/core/src/messagingProtocols/mls/E2EIdentityService/Steps/Authorization.ts
```diff
@@ -20,46 +20,55 @@
...
-    return {
-      authorization,
-      nonce: response.nonce,
-    };
+  for (const authzUrl of authzUrls) {
+    const reqBody = await identity.newAuthzRequest(authzUrl, nonce);
+    const response = await connection.getAuthorization(authzUrl, reqBody);
+    const backendChallenge = response.data.challenges[0];
+    const challenge = await identity.newAuthzResponse(jsonToByteArray(response.data));
+    challenges.push({type: backendChallenge.type, challenge});
```
atomrc, line 45: This is the main change. Now the `order` gives us back 2 urls, and we need to create 2 challenges from CoreCrypto with those 2 urls

packages/core/src/messagingProtocols/mls/E2EIdentityService/Steps/Authorization.ts
```diff
@@ -20,46 +20,55 @@
...
-      },
-    };
-    return {
-      authorization,
-      nonce: response.nonce,
-    };
+  for (const authzUrl of authzUrls) {
+    const reqBody = await identity.newAuthzRequest(authzUrl, nonce);
+    const response = await connection.getAuthorization(authzUrl, reqBody);
+    const backendChallenge = response.data.challenges[0];
```
PatrykBuniX, line 43: **question**: is that always a first item on the challenges list?

PatrykBuniX (approved):

packages/core/src/messagingProtocols/mls/E2EIdentityService/Steps/Authorization.ts
atomrc: That's a very valid question. 
yes, it's guarantee that there is only going to be a single value in this array (I've checked).

I'll add a comment, good catch

PR #5901, opened 2024-01-25
feat(e2ei): Allow configuring the certificate time to live

atomrc:
When calling `account.enrollE2EI` you can now give a `certificateTtl` that will define the number of days the generated certificate will be valid


PatrykBuniX (approved):
Looks good

packages/core/src/Account.ts
```diff
@@ -240,12 +240,15 @@ export class Account extends TypedEventEmitter<Events> {
     teamId,
     discoveryUrl,
     oAuthIdToken,
+    certificateTtl = 90,
```
atomrc, line 243: oh no, this is wrong, it should not be 90, it should be 90 days 😅

PatrykBuniX (approved):

PR #5924, opened 2024-01-31
runfix: epoch mismatch subconversations

PatrykBuniX:
When detecting epoch mismatch in a "conference" subconversation (a MLS group we use for mls calls) we were mistakenly rejoining its parent conversation via external commit instead of rejoining the actual subconversation.


atomrc (approved):
LGTM

PR #5947, opened 2024-02-07
feat: Queue recovering MLS conversations on notification stream processing

atomrc:
When we detect an epoch mismatch when handling the notification stream, we should avoid directly trying to rejoin the conversation. This could trigger some unexpected behavior. 

So we should first: 
- process the entire notification streams (we might even be able to decrypt some messages as there is a certain tolerance threshold for decrypting messages from future epochs)
- then rejoin the conversations that are on the wrong epoch.

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -56,6 +56,7 @@ import {LinkPreviewService} from './linkPreview';
 import {MLSService} from './messagingProtocols/mls';
 import {
   pauseRejoiningMLSConversations,
+  queueConversationRejoin,
   resumeRejoiningMLSConversations,
 } from './messagingProtocols/mls/conversationRejoinQueue';
 import {E2EIServiceExternal, User} from './messagingProtocols/mls/E2EIdentityService';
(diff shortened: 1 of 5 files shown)


packages/core/src/Account.ts
```diff
@@ -654,25 +658,20 @@ export class Account extends TypedEventEmitter<Events> {
       }
     });
 
-    const handleMissedNotifications = async (notificationId: string) => {
-      if (this.service?.mls) {
-        await this.service?.conversation.handleConversationsEpochMismatch();
-      }
-      return onMissedNotifications(notificationId);
-    };
```
atomrc, line 662: We don't need to eagerly rejoin conversation as soon as we detect missed messages. This process will be kicked off as soon as we receive (or send) a messages in the out-of-sync conversations

PatrykBuniX: The goal of this method is to run only if we've detected that we've missed messages (usually after 28 days). For each known conversation, we would then compare local epoch with the remote one and rejoin if necessary. Is this scenario still covered with this PR?

PatrykBuniX (approved):
Looks good!

PR #5949, opened 2024-02-07
runfix: add backoff when uploading commit bundle

PatrykBuniX:
When uploading a commit fails with 409 error (conflict), we should not retry immediately but rather wait some time before retrying. This Pr adds exponential backoff and waits 0.5, 1, 2, 4, 8, 16 or 32 seconds between retries.


atomrc (approved):
solid PR

PR #5972, opened 2024-02-13
refactor: Improve and streamline client initialization

atomrc:
BREAKING CHANGE: now the `registerClient` does not automatically initialize the client. It needs to be done in 2 steps:

```js
const localClient = await account.registerClient(...)
await account.initClient(localClient);
```

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -34,7 +34,6 @@ import {CONVERSATION_EVENT} from '@wireapp/api-client/lib/event';
 import {Notification} from '@wireapp/api-client/lib/notification/';
 import {AbortHandler, WebSocketClient} from '@wireapp/api-client/lib/tcp/';
 import {WEBSOCKET_STATE} from '@wireapp/api-client/lib/tcp/ReconnectingWebsocket';
-import {FEATURE_KEY, FeatureStatus} from '@wireapp/api-client/lib/team';
 import {QualifiedId} from '@wireapp/api-client/lib/user';
 import {TimeInMillis} from '@wireapp/commons/lib/util/TimeUtil';
 import logdown from 'logdown';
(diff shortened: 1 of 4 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.types.ts
```diff
@@ -44,10 +44,6 @@ export interface MLSServiceConfig {
    * default ciphersuite to use for MLS (MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519 = 1 by default)
    */
   cipherSuite: Ciphersuite;
-  /**
-   * signals if the E2E - Identity process should be used
-   */
-  useE2EI?: boolean;
```
atomrc, line 50: This config is now only given when initializing the localClient (to indicate whether keypackages should be uploaded or not)

packages/core/src/Account.ts
```diff
@@ -351,69 +341,45 @@ export class Account extends TypedEventEmitter<Events> {
...
-    if (this.service.mls) {
-      const {userId, domain = ''} = this.apiClient.context;
-      await this.initMLSClient({id: userId, domain}, client);
-    }
-    this.logger.info(`Created new client {mls: ${!!this.service.mls}, id: ${clientId}}`);
-
     await this.service.notification.initializeNotificationStream(clientId);
     await this.service.client.synchronizeClients(clientId);
-
-    return this.initClient(client);
```
atomrc, line 363: No more hidden agenda of the `registerClient` method. It now really only registers the client on backend (the cryptographic material has to be initialized by the consumer)

packages/core/src/Account.ts
```diff
@@ -230,16 +229,6 @@ export class Account extends TypedEventEmitter<Events> {
...
 
-  private async getE2EIStatus() {
-    const features = await this.apiClient.api.teams.feature.getAllFeatures();
-    const clientCanUseE2EI = this.coreCryptoConfig?.mls?.useE2EI;
-    const teamCanUseE2EI = features[FEATURE_KEY.MLSE2EID]?.status === FeatureStatus.ENABLED;
-
-    return {
-      isFeatureEnabled: clientCanUseE2EI && teamCanUseE2EI,
-    };
-  }
```
atomrc, line 241: It's the responsibility of the consumer to get the config. The core is too low level to parse those kind of config

PatrykBuniX (approved):
Looks much cleaner! Good job 😎

PR #5989, opened 2024-02-19
chore: upgrade to CoreCrypto rc38

atomrc:
## Pull Request Checklist


packages/core/src/Account.ts
```diff
@@ -431,6 +431,7 @@ export class Account extends TypedEventEmitter<Events> {
     const assetService = new AssetService(this.apiClient);
 
     const [clientType, cryptoClient] = await this.buildCryptoClient(context, this.storeEngine, this.encryptedDb);
+    this.logger.info(`CryptoClient of type ${clientType} created (version ${cryptoClient.version})`);
```
atomrc, line 434: Adding some logs in order to know exactly which version of CC we have loaded

PR #5991, opened 2024-02-19
runfix: recover from commit proposals failure [WPB-6729]

PatrykBuniX:
It's possible that a MLS group get stuck in state where it had a pending commit, which prevents any further operations to be performed in a that group. This PR allows a client to recover from that state by clearing all the pending commits and trying to commit them one more time.

### DIFF
--- a/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
+++ b/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
@@ -61,7 +61,7 @@ import {
   queueIncomingMLSMessage,
   withLockedMLSMessagesQueue,
 } from '../EventHandler/events/messageAdd';
-import {ClientId, CommitPendingProposalsParams, HandlePendingProposalsParams} from '../types';
+import {ClientId, HandlePendingProposalsParams} from '../types';
 import {generateMLSDeviceId} from '../utils/MLSId';
 
 //@todo: this function is temporary, we wait for the update from core-crypto side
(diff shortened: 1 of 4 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -717,17 +714,32 @@ export class MLSService extends TypedEventEmitter<Events> {
...
+
+      this.logger.warn('Failed to commit proposals, clearing the pending commit and retrying', error);
+
+      // If we failed to commit the proposals, we need to clear the pending commit and retry
+      // this is to avoid a situation where we are stuck with pending proposals that we can't commit.
+      // If there's nothing to clear the methods might throw an error, which we can ignore.
+      await this.coreCryptoClient.clearPendingCommit(groupIdBytes).catch(() => undefined);
+      await this.coreCryptoClient.clearPendingGroupFromExternalCommit(groupIdBytes).catch(() => undefined);
+
+      return this.commitPendingProposals(groupId, false);
```
atomrc, line 742: I'm not sure I understand the logic of clearing everything and retrying. Aren't we going to retry with an empty list of pending proposals (basically a no-op)?

PatrykBuniX: We're not clearing the pending proposal, but the local commit that contains the proposal (which we want to get rid of).

atomrc (approved):

PR #5996, opened 2024-02-20
fix: Use indexeddb to store e2ei enrollment temporary data

atomrc:
Will use indexedDB as storage for temporary e2ei enrollment flow. 

Since indexeddb is unique per user, this will ensure that no other user logged in on the same browser will share enrollment data


PatrykBuniX (approved):

packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
```diff
@@ -205,9 +216,9 @@ export class E2EIServiceExternal extends TypedEventEmitter<Events> {
    *
    * Both must be registered before the first enrollment.
    */
-  public async registerServerCertificates(): Promise<void> {
+  private async registerServerCertificates(): Promise<void> {
     const ROOT_CA_KEY = 'e2ei_root-registered';
-    const store = LocalStorageStore(ROOT_CA_KEY);
+    const store = LocalStorageStore(this.coreDatabase.name);
```
atomrc, line 221: We need to make sure that this key is unique per user (in case someone logs in to another account on the same browser)

PatrykBuniX (approved):

PR #6004, opened 2024-02-22
refactor: lazily ask consumer for oAuth token

atomrc:
BREAKING CHANGE: enrolling to E2EI doesn't take an `oAuthToken` directly but rather a function that will be called when the token is needed. This makes the consumer not needing to be aware of the start/continuation of the flow (in case there is a redirect needed). 

## Pull Request Checklist

### DIFF
--- a/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
+++ b/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
@@ -854,7 +854,7 @@ export class MLSService extends TypedEventEmitter<Events> {
     certificateTtl: number,
     getOAuthToken: getTokenCallback,
   ): Promise<EnrollmentProcessState> {
-    const hasActiveCertificate = await this.coreCryptoClient.e2eiIsEnabled(this.config.cipherSuite);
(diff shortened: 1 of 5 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -864,25 +864,11 @@ export class MLSService extends TypedEventEmitter<Events> {
...
-    const rotateBundle = !hasActiveCertificate
-      ? // If we are not refreshing the active certificate, we need to continue the certificate process with Oauth
-        await e2eiServiceInternal.continueCertificateProcess(oAuthIdToken)
-      : // If we are refreshing the active certificate, can start the refresh process
-        await e2eiServiceInternal.renewCertificate(oAuthIdToken, hasActiveCertificate);
+    const rotateBundle = await e2eiServiceInternal.generateCertificate(getOAuthToken, isCertificateRenewal);
 
-    if (rotateBundle === undefined) {
-      throw new Error('Could not get the rotate bundle');
-    }
```
atomrc, line 882: This logic is now in the `E2EIServiceInternal`

PatrykBuniX (approved):
Good job!

PR #6018, opened 2024-02-27
runfix: return undefined for user identities if group does not exist [WPB-6864]

PatrykBuniX:
`CoreCrypto.getUserIdentities` would fail when called if the group does not exist in the store anymore (e.g. after user has left it or got removed).
Before trying to call this method, we check if group exists, if it doesn't we return `undefined`.

### DIFF
--- a/packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
+++ b/packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
@@ -81,7 +81,13 @@ export class E2EIServiceExternal extends TypedEventEmitter<Events> {
     return this.coreCryptoClient.e2eiIsEnabled(this.mlsService.config.cipherSuite);
   }
 
-  public async getAllGroupUsersIdentities(groupId: string): Promise<Map<string, DeviceIdentity[]>> {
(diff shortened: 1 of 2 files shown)


packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
```diff
@@ -81,7 +81,13 @@ export class E2EIServiceExternal extends TypedEventEmitter<Events> {
...
   }
 
-  public async getAllGroupUsersIdentities(groupId: string): Promise<Map<string, DeviceIdentity[]>> {
+  public async getAllGroupUsersIdentities(groupId: string): Promise<Map<string, DeviceIdentity[]> | undefined> {
+    const conversationExists = await this.mlsService.conversationExists(groupId);
+
+    if (!conversationExists) {
+      return undefined;
+    }
+
```
atomrc, line 90: do we really need the check here since `getUsersIdentities` also perform this check?

PatrykBuniX: `mlsService.getClientIds` on a line below would throw as well if the mls group does not exist

atomrc: oh I see, makes sense :)

atomrc (approved):

PR #6021, opened 2024-02-28
fix: strip https scheme from fetched self crls

atomrc:
## Description

We should not try to eagerly fetch self CRLs. This can cause some problems (as the url we use to fetch them might not match the url set in the certificate). 

Instead we rely only on the new distribution point detection mechanism in place to update CRLs lazily. 

## Pull Request Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "@wireapp/api-client": "workspace:^",
     "@wireapp/commons": "workspace:^",
-    "@wireapp/core-crypto": "1.0.0-rc.44",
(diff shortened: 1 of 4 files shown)


packages/core/package.json
```diff
@@ -13,7 +13,7 @@
   "dependencies": {
     "@wireapp/api-client": "workspace:^",
     "@wireapp/commons": "workspace:^",
-    "@wireapp/core-crypto": "1.0.0-rc.44",
+    "@wireapp/core-crypto": "1.0.0-rc.43",
```
atomrc, line 16: reverting cc upgrade as this version broke enrollment

PatrykBuniX (approved):

PR #6022, opened 2024-02-28
runfix(commons): do not clear exponential backoff timeout

PatrykBuniX:
We should never clear the backoffed timeout - it could result in being stuck in an unresolved promise.


PatrykBuniX:
Closing as the better solution is on the way

PR #6030, opened 2024-03-01
feat(e2ei): use qualifiedId as key and add to DeviceIdentity

aweiss-dev:
**What's New:**

1. added qualifiedId to our DeviceIdentity type. Now, devices are easier to match with users.
2. qualifiedId is also now the key we use in our maps, making sure every device is unique, no matter what environment it's from.


PatrykBuniX (approved):
Nice 💪

PR #6046, opened 2024-03-05
runfix: low precision task scheduler with exact same delay and key

PatrykBuniX:
When LowPrecisionTaskScheduler.addTask is called, the pending task with exactly the same delay and key should be overwritten by new values. Previously we were just pushing the task to the list of tasks, allowing duplicate tasks to be executed.

### DIFF
--- a/packages/core/src/util/LowPrecisionTaskScheduler/LowPrecisionTaskScheduler.test.ts
+++ b/packages/core/src/util/LowPrecisionTaskScheduler/LowPrecisionTaskScheduler.test.ts
@@ -86,6 +86,30 @@ describe('LowPrecisionTaskScheduler', () => {
     expect(mockedTask2).toHaveBeenCalled();
   });
 
+  it('adding a task with the same delay and key should overwrite the previous task', async () => {
+    const mockedTask1 = jest.fn().mockReturnValue(Promise.resolve('hello task 1'));
+
+    LowPrecisionTaskScheduler.addTask({
+      key: 'same-key',
+      firingDate: 5000,
+      intervalDelay: 1000,
+      task: mockedTask1,
+    });
+
+    LowPrecisionTaskScheduler.addTask({
+      key: 'same-key',
+      firingDate: 7000,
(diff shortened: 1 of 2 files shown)


packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
```diff
@@ -319,7 +319,8 @@ export class E2EIServiceExternal extends TypedEventEmitter<Events> {
   }
 
   private async handleNewCrlDistributionPoints(distributionPoints: string[]): Promise<void> {
-    for (const distributionPointUrl of distributionPoints) {
+    const uniqueDistributionPoints = Array.from(new Set(distributionPoints));
```
atomrc, line 322: 👍

packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
```diff
@@ -285,7 +285,7 @@ export class E2EIServiceExternal extends TypedEventEmitter<Events> {
   }
 
   private async addCrlDistributionTimer({expiresAt, url}: {expiresAt: number; url: string}): Promise<void> {
-    await this.coreDatabase.add('crls', {expiresAt, url}, url);
+    await this.coreDatabase.put('crls', {expiresAt, url}, url);
```
atomrc, line 288: what's the exact difference here?

atomrc (approved):

packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
PatrykBuniX: Docs says that `.add` rejects if an item of a given key already exists in the store, while `.put` puts an item in the database and replaces any item with the same key. 

It's very likely that the method with an already existing in the store key (url), so it's more safe to use put which will alternatively just update the existing entry.

PR #6050, opened 2024-03-08
runfix: join with ext commit on welcome with deleted key

PatrykBuniX:
When following corecrypto error is thrown, we should join the conversation with external commit:

`Although this Welcome seems valid, the local KeyPackage it references has already been deleted locally. Join this group with an external commit`

This is so called Orphan Welcome message, backend does not track claimed KeyPackages. Only way to recover is to rejoin with ext commit. 

This can happen when a client is being added to multiple mls groups at the same time.
Screenshot 2024-03-07 at 15 47 52

### DIFF
--- a/packages/core/src/conversation/ConversationService/ConversationService.ts
+++ b/packages/core/src/conversation/ConversationService/ConversationService.ts
@@ -63,7 +63,7 @@ import {decryptAsset} from '../../cryptography/AssetCryptography';
 import {MLSService} from '../../messagingProtocols/mls';
 import {queueConversationRejoin} from '../../messagingProtocols/mls/conversationRejoinQueue';
 import {
-  isCoreCryptoMLSWelcomeMessageDeletedKeyPackageError,
+  isCoreCryptoMLSOrphanWelcomeMessageError,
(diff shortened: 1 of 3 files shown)


aweiss-dev (approved):

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -716,7 +719,29 @@ export class ConversationService extends TypedEventEmitter<Events> {
...
+          throw new Error('Qualified conversation id is missing in the event');
+        }
+
+        this.logger.info(
+          `Received an orphan welcome message, joining the conversation (${conversationId.id}) via external commit...`,
+        );
+
+        void queueConversationRejoin(conversationId.id, () => this.joinByExternalCommit(conversationId));
+        return null;
+      }
```
atomrc, line 741: I believe this logic should live in the `MLSService` rather than `ConversationService`. 

As much as possible the ConverstaionService should not be aware of MLS specific logic. 

Anyhow, not a blocker for this PR, can be done in a follow refactor PR :)

atomrc (approved):

PR #6070, opened 2024-03-13
fix: Address revoked devices not being detected

atomrc:
see https://github.com/wireapp/core-crypto/blob/develop/CHANGELOG.md#100-rc51---2024-03-13

### DIFF
--- a/packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIService.types.ts
+++ b/packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIService.types.ts
@@ -24,6 +24,7 @@ import {
   RotateBundle,
   WireIdentity,
   E2eiConversationState,
+  DeviceStatus,
   CredentialType,
 } from '@wireapp/core-crypto';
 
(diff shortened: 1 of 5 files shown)


V-Gira (approved):
:tada:

packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIService.types.ts
```diff
@@ -24,6 +24,7 @@ import {
   RotateBundle,
   WireIdentity,
   E2eiConversationState,
+  DeviceStatus,
```
atomrc, line 27: we need to expose this as it's being consumed in the webapp now

packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceExternal.ts
```diff
@@ -47,6 +47,24 @@ type Events = {
...
+}
+
+// TODO coreCrypto types are wrong here. They return a string (the key of the enum) instead of the enum value
+function fixConversationState(value: any): E2eiConversationState {
+  const fixedValue = E2eiConversationState[value];
+  if (!fixedValue) {
+    throw new Error(`Invalid conversation status: ${value}`);
+  }
+  return fixedValue as unknown as E2eiConversationState;
+}
```
atomrc, line 66: Those are temporary functions that can be removed when CC fixes the type mismatch

aweiss-dev (approved):

PatrykBuniX (approved):

PR #6110, opened 2024-03-26
feat(core): Allow downloading and decrypting assets

atomrc:
This will allow the consuming app to decrypt assets as they are being downloaded

### DIFF
--- a/packages/core/src/conversation/ConversationService/ConversationService.ts
+++ b/packages/core/src/conversation/ConversationService/ConversationService.ts
@@ -59,7 +59,6 @@ import {
 } from './ConversationService.types';
 
 import {MessageTimer, MessageSendingState, RemoveUsersParams} from '../../conversation/';
-import {decryptAsset} from '../../cryptography/AssetCryptography';
 import {MLSService} from '../../messagingProtocols/mls';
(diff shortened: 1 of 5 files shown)


packages/core/src/cryptography/AssetCryptography/AssetCryptography.ts
atomrc, line 1: This file is just a rename of the `index.ts`. There are no changed in there (git cannot know that this is a rename since there is also a new file with the same name)

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -164,22 +162,6 @@ export class ConversationService extends TypedEventEmitter<Events> {
...
-      cipherText: new Uint8Array(encryptedBuffer),
-      keyBytes: otrKey,
-      sha256: sha256,
-    });
-  }
-
-  public async getUnencryptedAsset(assetId: string, assetToken?: string): Promise<ArrayBuffer> {
-    const request = await this.apiClient.api.asset.getAssetV3(assetId, assetToken);
-    return (await request.response).buffer;
-  }
```
atomrc, line 181: Those methods did not really make sense here

PR #6167, opened 2024-04-22
feat: Add method to subscribe to wire pro plans

atomrc:
The subscription flow will change a bit in the near future. 
This new method will be the preferred way of subscription to a Wire pro plan


PatrykBuniX (approved):

packages/api-client/src/team/payment/PaymentData.ts
```diff
@@ -56,3 +56,298 @@ export interface PaymentData {
   suspend?: PaymentSuspension;
   trialEndsAt: number;
 }
+
+export type Subscription = {
```
atomrc, line 60: This can be ignored in the review, this is just a bump of the backend's reponse

PatrykBuniX (approved):

PR #6168, opened 2024-04-24
feat: Allow consumer to configure mls ciphersuite when client is init

atomrc:
This will allow a consumer to initial the MLS client with some custom cipherSuite. 

The MLS Config have been moved close to the `initClient` method to allow loading the config as late as possible in the general init process 

## Pull Request Checklist


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -66,23 +66,33 @@ import {
...
 //they are returning regular array instead of Uint8Array for commit and welcome messages
 export const optionalToUint8Array = (array: Uint8Array | []): Uint8Array => {
   return Array.isArray(array) ? Uint8Array.from(array) : array;
 };
 
-interface LocalMLSServiceConfig extends MLSServiceConfig {
-  /**
-   * minimum number of key packages client should have available (configured to half of nbKeyPackages)
-   */
-  minRequiredNumberOfAvailableKeyPackages: number;
```
atomrc, line 79: This has been moved to a `getter` that computes the value when needed

packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -66,23 +66,33 @@ import {
...
-}
-
-const defaultConfig: MLSServiceConfig = {
+const defaultConfig = {
   keyingMaterialUpdateThreshold: 1000 * 60 * 60 * 24 * 30, //30 days
   nbKeyPackages: 100,
-  cipherSuite: Ciphersuite.MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519,
```
atomrc, line 85: we don't provide default ciphersuite, the consumer needs to manually set the suite they want to use

packages/core/src/Account.ts
```diff
@@ -112,7 +112,7 @@ interface AccountOptions {
   nbPrekeys: number;
 
   /**
-   * Config for MLS and proteus devices. Will fallback to the old cryptobox logic if not provided
+   * Config for coreCrypto in case it supposed to be used. Will fallback to the old cryptobox logic if not provided
```
V-Gira, line 115:
```suggestion
   * Config for coreCrypto in case it's supposed to be used. Will fallback to the old cryptobox logic if not provided
```
the nitpicks are back :sweat_smile:

aweiss-dev (approved):

packages/core/src/Account.ts
atomrc: good catch, eagle eye 👁️

PatrykBuniX (approved):
Looks solid :D

PR #6184, opened 2024-04-29
fix: Pass the correct ciphersuite when starting the e2ei enrollment process

atomrc:
Previous PR enabling dynamic ciphersuite selection missed one hardcoded ciphersuite when enrolling to the e2ei process

### DIFF
--- a/packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceInternal.ts
+++ b/packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceInternal.ts
@@ -60,7 +60,7 @@ export class E2EIServiceInternal {
    * @param getOAuthToken function called when the process needs an oauth token
    * @param refresh should the process refresh the current certificate or get a new one
    */
-  public async generateCertificate(getOAuthToken: getTokenCallback, refresh: boolean) {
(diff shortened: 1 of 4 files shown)


packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceInternal.ts
```diff
@@ -102,12 +102,9 @@ export class E2EIServiceInternal {
 
   // ############ Internal Functions ############
 
-  private async initIdentity(hasActiveCertificate: boolean) {
+  private async initIdentity(hasActiveCertificate: boolean, ciphersuite: Ciphersuite) {
     const {user} = this.initialData;
 
-    // How long the issued certificate should be maximal valid
-    const ciphersuite = Ciphersuite.MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519;
-
```
V-Gira, line 110: by "missed", you mean it wasn't removed and we should get it from the config only?

atomrc: Exactly :)

V-Gira (approved):

PR #6185, opened 2024-04-30
fix: Correctly detect if current device is MLS capable

atomrc:
This will allow the consumer to know if an MLS device has been setup or not. 
This is to prevent trying to do MLS operations if the devices has not been initialized

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -255,7 +255,7 @@ export class Account extends TypedEventEmitter<Events> {
       throw new Error('Client has not been initialized - please login first');
     }
 
-    if (!this.service?.mls || !this.service?.e2eIdentity) {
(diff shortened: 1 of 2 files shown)


packages/core/src/Account.ts
```diff
@@ -620,7 +627,7 @@ export class Account extends TypedEventEmitter<Events> {
     });
 
     const handleMissedNotifications = async (notificationId: string) => {
-      if (this.service?.mls) {
```
atomrc, line 623: now, the fact that the mls service is instantiated doesn't necesarily means the mls device is ready to use. We need to add some more checks before doing any kind of mls operation. 

This is due to the fact that we do not yet have the MLS config when the core is instantiated. So we cannot decide whether the MLS service should be instantiated at build time. Thus we need some extra checks

V-Gira (approved):

PR #6202, opened 2024-05-06
runfix: mls default config

PatrykBuniX:
We should not overwrite the consumer-provided MLS configuration by the default config, but rather use this config on top of the default config. This PR addresses that. We also make sure we filter out undefined values, in this case the default value will be used (consumer might want to set a value to `undefined` explicitly to make sure the default will be used).

### DIFF
--- a/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
+++ b/packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
@@ -152,9 +152,10 @@ export class MLSService extends TypedEventEmitter<Events> {
     {skipInitIdentity, ...mlsConfig}: InitClientOptions,
   ) {
     this._config = {
-      ...mlsConfig,
       ...defaultConfig,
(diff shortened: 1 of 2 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -151,10 +151,16 @@ export class MLSService extends TypedEventEmitter<Events> {
...
+    // filter out undefined values from mlsConfig
+    const filteredMLSConfig = Object.fromEntries(
+      Object.entries(mlsConfig).filter(([_, value]) => value !== undefined),
+    ) as typeof mlsConfig;
+
     this._config = {
-      ...mlsConfig,
       ...defaultConfig,
+      ...filteredMLSConfig,
     };
```
V-Gira, line 162: Question: So the main fix is changing the order here, and the secondary fix is to filter out the `undefined` correct?

PR #6287, opened 2024-06-10
fix(api-client): socket reconnect after coming back from sleep

PatrykBuniX:
Once we've realised that the app was (**probably**) sleeping and during the possible "sleep time" the socket was disconnected, we will immediately reconnect. This should fix the bug where after coming back from sleep mode, messages are not synced immediately.

### DIFF
--- a/packages/api-client/src/tcp/ReconnectingWebsocket.ts
+++ b/packages/api-client/src/tcp/ReconnectingWebsocket.ts
@@ -84,6 +84,12 @@ export class ReconnectingWebsocket {
 
     this.hasUnansweredPing = false;
 
+    /**
+     * According to https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine, navigator.onLine attribute and 'online' and 'offline' events are not reliable enough (especially when it's truthy).
+     * We won't receive the 'offline' event when the system goes to sleep (e.g. closing the lid of a laptop).
+     * In this case navigator.onLine will still return true, but the WebSocket connection could be closed.
+     * To handle this, we need a custom approach to detect when the system goes to sleep and when it wakes up.
+     * **/
     onBackFromSleep({
       callback: () => {
         if (this.socket) {
--- a/packages/api-client/src/tcp/ReconnectingWebsocket.ts
+++ b/packages/api-client/src/tcp/ReconnectingWebsocket.ts
@@ -48,7 +48,7 @@ export class ReconnectingWebsocket {
   private static readonly RECONNECTING_OPTIONS: Options = {
     WebSocket: WebSocketNode,
     connectionTimeout: TimeUtil.TimeInMillis.SECOND * 4,
-    debug: false,
+    debug: true,
     maxReconnectionDelay: TimeUtil.TimeInMillis.SECOND * 10,
     maxRetries: Infinity,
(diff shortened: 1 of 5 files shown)


packages/api-client/src/tcp/ReconnectingWebsocket.ts
```diff
@@ -173,16 +184,10 @@ export class ReconnectingWebsocket {
     return this.socket ? this.socket.readyState : WEBSOCKET_STATE.CLOSED;
   }
 
-  public disconnect(reason = 'Closed by client', keepClosed = true): void {
+  public disconnect(reason = 'Closed by client'): void {
     if (this.socket) {
       this.logger.info(`Disconnecting from WebSocket (reason: "${reason}")`);
-      // TODO: 'any' can be removed once this issue is resolved:
-      // https://github.com/pladaria/reconnecting-websocket/issues/44
-      (this.socket as any).close(CloseEventCode.NORMAL_CLOSURE, reason, {
```
PatrykBuniX, line 181: This method does not accept the 3rd (options) parameter any more.

packages/api-client/src/utils/BackFromSleepHandler/BackFromSleepHandler.ts
```diff
@@ -0,0 +1,72 @@
...
+    const wasAsleep = currentTime > lastTime + TOLERANCE;
+
+    lastTime = currentTime;
+
+    if (isDisconnectedCallback && isDisconnectedCallback()) {
+      wasDisconnected = true;
+    }
+
+    if (wasAsleep) {
+      if (!isDisconnectedCallback || wasDisconnected) {
```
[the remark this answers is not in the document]
PatrykBuniX: No, if the check callback is not provided at all we want to always run the callback

packages/api-client/src/tcp/ReconnectingWebsocket.ts
```diff
@@ -82,6 +83,22 @@ export class ReconnectingWebsocket {
...
+     * To handle this, we need a custom approach to detect when the system goes to sleep and when it wakes up.
+     * **/
+    onBackFromSleep({
+      callback: () => {
+        if (this.socket) {
+          this.logger.debug('Back from sleep, reconnecting WebSocket');
+          this.socket?.reconnect();
+        }
+      },
+      isDisconnected: () => this.getState() === WEBSOCKET_STATE.CLOSED,
```
atomrc, line 100: Wasn't the ping/pong mechanism enough to detect that the connection was closed? 
Is there some kind of timer that runs to check if a message has been received in a certain amount of time?

PatrykBuniX: Yeah, we have the mechanism for that, but it's a ping every 20 seconds, which doesn't seem to be efficient enough to reconnect immediately after coming back from sleep. From what I was testing, it was the 20s that we had to wait for the weboscket to reconnect after coming back from sleep.

atomrc: Humm that makes sense, 20s is, indeed, pretty long time. 

Then your solution seems solid to me 😎

atomrc (approved):

PR #6324, opened 2024-06-26
fix: Improve logger verbosity

atomrc:


PatrykBuniX (approved):

V-Gira (approved):

tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "removeComments": false,
     "sourceMap": false,
     "strict": true,
-    "target": "es2017",
+    "target": "es2022",
```
atomrc, line 15: WE can safely put a recent target as the consuming app will, in the end, transpile again using babel (or other tool)

PR #6326, opened 2024-06-28
fix: do not create an RTC object for check

EnricoSchw:
To check if `RTCDataChannel's` are offered we creating `RTCPeerConnection` objects
Creating `RTCPeerConnection` objects leads to too many unnecessary ghost objects in the background that are not cleared away. This can potentially lead to errors in the app.

Bildschirmfoto 2024-06-19 um 17 16 39

## Pull Request Checklist


EnricoSchw:
Reminder: @wireapp/web

PR #6502, opened 2024-09-03
feat: establish 1on1 mls conversation with new removal key [WPB-10744]

V-Gira:
## Description

See https://wearezeta.atlassian.net/browse/WPB-10737 for a more in depth description of the issue.

In a nutshell, it's possible for a federated user to create a 1:1 conversation even if it's not hosted on their own backend, in which case it can introduce a host of issues because they're lacking the correct removal key.

to address that issue, the response from the endpoint `GET /conversations/one2one/{domain}/{userId}` was changed to include that removal key, from
```
{ <conversation metadata> }
```
to
```
{
  "conversation": { <conversation metadata> },
  "public_keys": {
    "removal": {
      "<sig-algo>": "string",
      ...
    }
  }
}
```

## Checklist

- [x] mentions the JIRA issue in the PR name (Ex. [WPB-423])
- [x] PR has been self reviewed by the author;
- [x] Hard-to-understand areas of the code have been commented;

[WPB-423]: https://wearezeta.atlassian.net/browse/WPB-423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api-client/src/conversation/Conversation.ts
+++ b/packages/api-client/src/conversation/Conversation.ts
@@ -19,6 +19,7 @@
 
 import {RECEIPT_MODE} from './data';
 
+import {MLSPublicKeyRecord} from '../client';
 import {QualifiedId} from '../user';
 
 import {ConversationMembers, ConversationProtocol} from './';
@@ -106,4 +107,10 @@ export interface MLSConversation extends Conversation {
   epoch: number;
   cipher_suite: number;
   protocol: ConversationProtocol.MLS;
+}
+
+export interface MLS1on1Conversation extends MLSConversation {
+  public_keys: {
+    removal: MLSPublicKeyRecord;
+  };
 }
--- a/packages/api-client/src/conversation/Conversation.ts
+++ b/packages/api-client/src/conversation/Conversation.ts
@@ -118,9 +118,9 @@ export interface MLS1to1Conversation {
 
 export function isMLS1to1Conversation(response: unknown): response is MLS1to1Conversation {
   if (typeof response === 'object' && response !== null) {
-    const conversation = response as MLS1to1Conversation;
+    const mls1to1Conversation = response as MLS1to1Conversation;
 
-    return !!conversation.conversation && !!conversation.public_keys;
+    return !!mls1to1Conversation.conversation && !!mls1to1Conversation.public_keys;
   }
 
   return false;
--- a/packages/api-client/src/conversation/Conversation.ts
+++ b/packages/api-client/src/conversation/Conversation.ts
@@ -120,7 +120,7 @@ export function isMLS1to1Conversation(response: unknown): response is MLS1to1Con
   if (typeof response === 'object' && response !== null) {
     const conversation = response as MLS1to1Conversation;
 
-    return !!conversation.conversation && !!conversation.public_keys
+    return !!conversation.conversation && !!conversation.public_keys;
   }
 
   return false;
--- a/packages/api-client/src/conversation/Conversation.ts
+++ b/packages/api-client/src/conversation/Conversation.ts
@@ -117,10 +117,8 @@ export interface MLS1to1Conversation {
 }
 
 export function isMLS1to1Conversation(response: unknown): response is MLS1to1Conversation {
-  if (typeof response === 'object' && response !== null) {
-    const mls1to1Conversation = response as MLS1to1Conversation;
-
-    return !!mls1to1Conversation.conversation && !!mls1to1Conversation.public_keys;
+  if (typeof response === 'object' && response !== null && 'conversation' in response && 'public_keys' in response) {
+    return true;
   }
 
   return false;
--- a/packages/api-client/src/conversation/Conversation.ts
+++ b/packages/api-client/src/conversation/Conversation.ts
@@ -115,3 +115,17 @@ export interface MLS1to1Conversation {
     removal: MLSPublicKeyRecord;
   };
 }
+
+export function isMLS1to1Conversation(response: unknown): response is MLS1to1Conversation {
+  if (typeof response === 'object' && response !== null) {
+    const conversation = response as MLS1to1Conversation;
+
+    if (!!conversation.conversation && !!conversation.public_keys) {
+      return true;
+    }
+
+    return false;
+  }
+
+  return false;
+}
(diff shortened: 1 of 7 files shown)


packages/core/src/conversation/SubconversationService/SubconversationService.ts
```diff
@@ -151,46 +151,6 @@ export class SubconversationService extends TypedEventEmitter<Events> {
...
-    } catch (error) {
-      this.logger.error(`Failed to delete conference subconversation:`, error);
-    }
-
-    await this.mlsService.wipeConversation(subconversationGroupId);
-
-    // once we've deleted the subconversation, we can remove it from the store
-    await this.clearSubconversationGroupId(conversationId, SUBCONVERSATION_ID.CONFERENCE);
-  }
-
```
V-Gira, line 193: This code block was a previous attempt to solve a related issue without requiring a change to the backend response, by calling the `deleteSubconversation` endpoint instead of `deleteSubconversationSelf`.

see https://github.com/wireapp/wire-web-packages/pull/6434

This is no longer required

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -580,19 +587,19 @@ export class ConversationService extends TypedEventEmitter<Events> {
...
+      return (await this.getMLS1to1Conversation(otherUserId)).conversation;
     }
 
     // If group is not established on backend,
     // we wipe the it locally (in case it exsits in the local store) and try to register it.
     await this.mlsService.wipeConversation(groupId);
 
     try {
-      await this.mlsService.register1to1Conversation(groupId, otherUserId, selfUser);
+      await this.mlsService.register1to1Conversation(groupId, otherUserId, selfUser, public_keys?.removal);
```
V-Gira, line 598: This is the main change, we can register a conversation with a removal key that comes from a different endpoint

packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -481,7 +485,8 @@ export class MLSService extends TypedEventEmitter<Events> {
     } else {
       const mlsKeys = (await this.apiClient.api.client.getPublicKeys()).removal;
       const ciphersuiteSignature = getSignatureAlgorithmForCiphersuite(this.config.defaultCiphersuite);
-      const removalKeyForSignature = mlsKeys[ciphersuiteSignature];
+      const removalKeyForSignature =
+        removalKeyFor1to1Signature?.[ciphersuiteSignature] ?? mlsKeys[ciphersuiteSignature];
```
V-Gira, line 489: if the removal key is provided by the conversation endpoint, we use that one, else, we use a removal key provided by getPublicKey (our own backend removal key)

packages/api-client/src/conversation/Conversation.ts
```diff
@@ -107,3 +108,24 @@ export interface MLSConversation extends Conversation {
...
+    const conversation = response as MLS1to1Conversation;
+
+    if (!!conversation.conversation && !!conversation.public_keys) {
+      return true;
+    }
+
+    return false;
+  }
+
+  return false;
```
e-maad: It is possible to normalize the logic?
Can we return early with false value?

packages/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -536,7 +537,12 @@ export class ConversationService extends TypedEventEmitter<Events> {
    * @param userId - qualified user id
    */
   async getMLS1to1Conversation(userId: QualifiedId) {
-    return this.apiClient.api.conversation.getMLS1to1Conversation(userId);
+    const response = await this.apiClient.api.conversation.getMLS1to1Conversation(userId);
+
+    if (isMLS1to1Conversation(response)) {
+      return response;
+    }
+    return {conversation: response};
```
e-maad, line 545:
```suggestion
    const conversation = await this.apiClient.api.conversation.getMLS1to1Conversation(userId);

    if (isMLS1to1Conversation(response)) {
      return conversation;
    }
    return {conversation};
```

e-maad (commented):
Added few suggestions.

packages/core/src/conversation/ConversationService/ConversationService.ts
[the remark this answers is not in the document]
V-Gira: > Can't we return response.conversation and response? 🤔 Code will be more readable.

This function returns 
```
export interface MLS1to1Conversation {
  conversation: MLSConversation;
  public_keys?: {
    removal: MLSPublicKeyRecord;
  };
}
```
the optional `public_keys` is needed in some cases so we can't only return MLSConversation

V-Gira: darn, this *is* more readable :D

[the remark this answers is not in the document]
V-Gira: @przemvs see if you like @e-maad 's suggestion I applied

packages/api-client/src/conversation/Conversation.ts
V-Gira: Something like this?

```typescript
export function isMLS1to1Conversation(response: unknown): response is MLS1to1Conversation {
  if (typeof response === 'object' && response !== null && 'conversation' in response && 'public_keys' in response) {
    return true;
  }

  return false;
}

File: packages/core/src/conversation/ConversationService/ConversationService.ts [hunk 6]
```diff
@@ -580,19 +587,19 @@ export class ConversationService extends TypedEventEmitter<Events> {
       );
 
       await this.joinByExternalCommit(mlsConversation.qualified_id);
-      return this.getMLS1to1Conversation(otherUserId);
+      return (await this.getMLS1to1Conversation(otherUserId)).conversation;
```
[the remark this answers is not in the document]
V-Gira: I went with this, WDYT?
```typescript
      const {conversation: updatedMLSConversation} = await this.getMLS1to1Conversation(otherUserId);
      return updatedMLSConversation;
```

PR #6636, opened 2024-10-31
feat(core): Add support for InCallHandRaise proto message

thisisamir98:
## Description

Updated core logic to handle the new message type.

Implements https://github.com/wireapp/generic-message-proto/pull/90


packages/core/src/conversation/InCallHandRaiseType.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+/**
+ * true if the hand is raised, false if lowered
+ */
+export type InCallHandRaiseType = boolean;
```
e-maad, line 23: One file for one type 😟 
Do you think we should move all the types in the same file?

e-maad: I think we don't need it we can directly use boolean.

packages/core/src/conversation/content/InCallHandRaiseContent.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {InCallHandRaiseType} from '../InCallHandRaiseType';
+
+export interface InCallHandRaiseContent {
+  is_hand_up: InCallHandRaiseType;
```
e-maad, line 23: Out of curiosity, why we are not following variable naming convention here?
```suggestion
  isHandUp: boolean;
```

e-maad (commented):
Added few suggestions.

packages/core/src/conversation/content/InCallHandRaiseContent.ts
thisisamir98: this is because of the messaging proto

thisisamir98: done in https://github.com/wireapp/wire-web-packages/pull/6636/commits/59db7920db7ee8d294af0895a52eefc607f0fbd7

packages/core/src/conversation/InCallHandRaiseType.ts
thisisamir98: done in https://github.com/wireapp/wire-web-packages/pull/6636/commits/59db7920db7ee8d294af0895a52eefc607f0fbd7

packages/core/src/conversation/message/MessageBuilder.ts
```diff
@@ -309,6 +311,20 @@ export function buildInCallEmojiMessage(payloadBundle: InCallEmojiMessage['conte
   });
 }
 
+export function buildInCallHandRaiseMessage(payloadBundle: InCallHandRaiseMessage['content']): GenericMessage {
+  const {isHandUp} = payloadBundle;
+
+  const handRaise = InCallHandRaise.create({
+    isHandUp: isHandUp,
```
e-maad, line 318:
```suggestion
    isHandUp,
```

e-maad (approved):
LGTM 🚀

PR #6696, opened 2024-11-26
feat(core, api-client): Serialize object arguments in desktop runtime

thisisamir98:
See https://github.com/wireapp/wire-webapp/pull/18359

### DIFF
--- a/packages/api-client/src/APIClient.ts
+++ b/packages/api-client/src/APIClient.ts
@@ -21,6 +21,8 @@ import logdown from 'logdown';
 
 import {EventEmitter} from 'events';
 
+import {LogFactory} from '@wireapp/commons';
+
 import {AccountAPI} from './account/AccountAPI';
 import {AssetAPI} from './asset/';
 import {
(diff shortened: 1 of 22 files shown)


packages/api-client/src/APIClient.ts
```diff
@@ -200,7 +200,7 @@ export class APIClient extends EventEmitter {
   }
 
   private configureApis(backendFeatures: BackendFeatures): Apis {
-    this.logger.info('configuring APIs with config', backendFeatures);
+    this.logger.info('configuring APIs with config', JSON.stringify(backendFeatures));
```
olafsulich, line 203: after changes in https://github.com/wireapp/wire-webapp/pull/18359, shouldn't it be done automatically?

thisisamir98: web-packages use a separate instance of logdown of their own

thisisamir98: see https://github.com/search?q=repo%3Awireapp%2Fwire-web-packages%20%3D%20logdown(&type=code

olafsulich: okay, so maybe we should copy-paste the solution from the webapp you have provided? then we wouldn't have to stringify those args in the future, WDYT?

thisisamir98: doing a big refactor

thisisamir98: https://github.com/wireapp/wire-web-packages/pull/6696/commits/b25192005674d1fb61b74012850a9fc0fa9a4030

olafsulich (approved):

e-maad (approved):

PR #6727, opened 2024-12-03
runfix: remove slashes from conversation and user api constants

V-Gira:
## Description

Slashes are hardcoded into some of the url constants in the conversation and user api;

This remove those slashes, allowing us to use those constant where the API endpoints change the order around, as is the case for V&, see https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/1309868033/API+changes+v6+v7

## Checklist

### DIFF
--- a/packages/api-client/src/user/UserAPI.ts
+++ b/packages/api-client/src/user/UserAPI.ts
@@ -74,27 +74,27 @@ export class UserAPI {
   public static readonly DEFAULT_USERS_CHUNK_SIZE = 50;
   public static readonly DEFAULT_USERS_PREKEY_BUNDLE_CHUNK_SIZE = 128;
   public static readonly URL = {
-    ACTIVATE: '/activate',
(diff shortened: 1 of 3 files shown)


packages/api-client/src/user/UserAPI.ts
```diff
@@ -171,7 +171,7 @@ export class UserAPI {
    * @see https://staging-nginz-https.zinfra.io/swagger-ui/#!/users/getUserClient
    */
   public async getClient(userId: QualifiedId, clientId: string): Promise<PublicClient> {
-    const url = `${UserAPI.URL.USERS}/${userId.domain}/${userId.id}/${UserAPI.URL.CLIENTS}/${clientId}`;
+    const url = `//${UserAPI.URL.USERS}/${userId.domain}/${userId.id}/${UserAPI.URL.CLIENTS}/${clientId}`;
```
e-maad, line 174: Double slash?

PR #6770, opened 2024-12-16
!fix(MLS): check feature flag before creating MLS clients [WPB-15020]

aweiss-dev:
## Description

Fix an issue with creating MLS clients even tho the Feature_Flag was diabled

### DIFF
--- a/packages/core/src/Account.ts
+++ b/packages/core/src/Account.ts
@@ -34,6 +34,7 @@ import {CONVERSATION_EVENT} from '@wireapp/api-client/lib/event';
 import {Notification} from '@wireapp/api-client/lib/notification/';
 import {WebSocketClient} from '@wireapp/api-client/lib/tcp/';
 import {WEBSOCKET_STATE} from '@wireapp/api-client/lib/tcp/ReconnectingWebsocket';
+import {FEATURE_KEY, FeatureStatus} from '@wireapp/api-client/lib/team';
 import {QualifiedId} from '@wireapp/api-client/lib/user';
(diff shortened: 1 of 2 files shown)


packages/core/src/Account.ts
```diff
@@ -719,4 +721,24 @@ export class Account extends TypedEventEmitter<Events> {
...
+      return false;
+    }
+
+    // Backend Supports MLS trough removal keys
+    const isMLSSupported = await this.apiClient.supportsMLS();
+    if (!isMLSSupported) {
+      return false;
+    }
+
+    // MLS is enabled for the team
```
aweiss-dev, line 738:
```suggestion
    // MLS is enabled for the public via feature flag
```

packages/core/src/Account.ts
```diff
@@ -363,7 +364,8 @@ export class Account extends TypedEventEmitter<Events> {
     await this.apiClient.transport.http.associateClientWithSession(client.id);
 
     await this.service.proteus.initClient(this.storeEngine, this.apiClient.context);
-    if (this.service.mls && mlsConfig) {
+
+    if ((await this.isMLSActiveForClient()) && this.service.mls && mlsConfig) {
```
thisisamir98, line 368: this is already `this.service.mls` checked in isMLSActiveForClient

aweiss-dev: Double check doesn't hurt to satisfy TS. The function is public and available for web app, so we need to check the service availability there as well.

thisisamir98 (approved):

PR #6877, opened 2025-02-03
feat: update CC to 3.0.2

aweiss-dev:
## Description

There was a bad timing with another NPM package update. So we need to bump the lockgile change


V-Gira (changes requested):
Published on npm already

PR #6987, opened 2025-03-20
feat(api-client/cells): dynamic cells configuration [WPB-16524]

olafsulich:
## Description

We need to change how we configure Cells in the web application. Currently, the config is static (just an API key), but we have to make it dynamic, based on the current user data. 

## Checklist

### DIFF
--- a/packages/api-client/src/cells/CellsAPI.ts
+++ b/packages/api-client/src/cells/CellsAPI.ts
@@ -33,25 +33,66 @@ import {
 import {CellsStorage} from './CellsStorage/CellsStorage';
 import {S3Service} from './CellsStorage/S3Service';
 
-import {Config} from '../Config';
+import {AccessTokenStore} from '../auth';
 import {HttpClient} from '../http';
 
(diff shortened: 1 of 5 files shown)


packages/api-client/src/cells/CellsAPI.ts
```diff
@@ -67,6 +108,10 @@ export class CellsAPI {
     file: File;
     autoRename?: boolean;
   }): Promise<RestCreateCheckResponse> {
+    if (!this.client || !this.storageService) {
+      throw new Error(CONFIGURATION_ERROR);
+    }
```
e-maad, line 113: Why not create a function and use it in all the places?

e-maad (commented):
Added a suggestion.

packages/api-client/src/cells/CellsAPI.ts
olafsulich: Then it doesn't check the correct types, I tried to create a type guard / assert the correct type with a function, but it didn't work

olafsulich: e.g. `this.client` could be `null`, after this check, it'll be the non-nullable type

PR #7044, opened 2025-05-13
feat(react-ui-kit): dropdown menu & breadcrumbs components

olafsulich:
## Description

Adds Breadcrumbs and DropdownMenu components.

DropdownMenu is meant to replace an internal web app implementation (showContextMenu) in the future. Currently, it's missing some options (like adding icons, etc.), but they should be added later, while migrating specific use cases. 

We should have a similar component to DropdownMenu: ContextMenu. From the UI point of view, there's no difference, but from an accessibility perspective, it is (ContextMenu fires on the right click). ContextMenu is available in Radix as well.

--- 

Besides new components, I added some CSS variables (like z-index) - we should aim to unify the most common ones, and have them in here, instead of in the web app.

## Demo

**DropdownMenu:**

https://github.com/user-attachments/assets/b18e0f4a-cb67-459c-b0bd-e3d11afb6759

**Breadcrumbs:**

https://github.com/user-attachments/assets/f6e96068-1607-4f95-9b00-73d282334adb

## Checklist


e-maad (approved):
LGTM 👍

thisisamir98 (approved):

PR #7051, opened 2025-05-20
feat!: CC bump to 6. 

aweiss-dev:
## Description

The new CC version includes a breaking change for us regarding the Database Key. The format changes from 128 to 256-bit key length. Additionally, the library introduced its Key type. 

To make this work, we have to run a database migration for existing clients (only once).
To achieve this goal, we generate a new encryption key and store it in our database next to the existing one.

Following cases exist:

1. If the old key exists and the new key is freshly generated -> We run the migration, delete the old key and return the new key
2. If the old key is freshly generated and the new key is freshly generated -> We return the new key
3. If the old key is freshly generated and the new key exists -> We return the new key

### DIFF
--- a/packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceInternal.ts
+++ b/packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceInternal.ts
@@ -270,14 +270,14 @@ export class E2EIServiceInternal {
     return this.coreCryptoClient.transaction(async cx => {
       const conversations = await getAllConversations();
       const newCrlDistributionPoints = await cx.saveX509Credential(identity, certificate);
-      conversations.forEach(async conversation => {
+      for (const conversation of conversations) {
         if (Boolean(conversation.group_id?.length)) {
           const idAsBytes = new TextEncoder().encode(conversation.group_id);
           await cx.e2eiRotate(idAsBytes);
         } else {
           this.logger.error('No group id found in conversation');
         }
-      });
+      }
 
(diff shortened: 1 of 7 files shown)


packages/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -366,7 +366,7 @@ export class MLSService extends TypedEventEmitter<Events> {
     const welcomeBundle = await this.coreCryptoClient.transaction(cx =>
       cx.joinByExternalCommit(groupInfo, credentialType),
     );
-    await this.dispatchNewCrlDistributionPoints(welcomeBundle.crlNewDistributionPoints);
+    await this.dispatchNewCrlDistributionPoints(welcomeBundle.crl_new_distribution_points);
```
aweiss-dev, line 369: This is a change from the library. I already contacted CC team to revert this

packages/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceInternal.ts
```diff
@@ -270,14 +270,14 @@ export class E2EIServiceInternal {
     return this.coreCryptoClient.transaction(async cx => {
       const conversations = await getAllConversations();
       const newCrlDistributionPoints = await cx.saveX509Credential(identity, certificate);
-      conversations.forEach(async conversation => {
+      for (const conversation of conversations) {
```
aweiss-dev, line 273: forEach doesn't wait for the completion of async operations. For...of does

PR #7109, opened 2025-07-30
feat(core, api-client): Support legacy notifications stream [WPB-19042]

thisisamir98:
We want to support old notification system and new (Async) notification system and implement FF to be able to switch.


packages/api-client/src/tcp/ReconnectingWebsocket.ts
```diff
@@ -53,14 +58,27 @@ export class ReconnectingWebsocket {
 
   private readonly logger: logdown.Logger;
   private socket?: RWS;
-
+  private pingerId?: NodeJS.Timeout;
```
e-maad, line 61: Shouldn't it be a number?
```suggestion
  private pingerId?: number;
```

e-maad (approved):
LGTM, added one concern.

packages/api-client/src/tcp/ReconnectingWebsocket.ts
thisisamir98: image

not really. this is node.js context.

PR #7195, opened 2025-08-06
feat(api-client): Implement breaking changes of api V10 & V11 [WPB-19039]

thisisamir98:
In this PR only the breaking changes of api V10 & V11 have been addressed. support for new endpoints is to be done.


packages/api-client/src/team/sso/SSOAPI.ts
```diff
@@ -55,11 +55,11 @@ export class SSOAPI {
     return await this.client.sendJSON(config);
   }
 
-  public async updateDomainRedirect({domain, backendUrl, domainRedirect}: VerifyDomainRedirectConfig) {
+  public async updateDomainRedirect({domain, backend, domainRedirect}: VerifyDomainRedirectConfig) {
```
e-maad, line 58: @przemvs, I didn't find the usage. Do you remember why we added it?

PR #7255, opened 2025-08-20
fix: endless loading after password reset by mapping token error correctly(WPB-19061)

arjita-mitra:
## Description

After a user resets their password, the webapp fails to redirect to login screen on reload and stays stuck on an empty screen.
Backend responses:
403 /access → {"code":403,"label":"invalid-credentials","message":"Invalid zauth token"}
401 /self

The error message "Invalid zauth token" was not handled in BackendErrorMapper.
As a result, the app never recognized it as an InvalidTokenError, and the logout flow was not triggered.

## Fix
Added explicit mapping for "Invalid zauth token" under StatusCode.FORBIDDEN / INVALID_CREDENTIALS to InvalidTokenError.
Now the app correctly handles the error and redirects to login.

Verified: instead of endless loading, user is redirected to login after password reset and reloading the app.

## Checklist

- [x] mentions the JIRA issue in the PR name (Ex. [WPB-XXXX])
- [x] PR has been self reviewed by the author;

### DIFF
--- a/packages/api-client/src/http/BackendErrorMapper.test.ts
+++ b/packages/api-client/src/http/BackendErrorMapper.test.ts
@@ -19,7 +19,13 @@
 
 import {BackendErrorMapper} from './BackendErrorMapper';
 
-import {InvalidCredentialsError, MissingCookieError, SuspendedAccountError, TokenExpiredError} from '../auth/';
+import {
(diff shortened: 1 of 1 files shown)


e-maad (approved):
LGTM 👍

PR #7279, opened 2025-08-29
fix(core, promise-queue): Timeout error for message decryption tasks (WPB-18410)

thisisamir98:
## Description

- promise queue will reject promises that timeout instead of letting them float forever and advancing to the next item in the queue
- added logs on message decryption flow
- flushing pending proposals queue if the connection drops, same as notifications processing queue it'll be populated once socket reconnects

### DIFF
--- a/packages/promise-queue/src/PromiseQueue.test.ts
+++ b/packages/promise-queue/src/PromiseQueue.test.ts
@@ -234,4 +234,33 @@ describe('PromiseQueue', () => {
       expect(queue.getLength()).toBe(0);
     });
   });
+
+  describe('stack safety', () => {
+    jest.setTimeout(30000);
+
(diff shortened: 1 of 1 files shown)


aweiss-dev (changes requested):
The current implementation of the Promise Queue doenst work like a "background task" and will run into call stack size issues fast in larger teams. 

We have a syncronous recursion, which is considered unsafe in javascript. Unless we would have a set maximum for the queue. 
What we need is a async recursion, since the queue could potentially stack up to inifinite.


// Test this in chrome dev tools for example
const func1 = () => {
  console.log('Running the Function recursive and break the call stack');
  setTimeout(func1);
}
func1();

const func2 = () => {
  console.log('Running the Function recursive');
  func2();
}
func2();

aweiss-dev (approved):

